PR #6, opened 2015-03-21
Documented String Resources

scento:
This pull request adds all [officially documented string resources](https://developer.android.com/reference/com/google/android/gms/R.string.html), because developers might use them when developing applications using the official Play Services API.


mar-v-in:
I think these strings must go into the developer api (=GmsLib) and not the service code (=GmsCore) or am I wrong? (same applies to the colors)

scento:
Okay, I reverse engineered the `GmsCore.apk` included in `gapps-lp-20141212-signed` and realized, that all of the resources I added are included, but I there are ~2500 additional ones. I will add those too.

mar-v-in:
Resources that are not publicly described should never be used by developers so we don't have to care them.

scento:
I am sure, that the original google apps are using them. But feel free to close this pull request - the entire rest of apps with an gapps requirement will work without them.

mar-v-in:
I had no problems with original gapps yet and I tried a lot of them...

scento:
Okay, then thank you for clarification.

PR #33, opened 2015-08-27
Added travis build

oNaiPs:
Useful to double check all changes are good and package is built.
Additional feature could be uploading the APK to github when tagging a release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: android
+git:
+  submodules: false
+before_install:
+  - git submodule update --init --recursive
+script:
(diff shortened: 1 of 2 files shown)


mar-v-in:
I removed the components `extra-google-google_play_services` and `extra-google-m2repository` as they should not be required and might cause problems in the future.

Merged with commit 3519d33 Thanks.

oNaiPs:
Thanks! I think you'll need to go to
[travis](https://travis-ci.org/microg/android_packages_apps_GmsCore) yourself so that the build status will appear correctly in the readme

mar-v-in:
Already did that, Travis will start building with the next commit

PR #44, opened 2015-10-09
Add German translation

julienschmidt:
Since there are already Serbian and Romanian translations, here is one translation I can provide.
I think you are a native German speaker, too. So you might want to change a few things :wink:


mar-v-in:
Please do not put non-traslatables into a translation file.

julienschmidt:
Sorry, should be fixed now.

PR #67, opened 2015-12-31
Bugfix for fixing AbstractMethodError in Marshmallow

moosd:
Using play-services in Marshmallow gives an `AbstractMethodError` because InputMethodManager calls `getOpPackageName()` instead of `getPackageName()` on Context. See line 1179 in https://github.com/android/platform_frameworks_base/blob/lollipop-release/core/java/android/view/inputmethod/InputMethodManager.java vs https://github.com/android/platform_frameworks_base/blob/marshmallow-release/core/java/android/view/inputmethod/InputMethodManager.java

Since ContextContainer implements Context but does not fill in this method, errors happen. This small patch will fix the error (tested on cm13).

Also fixes moosd/Needle#1

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/maps/ContextContainer.java
+++ b/play-services-core/src/main/java/org/microg/gms/maps/ContextContainer.java
@@ -101,6 +101,10 @@ public class ContextContainer extends Context {
         return original.getPackageName();
     }
 
+    public String getOpPackageName() {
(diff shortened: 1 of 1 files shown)


mar-v-in:
Commit 38364b79ee7fd1de4de5ef565538acc532474061 includes `getOpPackageName` and other hidden Context methods to properly run on M

PR #102, opened 2016-03-01
Android Marshmallow Patch

scento:
This commit adds a patch for Android Marshmallow including:

1. Signature Spoofing
2. Location Provider Support

Before building a patched Android Marshmallow system, it is required to execute `make update-api` in the root directory.


mar-v-in:
It would be great if you'd use https://gerrit.omnirom.org/#/c/16080/ as a base patch for signature faking instead of the LP version as it properly handles dynamic permissions on Android M (when using the LP patch, denying the permission will not disable signature spoofing).

scento:
The pull request has been updated.

PR #140, opened 2016-06-03
Fixed release build with ProGuard

SWW13:
Added wearable-lib as libraryjar.
This fixes the following errors:
```
Warning: org.microg.wearable.SocketWearableConnection$1: can't find enclosing method 'void serverListen(int,org.microg.wearable.WearableConnection$Listener)' in program class org.microg.wearable.SocketWearableConnection
Warning: org.microg.wearable.SocketWearableConnection$2: can't find enclosing method 'void clientConnect(int,org.microg.wearable.WearableConnection$Listener)' in program class org.microg.wearable.SocketWearableConnection

Warning: there were 2 unresolved references to program class members.
         Your input classes appear to be inconsistent.
         You may need to recompile the code.
         (http://proguard.sourceforge.net/manual/troubleshooting.html#unresolvedprogramclassmember)
Warning: Exception while processing task java.io.IOException: Please correct the above warnings first.
:play-services-core:transformClassesAndResourcesWithProguardForRelease FAILED
```

### DIFF
--- a/proguard.flags
+++ b/proguard.flags
@@ -4,6 +4,7 @@
 
 # We're referencing stuff that is unknown to the system
 -libraryjar ../unifiednlp-compat/build/classes/main
+-libraryjar ../wearable-lib/build/classes/main
 -dontwarn java.awt.**
 -dontwarn javax.annotation.**
 


mar-v-in:
I do not get this error and the proposed fix can cause broken builds. 
wearable-lib is not  a libraryjar (which is a library present on the system but unknown to the compiler, obviously not the case for wearable-lib).
if it is required for you to add this flag, it's likely that wearable is broken in your build (it's not very functional right now anyway, but this cannot be a desired solution).

Please make sure you did update all submodules and cleaned your temporary build dirs (`./gradlew clean`), maybe this will already fix your errors.

posativ:
Yup, had the same error. `./gradlew clean build` fixed it.

PR #151, opened 2016-07-10
Build with circleCI

zoff99:
build command for circleci.com

https://circleci.com/gh/zoff99/android_packages_apps_GmsCore/3#artifacts

capture_app_running

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -51,7 +51,6 @@ dependencies:
     - pwd ; ls -al
 
     - echo sdk.dir $ANDROID_HOME > local.properties
-
     - git submodule update --init --recursive
 
     - ./gradlew assembleDebug --stacktrace --info --no-daemon
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,142 @@
+machine:
+  timezone:
+    Europe/Vienna
+  environment:
+    ZEMU_: "appemu23"
+    EMU_: "appemu21"
+    XEMU_: "appemu10"
+    Zsdpath: "/sdcard"
+    sdpath: "/storage/sdcard"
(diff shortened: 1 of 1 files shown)


mar-v-in:
What is the advantage of using Circle CI over Travis CI?

zoff99:
to keep all the files online at circle.
and it can't hurt

mar-v-in:
Do you have unlimited storage at circle ci?

Also note that as the signing key is a different one every time and not trustworthy you can/should not use the apk. So, you might have all files online all time, but not in a way you could use them. And this git repository can be used to build older versions all the time if necessary.

@ "can't hurt": every additional service to manage (update configuration/script etc) is additional work, and additional work does hurt. If it's not managed it will become useless. If you continue to support it (if build breaks because of any changes) and provide updates this is not a problem though.

zoff99:
circle give lots of storage. i have not found a limit yet.
also it's not supposed to be used for daily use. just to debug and test the apk.

you can choose to merge or not. as i already did the work for me, its done already.
if you dont merge keep this PR open, maybe other will want to use it

zoff99:
does travis store the resulting apk for download on travis server?

mar-v-in:
No, travis does not provide storage option. I only use travis to have a defined system for builds, so I can ensure builds also do work on other systems than mine. I already build everytime I git push and upload the build as experimental to the f-droid repo, so there is no real need for a second apk build imho.

I see that Circle CI has enormous possibilities for testing, running the emulator can be used in the future to run full-featured tests including screenshot comparison, but until now we don't have such a test suite and it would be quite some work to establish it.

PR #197, opened 2016-09-27
Added item to disable battery optimizations

Nutomic:
I wasn't sure where to put the item, so I can move it to `PermissionCheckGroup` if you prefer.

Note that the Android settings don't seem to be updated properly, so you have to restart your device to see the change in the "Battery Optimization" screen. 

Requires https://github.com/microg/android_packages_apps_UnifiedNlp/pull/100. Hope I did everything right with the submodule.

### DIFF
--- a/play-services-core/build.gradle
+++ b/play-services-core/build.gradle
@@ -80,7 +80,7 @@ android {
 
     lintOptions {
         // TODO: Remove MissingTranslation once we have stable strings and proper translations.
-        disable 'MissingTranslation', 'InvalidPackage'
+        disable 'MissingTranslation', 'InvalidPackage', 'BatteryLife'
     }
 
     buildTypes {
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -80,6 +80,8 @@
     <uses-permission android:name="com.google.android.c2dm.permission.SEND"/>
     <uses-permission android:name="org.microg.gms.STATUS_BROADCAST"/>
 
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+
     <application
         android:allowBackup="false"
         android:icon="@mipmap/ic_core_service_app"
--- a/play-services-core/src/main/java/org/microg/gms/ui/SettingsActivity.java
+++ b/play-services-core/src/main/java/org/microg/gms/ui/SettingsActivity.java
@@ -16,6 +16,8 @@
 
 package org.microg.gms.ui;
 
+import android.content.Intent;
+import android.os.Build;
 import android.os.Bundle;
 import android.preference.Preference;
 import android.support.annotation.NonNull;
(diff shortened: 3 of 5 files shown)


Nutomic:
I just added the submodule change to my commit, to make Travis happy.

mar-v-in:
Why do you split the feature across both apps this way?

Does UnifiedNlp actually require battery optimizations disabled? If not, all related code should be put into GmsCore repo. If it is, the commit in UnifiedNlp repo seems to be incomplete (permission in manifest)

Nutomic:
Oh I just checked where other items were located, and put it in the same place. Will move it all to this project.

Nutomic:
Okay I moved the permission into UnifiedNlp. I *think* the permission belongs there, but I'm not 100% sure because I don't know what exactly each project does. The reason I put it there is just because it makes sense in the code. I can also move it into GmsCore, but neither InstalledPackagesChecks nor RomSpoofSignatureChecks seem to be the right place, and I don't think it requires a new category either.

mar-v-in:
> The reason I put it there is just because it makes sense in the code.

It makes sense only if it is required for UnifiedNlp standalone to work properly, which I think is not the case. The unifiednlp-base is what is present in both GmsCore and standalone UnifiedNlp, so there shouldn't be any checks that are not required for UnifiedNlp functionality.

The current Self-Check interface is not very nice anyway, so it's not a big deal to have a new category there. At some point the self-check interface should be removed and replaced with a) a setup assistant and b) notification warnings if wrong settings are detected.

ale5000-git:
I would like to still have Self-Check even after.
Settings could be made broken at every time even after installation, so installation check isn't enough.

Please do not follow the trend that remove advanced options, altough there will noob users that would be happy there will be also be advanced users that wouldn't be so happy.

Nutomic:
Alright, I'll put it into a new category. Any idea how to name that?

ale5000-git:
I propose a category "Android settings" (for Android settings that must be changed to make microG work properly) like permissions and battery optimizations.

Nutomic:
Right, I could add it to the permissions group, but that would be kind of ugly in the code. `PermissionCheckGroup` takes all the permissions it should request as parameter. Adding something that's not a permission wouldn't be obvious unless you have a really close look at the code. But I don't know the project enough, so I'll just do whatever @mar-v-in says.

ale5000-git:
Or maybe also a "General" category since this doesn't properly fit.

Nutomic:
@mar-v-in Can you please tell me where I should put the preference? (both in the UI and in the code) Happy to finish this, but I'm not sure what's the best way.

mar-v-in:
As far as I can tell, UnifiedNlp does not require disabling battery optimizations for functionality. The only thing heavily relying on it right now is Cloud Messaging.

I suggest you put a new self-check category for this into GmsCore only. You can call the category other/misc/system/whatever.

Nutomic:
Alright will do that :)

Nutomic:
Moved everything into this project, rebased on master, and did a bit of cleanup in my code.

mar-v-in:
Thanks. Nice work!

PR #249, opened 2016-12-20
Add French translation

ArchangeGabriel:
This definitively not perfect (a lot of technical terms currently don’t have good equivalents in french to my knowledge), but better than nothing. ;)


mar-v-in:
Thanks!

PR #283, opened 2017-01-26
Add Chinese(Taiwan) Translation.

acsway878787:
Sorry if I did it wrong.


mar-v-in:
Looks fine (I can't verify the strings, but technically it's ok), thank you.

PR #296, opened 2017-02-09
Mark GmsCore as a core app

spacekitteh:


mar-v-in:
@spacekitteh what is the benefit of doing so for GmsCore?

spacekitteh:
@mar-v-in it's a precursor to updating the permission spoofing patches. I changed the patches so only packages marked as core, system packages can spoof signatures. This vastly reduces the security risks of using software such as microG.

n76:
Any app developer could add that to their manifest so how is requiring that in the manifest for signature spoofing increasing security?

spacekitteh:
AFAIK, core apps have to be system apps - which means either in a factory
image, or in an OTA update.

mar-v-in:
Apps in the system partition already have "signature spoofing abilities" (provided through an intended security weakness in AOSP), adding an additional requirement to be a core app to use proper signature spoofing thus does not increase security. Don't eyewash yourself or someone else by adding a useless security restriction.

Remember that an ota zip or root app that would be required to put an apk onto /system could also modify the system to support proper signature spoofing.

spacekitteh:
>Apps in the system partition already have "signature spoofing abilities"
(provided through an intended security weakness in AOSP)

Can you elaborate?

mar-v-in:
Signatures on system apps are not properly verified. As long as the AndroidManifest.xml keeps the same, you can change all files inside the apk (code, resources + assets) without breaking the signature. The relevant portion of the source code can be found here: https://github.com/CyanogenMod/android_frameworks_base/blob/cm-14.1/core/java/android/content/pm/PackageParser.java#L1283

spacekitteh:
Thanks, I'll have to fix that in my patch as well.

ale5000-git:
@spacekitteh: Restrict to core apps isn't secure at all.

A better patch would be implement signature spoofing like how UnifiedNLP handle UnifiedNLP backends.

The ROM should list all apps that require signature spoofing and all new apps must be disabled by default (matched against both name and real signature).
And the user the first time must manually enable spoofing for every app that need it in the ROM settings.

PS: If you do something like this, I'm in favor of modify Tingle to apply this patch.

mar-v-in:
@ale5000-git that's basically what the android 6.0+ patches already do, as this is what the dynamic permission system provides. The only thing that would be different when implementing this separately is that apps can no longer request the permission to spoof signatures, which I think is not to good for user experience.

If you build your ROM yourself or install GmsCore+FakeStore into /system/priv-app, you can modify the patch slightly to change `protectionLevel` of the permission to `system` instead `dangerous`. This will restrict the permission to apps in system. Note that if you have permission modify system partition, you can also modify the system, which means that any security system or further restriction is useless as it can just be removed/replaced.

ale5000-git:
@mar-v-in: The insecure part is allow app themself to enable spoofing (there is also tap jacking).

The list of apps with signature spoofing should be completely separated from other things because it is much more unsafe.
I know that an app can spoof only to other apps and not to the system but there may be badly programmed apps that use data from other apps and here come the problem.

mar-v-in:
@ale5000-git I disagree with your opinion that "signature spoofing is much more unsafe compared to other things". Although misuse of it is technically possible, it's very unlikely and actually an issue in the app that is misused (!= the one that is impersonated) itself (I know this includes all apps that use play services client library, but that doesn't mean Google's code is not terribly broken). 
IMHO my contacts, calendar events and messages are the things I want to keep most secret and these can be requested through usual permission.

Tap jacking should be prevented, but there already is proper overlay protection in android's dynamic permission dialogs and activities (which you would have to handle correctly if you implement your own settings). If you want it more secure you could add a confirmation delay (sth. even Lineage might consider adding as an optional feature), such that you can't be driven to accidentally touch the allow button because the dialog just popped-up in the moment you wanted to touch a button below it.

If you want to have additional security, think about why these additional security shouldn't be there (at least optional) for, let's say, the READ_CONTACTS permission.

ccaapton:
@mar-v-in 

> Apps in the system partition already have "signature spoofing abilities" (provided through an intended security weakness in AOSP)

Does this mean if we put microg into priv-app, we get sig spoof for free thus don't need xposed or tinkle/needle? That sounds great!

mar-v-in:
@ccaapton nothing is for free. ;) In this case it would be a huge amount of work to get it running properly and it would block updates via F-Droid because updates have to be installed on /system as well.

The initial idea was not to use Xposed for signature spoofing, the idea was that ROMs support it out of the box. If your UX is terrible, blame your ROM...

ccaapton:
> it would block updates via F-Droid because updates have to be installed on /system as well.
No it is not the case. I'm running microg as privileged right now by copying the apk(a quite old version as I did it quite long ago) into priv-app(with the other copy in /data/app intact). Updating microg is just like others, android will put the new version into /data/app, and still consider it as privileged, as long as the original old apk stays in /system/priv-app. In fact this is also how google updates its gmail, calendar, keybord via play store, not via OTA

mar-v-in:
@ccaapton: I should have been more precise: the inability to update from f-droid is not caused by installing on /system. It's because, when using this method, you are not spoofing the signature for third-party apps alone, you're spoofing it to the whole system. This means that the system will only accept updates that are signed by Google (=> you would be able update microG with the original Google Play Services, but not with microG, as the trick only works for apps on /system). Thus, the only way to update the app is to replace it on /system.

spacekitteh:
Can't packages be signed with multiple keys to produce multiple signatures? I know this breaks the assumption of many apps but doesn't the core system allow this, even with the update manager? To work around the broken assumption, you would probably need a "view hidden signatures" or "show different signatures to different apps" permission.

Further, surely it doesn't just check "you're in /system", but rather `ApplicationInfo.FLAG_SYSTEM_APP`? This way, updated system apps are included.

mar-v-in:
I don't see how multiple signatures will be any benefit here. Note that when having multiple signatures, all signatures have to be present in an update. And the play services check will probably handle multiple signatures and notice the difference.

Regarding check: it really checks app to be in /system (see https://github.com/CyanogenMod/android_frameworks_base/blob/cm-14.1/core/java/android/content/pm/PackageParser.java#L1284) and if it wasn't that would be a critical security issue.

spacekitteh:
Reeeeeally wish google would document their fucking code.

PR #300, opened 2017-02-11
Changed OUT_DIR to TARGET_COMMON_OUT_ROOT

spacekitteh:
Takes into account if compiling a debug build.


ale5000-git:
@spacekitteh: While you are at it, it would be really nice if you fix the same problem also on all other microG projects

PR #316, opened 2017-02-19
Enable compilation with Jack

spacekitteh:


ffgiff:
Jack toolchain is deprecated since 20170314, so it's not worth enabling now.

https://android-developers.googleblog.com/2017/03/future-of-java-8-language-feature.html
https://sites.google.com/a/android.com/tools/tech-docs/jackandjill

PR #337, opened 2017-03-04
Update Russian translation

Gena95:
Translation for new lines and minor string fixes


Gena95:
Can you accept my translation?

PR #365, opened 2017-05-02
Add patch for LineageOS

kylon:


mar-v-in:
@kylon How is this patch different from the one we already have?

Herny64:
i did a diff from the two patch files.
Hes using diff --ruN instead of diff --git.
Couldnt spot another differrence.

kylon:
It does apply with no errors
The last time i used the N patch i had a build error.

@Henry64
Yea i had to manually apply the patch, so i used that command.

No changes to the patch itself, only the line numbers.

kylon:
i just tried again a few minutes ago.
lineage does NOT build with your patch.

here are some lines (too many lines to copy all of them)

patch -p1 < microgpatch.patch

[makeerror.txt](https://github.com/microg/android_packages_apps_GmsCore/files/973585/makeerror.txt)

iRave:
I just built LineageOS with the official patch. I had to reapply it because there where some changes in framework/base. When I did that, there were two warnings about line numbers that changed. 
It patched and build just fine though. Signature Spoofing does work as well after installing the new build.

corna:
@kylon It's just because `patch` creates the `.orig` files in the same folder and the build system finds two sets of definitions, one in `strings.xml` and one in `strings.xml.orig`.

To fix it just run `git clean -f` (or `rm -f strings.xml.orig` to remove just the problematic file) in `frameworks/base` after applying the patch.

@mar-v-in This should be pointed out in the doc, as it applies to any ROM where the patch doesn't apply cleanly on `strings.xml`

roblabla:
You're supposed to apply this patch with `git am` as far as I know. It builds completely fine on my system if you do this.

Catfriend1:
What is needed to get this PR ready?

kylon:
uh! still open

@Catfriend1 
see comments

PR #367, opened 2017-05-06
Russian translation

Gena95:
Update russian translation


Gena95:
Can you add this translation fix?

PR #380, opened 2017-05-22
add Chinese simplified translation

WanderMax:
add Chinese simplified translation for microG


WanderMax:
need embellishment before PR, wait.

PR #476, opened 2018-01-24
Add patch for Android-O

poshul:
Adds the necessary patch information to patch O to allow signature spoofing.


poshul:
Missed existing pull request with equivalent functionality.

PR #477, opened 2018-01-27
Update spanish values/strings.xml traslation on play-services-core

Hattshire:
Added some missing translations


Hattshire:
Ohh, I didn't know the project has that, good to know, I will join there then 😉

PR #485, opened 2018-02-14
Add signature for the google daydream app

prurigro:
This fixes issue https://github.com/microg/android_packages_apps_GmsCore/issues/484


prurigro:
I should note that there still seems to be missing functionality required to get this running past the login screen. This PR just solves the first hurdle that prevents daydream from starting in the first place.

In case it saves someone time; the fact that daydream allows you to access the playstore through it made me think it might be related to why attempting to log in causes it to exit. To test this, I temporarily swapped out fakestore with the opengapps play store, and after checking to make sure the play store was running correctly and the microg self-test passed, Daydream continued to have the same behaviour and logcat output as before.

PR #528, opened 2018-04-18
More interface details for the cast framework

armills:
## Description
This PR stubs out and connects more of the cast framework components. The end result of this PR is that we're able to get the [cast video reference app](https://github.com/googlecast/CastVideos-android) to display the chromecast button with our mock route. It's not much, but it's a start.

### Before merging
- [x] Merge https://github.com/microg/android_external_GmsApi/pull/8 and update submodule ref to latest master.
- [x] Merge https://github.com/microg/android_external_GmsLib/pull/12 and update submodule ref to latest master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ gen/
 bin/
 build/
 .gradle/
+.idea/
 user.gradle
 local.properties
 .directory
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ def androidCompileSdk() { return 27 }
 
 def androidTargetSdk() { return 27 }
 
-def androidMinSdk() { return 9 }
+def androidMinSdk() { return 14 }
 
 def versionCode() {
     def stdout = new ByteArrayOutputStream()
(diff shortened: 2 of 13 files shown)


ale5000-git:
I'm not so much experienced but couldn't use `uses-sdk tools:overrideLibrary` of AndroidManifest.xml to ignore the minSdk set from the library and then only use the incompatible library if the sdk is 14 or higher?

armills:
Ah, I've never seen that before. It does seem possible, I can give it a try.

armills:
OK, I've included overrideLibrary here to keep the `androidMinSdk` version at 9. I suspect that since the entrypoints to this code are from apps requesting a cast context, or from the media router service, it will never get called on a lower SDK anyway, but I haven't included any additional safeguards or tested on an SDK 9 device.

mar-v-in:
@armills do you think I should merge this now or just merge #555 + microg/android_external_GmsApi#9 after I found time to test them?

armills:
I think at this point #555 is in a way better state than when I first split them, so the intermediate PR merge here probably isn't necessary anymore and just dirties the commit history with half working code. Going to close this PR assuming you agree, but feel free to re-open.

My current plan is to get #555 up-to-date with the new safe-parcel release and get a temporary fork published for the chromecast lib while we're figuring out the finder details of the API we need from them. After we've got a published fork can get a working build up and running from that PR for testing.

PR #538, opened 2018-04-29
Signature spoofing patch for Oreo

Golbinex:
Please add this patch from https://github.com/lineageos4microg/docker-lineage-cicd/blob/master/src/signature_spoofing_patches/android_frameworks_base-O.patch
It is not easy to find it


Golbinex:
I am blind -.-

PR #542, opened 2018-05-12
Fix path to generated apk file in Android.mk

marmarek:
Recent build as part of AOSP fails because output path have changed.

I've tried changing `gmscore_apk` (adding `release/` there), but then the build fails too. So I've added a symlink.


ale5000-git:
This should be already fixed I think.

ale5000-git:
@marmarek: This should be already fixed I think.

marmarek:
I've tried that (https://github.com/microg/android_packages_apps_GmsCore/commit/fca83bad2ac0d828676ae9a27e27582c8f6009b6) too, but as written in description here it didn't worked for me for some reason. I'll verify it later. Closing for now.

ale5000-git:
@marmarek: Does it work with the latest version of today?

PR #561, opened 2018-06-20
Added null check, fix #560

### LINKED ISSUE #560: Attempt to invoke virtual method isRoaming() on a null object reference
I get this at boot:
```
[ 06-17 21:30:50.251  3264: 3598 W/GmsGcmMcsSvc ]
Exception while connecting!
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.net.NetworkInfo.isRoaming()' on a null object reference
	at org.microg.gms.gcm.GcmPrefs.getNetworkPrefForInfo(GcmPrefs.java:103)
	at org.microg.gms.gcm.McsService.connect(McsService.java:370)
	at org.microg.gms.gcm.McsService.handleMessage(McsService.java:535)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:154)
	at org.microg.gms.gcm.McsService$HandlerThread.run(McsService.java:152)

[ 06-17 21:30:50.251  3264: 3598 D/GmsGcmMcsSvc ]
Teardown initiated, reason: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.net.NetworkInfo.isRoaming()' on a null object reference
```

The problem is [here](https://github.com/microg/android_packages_apps_GmsCore/blob/master/play-services-core/src/main/java/org/microg/gms/gcm/GcmPrefs.java#L102) but it is caused by the fact that there isn't a null check for the two call of `cm.getActiveNetworkInfo()` [here](https://github.com/microg/android_packages_apps_GmsCore/blob/master/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L366).

**ConnectivityManager.getActiveNetworkInfo():**
`a NetworkInfo object for the current default network or **null** if no default network is currently active.`

ale5000-git:
(need testing)

---
This change is [Reviewable](https://reviewable.io/reviews/microg/android_packages_apps_gmscore/561)

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
+++ b/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
@@ -367,8 +367,14 @@ public class McsService extends Service implements Handler.Callback {
         try {
             closeAll();
             ConnectivityManager cm = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
-            activeNetworkPref = GcmPrefs.get(this).getNetworkPrefForInfo(cm.getActiveNetworkInfo());
-            if (!GcmPrefs.get(this).isEnabledFor(cm.getActiveNetworkInfo())) {
+            NetworkInfo networkInfo = cm.getActiveNetworkInfo();
+            if(networkInfo == null) {
+                scheduleReconnect(this);
+                return;
+            }
+
+            activeNetworkPref = GcmPrefs.get(this).getNetworkPrefForInfo(networkInfo);
+            if (!GcmPrefs.get(this).isEnabledFor(networkInfo)) {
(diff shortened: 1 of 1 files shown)


mar-v-in:
Should be fixed through 7bbdcac990a8bcf939b8012c642cbe2249e64758 ?

ale5000-git:
@mar-v-in: The problem has been fixed.

**PS:** I have just a note, I'm not sure if it is possible but in case the return value of `cm.getActiveNetworkInfo()` change between line 371 and 372 doesn't it remain in a inconsistent state with the current code?

[Here](https://github.com/microg/android_packages_apps_GmsCore/blob/master/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L371).

PR #592, opened 2018-08-11
Increase Google Play Services version number

kazigk:
Just another version bump


mar-v-in:
Done as part of 4040e3df85db5dd59ed3028978a33eb92d7f4e46

PR #601, opened 2018-09-01
Added Android Pie patch file

gudenau:


poshul:
Poking this to see if anything else is needed before merging this.

steamp0rt:
Can confirm that this works. @mar-v-in

PR #628, opened 2018-10-02
change version code generation to ease frequent beta builds

Nanolx:
As F-Droid requires versionCode to be +1 at least for each new build I came up with this to differenciate stable from beta builds based on gitDirty.

Given 92 commits since last tag (or difference compared to upstream/master) and gmsVersion being 13.2.80, this results in:

* build: versionName / versionCode
* official: 0.2.6.13280 / 1328013
* dirty: 0.2.6.13280-dirty-92 / 13280104

I'm open for suggestions.


mar-v-in:
This way, releases can have a lower version number than nightly builds made before the release:

Version name | Current version code | Proposed version code
--- | --- | ---
0.2.6.13280 | 13280012 | 13280012
0.2.6.13280-1 | 13280013 | 13280013
0.2.6.13280-10 | 13280013 | 13280022
0.2.7.13280 | 13280014 | 13280014

Nanolx:
Ah yes, you're right. I simply assumed a new release would also have a new GCM-version, thus it worked out in my head bubble.

Will think of something else after work.

Nanolx:
What about changing the versioncode to *something* like this:

`GGGGGTMMM`

where

* G = GMS Version
* T = Type [0 = nightly ; 9 = stable]
* B = magic number
  * nightly: commits since last tag
  * stable: major minor micro
   * _assuming_ there's no 0.2.10, but that could be catched by `MMMM`

examples

* nightly: 0.2.6.13280-50 = 132800050
* stable: 0.2.7.13280 = 132809027

I have not _yet_ looked at all places where this string is used, AFAIK DroidGuard Helper does (which btw. could also deserve a new relase with @ThibG fix for SafetyNet attestation on non-arm devices).

mar-v-in:
This way stable versions would be always higher version code than nightly versions with the same gms version, even of they are older.

Maybe GGGGGVVCC would work, where VV is the minor of latest release and CC is the number of commits since, restricting us to minor<100 and commits<100 (I don't think it's a problem to bleed into GMS version if necessary)

However there is a problem with older versions of gms client library requiring the seventh digit to be less than 4 IIRC (the digit was used to encode device type before), so we cannot even freely choose that one :(

Nanolx:
Hmm, yes, _if_ you want the user to be able to switch between 'channels'. With my suggestion one could switch from nightly to stable, but not the other way round (except with a newer GMS-Version), that's true (one thing that came to my mind is some kind of bot that produces a new nightly build as soon as a new commit is available, and pushes it to some F-Droid repository).

If we're limited with the seventh digit it's not going to be easy, because all other approaches I came up with would conflict...

ale5000-git:
If I'm not wrong the version code isn't limited to numbers, if it is like this then there are hexadecimal numbers from 00 to FF

Nanolx:
@ale5000-git according to the [docs](https://developer.android.com/studio/publish/versioning) the maximum allows the Google Play is 2100000000, so we're likely not allowed to exceed this one (don't ask me why it's not `MAX_INT`).

I guess @mar-v-in previous idea of GGGGGVVCC is the best, if you want stable and beta channel to be interchangeable, though it limits us to 99 commits, which is easy to hit if you don't re-base and squash after each microG release (though that would only affect forks not upstream).

Nanolx:
new approach based on @mar-v-in last comment. If you build the `version-scheme-devbuild` branch clean you'll now get:

* versionCode: `132802619`
* versionName: `0.2.6.13280-beta-19 (30a127c)`

when building dirty:

* versionCode: `132802619`
* versionName: `0.2.6.13280-beta-dirty-19 (30a127c)`

I simply assumed that is not required for gitDirty to be reflected in the versionCode.

ale5000-git:
@Nanolx
It looks error prone.
I don't know perfectly the code but if "13.2.80" become "13280" and not "130280" then for example:
"15.2.00" will be (incorrectly) lower than "12.10.00".

Also if any of the part of the version became of three digit how the code will react?
I think gitCommitCount can likely become three digit but I think everithing should be checked (at least error out on three digits or lock maximum to 99).

Nanolx:
@ale5000-git that's not different as opposed to current code, also Google never had the minor version above 9, so they would break their versioning scheme themselves if they'd change it (see [play services releases](https://www.apkmirror.com/uploads/?q=google-play-services)).

I also thought about > 99 commits just yesterday and added a proposed change right now.

Nanolx:
Updated to either append either `-mapbox` or `-vtm`. Example with current state:

`Starting build for version 0.2.7.17455-beta-7 (0e1419b-mapbox) (174552707)`
`Starting build for version 0.2.7.17455-beta-7 (0e1419b-vtm) (174552707)`

Nanolx:
This PR also requires a change in RemoteDroidGuard, to strip off the last digit from the GmsCore versionCode (with this versionCode change here we get 9 instead of 8 digits):

```
diff --git a/remote-droid-guard/src/main/java/org/microg/gms/droidguard/DroidguardHelper.java b/remote-droid-guard/src/main/java/org/microg/gms/droidguard/DroidguardHelper.java
index 5cc8c08..fae1c84 100644
--- a/remote-droid-guard/src/main/java/org/microg/gms/droidguard/DroidguardHelper.java
+++ b/remote-droid-guard/src/main/java/org/microg/gms/droidguard/DroidguardHelper.java
@@ -59,7 +59,7 @@ public class DroidguardHelper {
     private static Map<String, Class<?>> loadedClass = new HashMap<>();
 
     public static byte[] guard(Context context, RemoteDroidGuardRequest request) throws Exception {
-        int versionCode = context.getPackageManager().getPackageInfo(Constants.GMS_PACKAGE_NAME, 0).versionCode;
+        int versionCode = context.getPackageManager().getPackageInfo(Constants.GMS_PACKAGE_NAME, 0).versionCode/10;
 
         SignedDGResponse signedResponse = request(new DGRequest.Builder()
                 .usage(new DGUsage(request.reason, request.packageName))
```

Nanolx:
@mar-v-in any further input regarding this proposal?

Nanolx:
Since I've got no response I assume there's no interest in it, so I'm closing this now.

Issue #86, opened 2016-02-01
Gmail sync


nagalun:
When gmail sync is enabled under:
Settings > Accounts > Google > (account) > Gmail: ON,
it does not wait between syncs, here is a logcat:
(Is there another email client that uses GCM like gmail?)
`02-01 11:27:13.863 11596 16471 W Gmail   : Sync started for account: account:1200572429
02-01 11:27:13.864 11596 16471 I Gmail   : notifyAccountChanged
02-01 11:27:13.867 11596 11620 I Gmail   : getAccountsCursor
02-01 11:27:13.871  1472 11676 D GmsAuthenticator: hasFeatures: Account {name=my-account@gmail.com, type=com.google}, [service_mail]
02-01 11:27:13.880 11596 16471 I Gmail   : MainSyncRequestProto: lowestBkwdConvoId: 0, highestHandledServerOp: 126568, normalSync: true
02-01 11:27:13.906 11596 16471 D ProviderInstallerImpl: yep, i should do something with Security here...
02-01 11:27:13.910  1472 12075 D GmsAuthenticator: getAuthToken: Account {name=my-account@gmail.com, type=com.google}, mail, Bundle[{androidPackageName=com.google.android.gm, callerPid=11596, callerUid=10139, notifyOnAuthFailure=true}]
02-01 11:27:13.912  1472 12075 D GmsAuthManager: token present, but expired
02-01 11:27:13.919  1472 12075 D GmsHttpFormClient: -- Request --
02-01 11:27:13.919  1472 12075 D GmsHttpFormClient: accountType=HOSTED_OR_GOOGLE&androidId=my-important-stuff
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: -- Response --
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: SID=my-important-stuff
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: LSID=my-important-stuff
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: Auth=my-important-stuff
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: issueAdvice=auto
02-01 11:27:15.403  1472 12075 D GmsHttpFormClient: services=mail,android,cl,talk,friendview,lso,youtube,chromiumsync,multilogin,lh2,urlshortener,omaha,writely,wise,oz,mymaps,googleme,esmobile
[...]

PR #624, opened 2018-09-29
Add support for Google Suite accounts with external login

jimbo1qaz:
Appears to fix https://github.com/microg/android_packages_apps_GmsCore/issues/622 for me?

PR #555, opened 2018-06-14
MVP Cast API

armills:
## Description
This PR is the next step in the cast implementation, and is capable of playing videos with Google's sample Cast-Videos app. Basic cast functionality is now functional. Device discovery is implemented. Standard cast apps can now connect, start videos, and control playback.

Fixes #79

### Still broken
- There's still something preventing Netflix from starting videos. It must make use of some cast code that isn't yet implemented
- Volume control is not functional
- Joining an existing session is not implemented

### Dependencies
- [x] https://github.com/microg/android_external_GmsApi/pull/9
- [x] https://github.com/microg/android_external_SafeParcel/pull/3
- [x] https://github.com/microg/android_external_SafeParcel/pull/4

PR #635, opened 2018-10-09
Add implementation of ProviderInstallerImpl

### LINKED ISSUE #72: Hangouts [$25]
Hi, what is likely needed in order for Hangouts to work reliably? Hangouts used to sometimes allow for calls and messages, but as of late seems to not even open anymore.

[Bountysource](https://www.bountysource.com/issues/29680305-hangouts)

### LINKED ISSUE #504: Could you implement ProviderInstaller
I see that there is no ProviderInstaller at all in MicroG.
https://github.com/microg/android_packages_apps_GmsCore/blob/master/play-services-core/src/main/java/com/google/android/gms/common/security/ProviderInstallerImpl.java#L24

Could you implement this?

Here is a google library for it:
https://github.com/google/conscrypt

And some implementation which may be useful https://github.com/maks/MGit/blob/a5f0b5d73b35a73eaa2c8870a7cb62efa32e2060/app/src/main/java/com/manichord/mgit/transport/SSLProviderInstaller.java

voidstarstar:
This is an implementation of the method `ProviderInstallerImpl` to help protect against SSL exploits.  A new dependency on conscrypt was added and it should be updated periodically.  Minimal testing was done on an unofficial build of CM13, but further testing is welcome.

Fixes #504 and fixes #72.

I tested Hangouts it no longer appears to crash on startup, even without the workaround https://github.com/microg/android_packages_apps_GmsCore/issues/72#issuecomment-303714674.

Issue #86 calls this code so it should probably be tested.  I have not tested Gmail Sync, but I suspect this only fixes one of multiple problems with it.

Other Google and non-Google apps may be affected as well if they attempted to [update the security provider](https://developer.android.com/training/articles/security-gms-provider).

Note: This is not a full implementation since a couple things are missing, but it seems to be enough to work.  For example, the default `HostnameVerifier` is not set, which I think Google added to prevent certain attack vectors.  Another thing missing is a test for the `conscrypt_gmscore_jni` version.  These should be added in the future for better security and stability.

### DIFF
--- a/play-services-core/build.gradle
+++ b/play-services-core/build.gradle
@@ -35,6 +35,8 @@ dependencies {
     implementation project(':vtm-extras')
     implementation project(':vtm-jts')
     implementation project(':vtm-microg-theme')
+
+    compile 'org.conscrypt:conscrypt-android:2.0.0'
 }
 
 def execResult(...args) {
--- a/play-services-core/src/main/java/com/google/android/gms/common/security/ProviderInstallerImpl.java
+++ b/play-services-core/src/main/java/com/google/android/gms/common/security/ProviderInstallerImpl.java
@@ -18,9 +18,11 @@ package com.google.android.gms.common.security;
 
 import android.content.Context;
 import android.util.Log;
+import org.conscrypt.OpenSSLProvider;
+import java.security.Security;
 
 public class ProviderInstallerImpl {
     public static void insertProvider(Context context) {
-        Log.d("ProviderInstallerImpl", "yep, i should do something with Security here...");
+        Security.insertProviderAt(new OpenSSLProvider(), 1);
     }
 }


F4uzan:
Tested the commit with Hangouts, everything seems to be working a-okay in the app. Notifications went through and sign-in works as expected.

ale5000-git:
@mar-v-in
Hi,
is it possible to give priority to these pull requests?

#635 (ProviderInstallerImpl - very important)
#555 (Cast)
#624 (Google Suite accounts)
#628 (Improve version code generation)

mar-v-in:
Hi @ale5000-git, Hi @voidstarstar 

There is a good reason why this PR is not going to be merged as is, but I won't share this reason publicly. If any of you want to know the reason, please contact me in private.

My plan is to create a similar change though.

ale5000-git:
I have written to you on XDA.

voidstarstar:
I also sent a PM on XDA.

I wish GitHub still had a PM feature. :(

voidstarstar:
@mar-v-in 

I updated the code to a commonly used one-liner that simply inserts the Conscrypt provider as the provider with highest preference.  It is also now using the latest version of Conscrypt as @ale5000-git requested.

I retested and it seems to still work fine, but further testing is always encouraged.

The new library appears to be released under the Apache License 2.0.  It should be noted that this library was created by Google, so we we should be extra careful when deciding if it can be safely merged while still adhering to the ethos of the microG project.

ale5000-git:
@voidstarstar: I have a doubt, how can be all the code that there was before be replaced with a single line of code?

ale5000-git:
@mar-v-in: I wonder if the eventual "throws" must be catched.

```
public static int insertProviderAt(Provider provider, int position)
Returns:
    the actual preference position in which the provider was added, or -1 if the provider was not added because it is already installed.
Throws:
    NullPointerException - if provider is null
    SecurityException - if a security manager exists and its SecurityManager.checkSecurityAccess(java.lang.String) method denies access to add a new provider
```

mar-v-in:
@ale5000-git both possible exception cases can't happen in our case: provider can't be null (is created just there) and security manager is not used on Android

ale5000-git:
@mar-v-in: I'm not so used to Java but can't a "new" return null if there isn't enough free RAM?

mar-v-in:
@ale5000-git no, Java can throw an OutOfMemoryError on object creation, but it would probably be a bad idea to catch those, as the calling app would need to properly act in such cases.

That said, I don't know if OutOfMemoryError happens on Android at all, I guess ART runtime would just panic and crash the app.

ale5000-git:
In this case is fine but it still would be nice to add return value to logcat (for future debugging of eventual problems).

Namnodorel:
@mar-v-in As somebody who is developing a resource intensive app, I can confirm that Android very much throws OutOfMemoryExceptions. This ends the thread the exception occours in, not necessarily the whole app.

ale5000-git:
@mar-v-in: Isn't possible in a different thread to detect if this action has failed and eventually retry later?

ale5000-git:
@mar-v-in

There are also these pull requests for FakeStore:
https://github.com/microg/android_packages_apps_FakeStore/pull/16 (Add CHECK_LICENSE permission in FakeStore) => this being the most important
https://github.com/microg/android_packages_apps_FakeStore/pull/21 (Enabling direct ROM build of FakeStore and update build tools)

mar-v-in:
I don't think Google's original code would catch the OOM either, so we don't need to handle it either. If apps want to, they obviously can catch this case themselves.

Also Android will kill background apps and services if necessary for memory usage of foreground apps, so I doubt we'll see any problems with this

ale5000-git:
@mar-v-in

There are also these pull requests for FakeStore:
https://github.com/microg/android_packages_apps_FakeStore/pull/16 (Add CHECK_LICENSE permission in FakeStore) => this being the most important
https://github.com/microg/android_packages_apps_FakeStore/pull/21 (Enabling direct ROM build of FakeStore and update build tools)

kaeptmblaubaer1000:
Hi, can I please know what was the original implementation of this PR and @mar-v-in what were your reasons for not using it?

voidstarstar:
Would checking the returned value and logging it as an error if it's -1 be sufficient?  Let me know if you'd like me to open https://github.com/voidstarstar/android_packages_apps_GmsCore/commit/bc242caca30bcf40bad64fdf6399667d90eb3c44 as a new PR.

This is untested, but very straightforward.

kaeptmblaubaer1000:
I don't think it's as advanced as the original Google code. It checks if the return value is neither 1 nor -1, and throws an exception instead. In the case of 1 it even does some much more advanced things than this. It might be possible though that I looked at an outdated version of the code.

voidstarstar:
@kaeptmblaubaer1000 A 1 or -1 are the only two possible return values of insertProviderAt since we pass in a position of 1.

The two documented exceptions do not apply.  We do not pass a null `Provider`.  As mentioned above, an unchecked exception like `OutOfMemoryError` may occur, but we shouldn't handle that here.  We also should not get a SecurityException since Android does not have a `SecurityManager`.  The documentation states ["Application developers can assume that there's no SecurityManager installed"](https://developer.android.com/reference/java/lang/SecurityManager)

I'm not sure what you mean by "advanced".  Remember that we don't know what the original implementation actually looks like as it's closed source.  Any code you're looking at is likely using ugly reflection or obfuscation and probably isn't especially useful or indicative of what good code should look like.  Our goal should be to match the API specs as closely as possible, not the implementation.

ale5000-git:
@mar-v-in: If possible it would be nice to find a way to fix this at microG level (despite the problem affect also Google Play Service).

Look:
> Calling ProviderInstaller.installIfNeeded
> ...
> Originally, we thought we were done after doing this step. However, after further testing, we noticed something peculiar. Many devices succeeded in updating the security Provider via installIfNeeded, but were still throwing SSLHandshakeExceptions when the server required TLS 1.2!
> 
> Our initial understanding of OkHttp’s HTTPS strategy was that it uses the highest-installed version of TLS, and it falls back to older protocols if newer ones are not installed. This is false.
> 
> OkHttp’s default behavior is to use an SSLSocketFactory to create SSLSockets with default parameters, and select from protocols that are enabled for a given SSLSocket. Here’s the kicker: The latest protocols installed on a device are not necessarily enabled on its default SSLSockets!
> 
> To fix this, we need to override OkHttp’s SSLSocketFactory to enable TLS 1.2 on all SSLSockets

Ref: https://medium.com/tech-quizlet/working-with-tls-1-2-on-android-4-4-and-lower-f4f5205629a

kaeptmblaubaer1000:
Look at what I did in MGit, it might fix the issue here too...
Yes, the origihal code uses reflection but AFAIK this is needed to access private fields, isn't it?
The most important thing is that it works for MGit at least.

voidstarstar:
> Look at what I did in MGit, it might fix the issue here too...
Yes, the origihal code uses reflection but AFAIK this is needed to access private fields, isn't it?
The most important thing is that it works for MGit at least.

@kaeptmblaubaer1000 

[For reference](https://github.com/maks/MGit/blob/master/app/src/main/java/com/manichord/mgit/transport/SSLProviderInstaller.java)

I actually did happen to come across MGit while working on this patch, but I didn't realize you were the one that committed this file (about 1 year ago).

Do you have an xda account where we can [chat privately](https://forum.xda-developers.com/private.php?do=newpm&u=7841795)?  If not, can we talk on [#microg on freenode](http://webchat.freenode.net/?channels=microg)?

ale5000-git:
@mar-v-in 

See the relevant changes from MGit project:
https://github.com/maks/MGit/pull/302/files
https://github.com/maks/MGit/pull/272/files
https://github.com/maks/MGit/pull/340

PR #572, opened 2018-07-09
Temporarily fix broadcasts not being properly sent

voidstarstar:
This allows apps without the *.permission.C2D_MESSAGE to receive the broadcast

This is an implementation of the fix described [here](https://github.com/microg/android_packages_apps_GmsCore/issues/259#issuecomment-270035886).

I have not tested this code and it may contain bugs or not work.  The motivation is to fix #444, #259, and potentially many other apps that do not use their package name as required by the GCM spec but that still work with the official Google implementation.

It should be considered a temporary fix since this completely removes the required permission rather than fixing it.  A more permanent solution might be to look for and save any permission matching `*.permission.C2D_MESSAGE` when the app is installed.  This could then be used later when a broadcast is sent.

Note: This may be a privacy or security risk as it allows any receiver to potentially consume the broadcast since the permission is no longer checked.

PR #637, opened 2018-10-09
Set package of intent prior to broadcasting

voidstarstar:
This should prevent an intent from being broadcast to registered receivers that do not match the specified package.

This PR might prevent issues if one installed app causes [`receiverPermission` to be null](https://github.com/microg/android_packages_apps_GmsCore/blob/f94d140dea71ee5418aa94861931445f67b3968f/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L478) while another app does not use an [`IntentFilter`](https://developer.android.com/reference/android/content/IntentFilter).

From the documentation on [registerReceiver](https://developer.android.com/reference/android/content/Context#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter)):
> As of `Build.VERSION_CODES.ICE_CREAM_SANDWICH`, receivers registered with this method will correctly respect the `Intent.setPackage(String)` specified for an Intent being broadcast. Prior to that, it would be ignored and delivered to all matching registered receivers. Be careful if using this for security.

This PR seems like a good addition after #572 (https://github.com/microg/android_packages_apps_GmsCore/commit/f94d140dea71ee5418aa94861931445f67b3968f).  With my minimal testing I have not seen this patch cause any regressions, but I have not seen it fix any problems either.  Further testing should be done before merging.

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
+++ b/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
@@ -485,6 +485,7 @@ public class McsService extends Service implements Handler.Callback {
             for (ResolveInfo resolveInfo : infos) {
                 logd("Target: " + resolveInfo);
                 Intent targetIntent = new Intent(intent);
+                targetIntent.setPackage(msg.category);
                 targetIntent.setComponent(new ComponentName(resolveInfo.activityInfo.packageName, resolveInfo.activityInfo.name));
                 sendOrderedBroadcast(targetIntent, receiverPermission);
             }


voidstarstar:
Oops, looks like this was already being done and I somehow just missed it.

The package of the intent was set [here](https://github.com/microg/android_packages_apps_GmsCore/blob/1e444dc4aaae56c381e3e6621f8c0d188963ca03/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L460) and a copy constructor was used [here](https://github.com/microg/android_packages_apps_GmsCore/blob/1e444dc4aaae56c381e3e6621f8c0d188963ca03/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L487).  Seems like my code is redundant and would explain why I didn't see any change, so I'm closing this PR.  Sorry about the confusion.

PR #646, opened 2018-10-18
bump spoofed Play Services Version to 14.3.66

Nanolx:
Solves the 'updating Play Support libraries' nag (Play Services in disguise) and the caused mobile data drain users are experiencing when using official or patched Play Store due the current official release spoofing the older version 13.2.80.

### DIFF
--- a/play-services-core/build.gradle
+++ b/play-services-core/build.gradle
@@ -46,7 +46,7 @@ def execResult(...args) {
     return stdout.toString().trim()
 }
 
-def gmsVersion = "13.2.80"
+def gmsVersion = "14.3.66"
 def gmsVersionCode = Integer.parseInt(gmsVersion.replaceAll('\\.', ''))
 def gitVersionBase = execResult('git', 'describe', '--tags', '--abbrev=0').substring(1)
 def gitCommitCount = Integer.parseInt(execResult('git', 'rev-list', '--count', "v$gitVersionBase..HEAD"))


Nanolx:
Needs 14.3.69 now, as it will otherwise still complain.

Vavun:
Maybe it will be better to use version of beta play services for future releases

DynaSpan:
@Nanolx that should be possible. I've wanted to try that for a while, unfortunately don't have the time yet.. I will try to setup a PoC soon.

ale5000-git:
I had a similar idea in the past (but never written).
It is really complicate to do but very good in the long term.

com.android.google.gms:
- Very small in size
- Reside in /system/priv-app
- Require minimal permissions since most of the code isn't here
- Use signature spoofing
- Receive requests from apps and do nothing more than proxy them to org.microg.gmscore
- Bind as location provider directly but really do nothing more than proxy requests to org.microg.gmscore (it is important so if org.microg.gmscore is crashing then can be restarted without loss system registration as location provider that require restart)

org.microg.gmscore:
- Most of the code go inside here
- Can be installed as user app
- Most permissions are requested here
- Do NOT use signature spoofing
- With a unique name and without signature spoofing can be perfectly installed and updated fon F-Droid

mar-v-in:
The split would indeed be possible for most, but not all APIs (and all versions). For example, some versions (not sure if still the case) of the Maps API required classes to be present in the `com.google.android.gms` package.

The other problematic thing that I see: The version number issue is still present because the version number of `com.google.android.gms` is what matters, so you still need to update this app.

The reason why I keep the version number "low" is to ensure compatibility. The beta versions of play services implement API updates that are not yet published, and by using a beta version number, gmscore would announce compatibility with those. Announcing compatibility without being compatible can cause worse effects (corrupt databases, unusable apps) than being just incompatible.

ale5000-git:
The part that contain the version would still need to be unofficially recompiled and put in /system/priv-app but it would be faster since it contain less code and instead the big part with most of the code could be kept official and updated from the main repo of F-Droid (hopefully).

ale5000-git:
Addiotionally if GCM registration are kept in the data of org.microg.gmscore, one can replace com.android.google.gms without lost them.

Nanolx:
@ale5000-git about this one:

```
org.microg.gmscore:
* Do NOT use signature spoofing
```

for the (modified) Play Store to allow (in-)app-purchases with microG on custom ROMs and some stock ROMs the signature of the GmsCore and Play Store needs to be identical, so I'm not 100 % sure that would work that way, if Android sees the Google Account created by org.microg.gmscore with a different signature as com.android.vending...

ale5000-git:
@Nanolx

`org.microg.gmscore` would be the one handling the account but request pass throw `com.google.android.gms` before reaching the Play Store so it don't see `org.microg.gmscore` at all.

`com.google.android.gms` should only be the man-in-the-middle that passthrough requests.
`org.microg.gmscore` do the real work but noone see it or know its existence.

Nanolx:
@ale5000-git

I thought so, but wasn't sure. Thanks.

@Mar-V-In

I do see the problem when you "simply" increase the version number. But there's also an issue when **not** increasing it: apps stop working. 

Also Google enforces GmsCore and Play Store updates now much more than before, several apps much quicker complain about the old version.

So we look at the most stupid of a reciprocal issue, by fixing one, you cause the other and vice-versa.

That's one (or the major) reasoning behind this split request:
* official microG spoofs a version that is actually supported, preventing the issue that microG GmsCore reports compatibility for a higher version then where's none
* unofficial versions / users can increase the spoofed version, preventing the issue that apps stop working, because of the low version number

mar-v-in:
@Nanolx I hardly see the reason for the split with your argumentation: If users do unofficial builds that have a higher version number than supported, they can also do the same for GmsCore. There is no difference with split vs non-split, in both cases you use a custom build of the app that has the package name `com.google.android.gms`.

Nanolx:
Well, it's your decision.

Since the pull request is redundant and obsolete I'm closing it now.

patrickdrd:
I would agree with @Nanolx, it's a minor issue but it is ugly, a distraction I would call it for users

Issue #491, opened 2018-02-27
Unable to add Google account


ClayMaster:
MicroG Services Core is 0.2.4-108.
Wher I try to add Google account:

![screenshot_2018-02-27-21-46-05](https://user-images.githubusercontent.com/36898345/36748355-add7a9b8-1c08-11e8-9e2b-6df2b872a33d.png)

Both buttons "Cancel" and "Sign in" are inactive and cannot be pressed.
Tried on some custom ROMs 4.2.1 and 4.4.2 with same result.

ale5000-git:
Duplicate of #368

ale5000-git:
It should be fixed in the latest code (it isn't yet released).
Ref: https://github.com/microg/android_packages_apps_GmsCore/commit/ceb7e721b1882b21814cb4333f77f93c0c13ed51

ale5000-git:
Bugs is fixed.
New version is out now.

PR #717, opened 2019-03-01
Add version check to skip first screen on KK (login fix)

Nanolx:
Fixes for the login screen issue ("Login" button not clickable) by @PorygonZRocks - reportedly works.

(I know it's not common for a thirdparty to send in a pull request, but still)


PorygonZRocks:
Afaik this never got tested, so I *think* it should work but am not sure. I see you said it "reportedly works" so I assume someone did end up testing it?

Nanolx:
It's part of my custom microG build (included along side the official one in my NanoDroid project) and two or three (known) KK users stopped their login-complains and reported "everything works now".

I don't have a KK device on my own, so I relying on user feedback for this one.

Vavun:
> 
> 
> Afaik this never got tested, so I _think_ it should work but am not sure. I see you said it "reportedly works" so I assume someone did end up testing it?

I can confirm that it works without problems. 
10+ people on russian community confirmed that microg fork with this fix solved their login problem on kitcat or lower and no one reported problems with login issues on KK after that. (1000+ fork downloads)

voidstarstar:
Fixes #491 and fixes #368.  These were not linked from this PR, so they need to be manually closed once verified.

PR #719, opened 2019-03-01
microG Settings in AppInfo and SystemSettings (+ adaptive icon)

Nanolx:
Show microG Settings not just in Launcher, but also in

* AppInfo (System Settings > Apps > All Apps > microG Services Core; depending on the ROM version either a gear icon will appear or a new entry for additional settings)
* Device Settings (System Settings > Device (or corresponding category, depending on the ROM version))

Also includes Adaptive icon and updated strings.xml for C, de and ru.

This commit is the result of manually squashed commits by me and @Vavun found in my fork's master branch. Reportedly working according to all testers accross different Android versions.

[I'm really sorry... but I wanted a clean pull request but somehow I've managed to mess up submodules twice]

Issue #413, opened 2017-08-15
Can't add google account on alternative android implementation.


V10lator:
I'm using a Jolla C ( https://jolla.com ) which uses aliendalvik ( http://www.myriadgroup.com/en/products/device-solutions/mobile-software/alien-dalvik/ ). On top of that I installed microg as described here: https://www.joachim-breitner.de/blog/713-microG_on_Jolla

This combinations works just fine except for one issue: I have an app which requires a google account linked to the play services. Now I read that this should be doable from android settings but the Jolla / aliendalvik is missing said setting. That said I think it would be great if microg would offer a way to add an account in its own settings.

//EDIT: Forgot to say: In the past I used the original Google Play Services on the same phone and they offered a way to add an account. Anyway, I don't want to use them anymore so I hope this feature request will make it. :)

erikderzweite:
I'm using Sailfish X (Sailfish OS on Sony Xperia X). Have installed microg as well using [this tutorial](http://talk.maemo.org/showpost.php?p=1509945&postcount=85). 

Got signature spoofing to work, but adding account is still not possible with alien dalvik. It would be very nice to have this option in microG directly. I know that on Android it works without (previous phone is Oneplus One with LineageOS and microG), but it would be a very nice feature to have in other systems as well.

ale5000-git:
@V10lator
@erikderzweite
Try to install microG with PlayStore and add an account through PlayStore, after it you can remove PlayStore and reinstall FakeStore if you want.

erikderzweite:
That did work, had to downgrade microg to the stable 0.2.4, but then it worked despite several errors shown.

hevanaa:
I have latest microG and have the same problem. I removed FakeStore and installed PlayStore. But when trying to start PlayStore I get the same "microG Services Core" dialog with a SIGN IN button that doesn't work. Otherwise latest microG works great.

hevanaa:
@voidstarstar Thanks, sounds exactly right from the description. The Android environment in Alien-Dalvik is 4.4. Hopefully we get an updated version at some point. I need to try an OTG cable (should have one somewhere from my N900 days).

hevanaa:
This worked fine with an OTG USB keyboard. I didn't have to install PlayStore. I tried with the Youtube app and was able to add my account.

ale5000-git:
Duplicate of #368

It should be fixed in the latest code (it isn't yet released).
[...]

PR #727, opened 2019-03-12
Allow to add a Google account and test location from microG settings (by @Vavun), ref: #719 #413

ale5000-git:
Allow to add a Google account and test location from microG settings (by @Vavun), ref: #719 #413

**Need testing.**

---
This change is [Reviewable](https://reviewable.io/reviews/microg/android_packages_apps_gmscore/727)


Vavun:
Nice but I didnt test it at all.
It was a part of my old microg mod and I copypaste it only as example =)

Nanolx:
I'm currently running a non-published build of my fork (`0.2.6.15180-dirty-194`) which contains this pull request and I can confirm it works as advertised.

Though maybe the geolocation part should be moved to uNlp, so that it is also available in the stand-alone version? Though, I don't know if that actually has a widespread user base (I _assume_ not). @mar-v-in ?

ale5000-git:
Closed in favour of #735.

@Nanolx: Please re-test it if possible

ale5000-git:
@Nanolx: For the "geolocation part should be moved to uNlp" it is a good idea but since the PR #735 is already ready I think it is better to do the change in a separate PR.

PR #732, opened 2019-03-18
Temporary whitelist support for Doze

### LINKED ISSUE #338: Put app into temporary whitelist when high priority gcm received [$5]
On android M and above and google's implementation, apps will be woken up and put into temporary whitelist by gcm priority message. This is done by addPowerSaveTempWhitelistApp() in IDeviceIdleController.aidl. But in microG there is no such setting, so it leads to some delay in notification. I've experience 30 minutes delay in whatsapp message even microG received from google server on time .

The relevant code is here:
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/IDeviceIdleController.aidl
https://android.googlesource.com/platform/frameworks/base/+/master/services/core/java/com/android/server/DeviceIdleController.java
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/content/Context.java#3367
https://android.googlesource.com/platform/frameworks/base/+/master/core/res/AndroidManifest.xml#2722

Hope microG could also do this.

voidstarstar:
This contains commits by @ccaapton @ale5000-git and myself.

Tested was done in a single user and multi-user setting.  Further testing is welcome.

This will likely fix issues many users are having involving delayed push notifications.  Some users might have even reported that their push notifications were not working at all when in fact Doze was really the issue.

This fixes #338.


Vavun:
Hey @Nanolx , could you please make a test build of your fork with this PR?

Any way, it's time to bump version to 15.1.80

Nanolx:
I'll roll-out a new custom test build later today.

Nanolx:
@Vavun uploaded custom microG build `0.2.6.15180-dirty-192` alongside with patched Play Store `14.0.33` to my F-Droid repo. This build of Play Store does not contain the anti-update patch as Google has changed around stuff and it's late here.

voidstarstar:
@Nanolx Thanks.  [Managed profiles/work profiles](https://developer.android.com/work/managed-profiles) remain untested.  Any feedback, especially if someone happens to use managed profiles, is appreciated.

Nanolx:
@voidstarstar I'll leave a message over at XDA for people to test.

From my experience I've had three "problem apps" since some months now, which I had push notification issues with (namely XDA Labs, Animal Crossing: Pocket Camp and Bandcamp), I have not yet unininstalled and re-installed them, will report later (and also check the logs).

For the others I can confirm that push messages reliably appear in time when device is in Doze, as before there's been kind of delay for some.

ale5000-git:
@Nanolx: Uninstall isn't needed, you can use [this code](https://github.com/micro5k/microg-unofficial-installer/commit/3ccb91d385ffbbaa217d069a717aa99d69c984f4) to force apps to refresh the GCM registration.

chris42:
@ale5000-git Couldn't that be included in Microg itself? So many people ask on how to reset or mess it up after an upgrade?

ale5000-git:
@chris42
Well, this code delete a part of app data from all apps, it works fine from recovery but not from a normal app (microG is an app after all).

Also this is needed just one time.

voidstarstar:
Would it be possible for an app running as root to do that?  It would be nice if this was done on first launch of microG or even manually via a GUI button within the microG settings.

Nanolx:
An app running as root would technically be possible to do so, yes, but it's questionable whether something like this would be included.

ale5000-git:
If something this is included, it may led antivirus app to flag it ad virus, if really neede I think a separate app is better.

Nanolx:
I've added it to my `nanodroid-util` script, see https://gitlab.com/Nanolx/NanoDroid/commit/445ce81d066ea45e4303f327063704d1eb15b220

Providing it as a script is better I suppose.

Vavun:
I think this commit can be improved with this
https://developer.android.com/training/articles/direct-boot?hl=EN

Vavun:
Hello again. One user from russian 4pda community reported problems with push messages on nanolx microg fork starting 192 build and newer. (167 works fine)

the only difference in the way messages work is this pull request (I did not miss anything?)
And to be honest, it does not seem that the problem is in these changes.
But other changes are either not related to the push notifications at all or just cosmetic fixex.

Build 167, push notification tester run
https://www.youtube.com/watch?v=cfm7-0pyVKo

Build 192/227/241, push notification tester run
https://www.youtube.com/watch?v=J0fAcPiaykA

[logcat_167.txt](https://github.com/microg/android_packages_apps_GmsCore/files/3054158/logcat_167.txt)
[logcat_241.txt](https://github.com/microg/android_packages_apps_GmsCore/files/3054159/logcat_241.txt)

I tried to figure out something but I found nothing useful except
`I/O error: java.net.SocketException: socket is closed`

PS

@Nanolx it's time to enable issues on your fork page =)
It would be better if this issue was placed there.

voidstarstar:
@Vavun Are you able to reliably reproduce this?

Did you reinstall the Push Notification Tester app after installing the new version of GmsCore?

I see a `SERVICE_NOT_AVAILABLE` error.  I think the issue is probably related to this.

I'm also wondering if the device was not registered yet.  Did Google Cloud Messaging say it was connected before you opened the Push Notification Tester app?

Is the Google Cloud Messaging -> Advanced -> Confirm new apps setting enabled or disabled?

I could be wrong, but I don't think this is caused by this PR.

Nanolx:
I can't reproduce this either. It works perfect for me with build 241. Any adblocker or DNS denying access to mtalk.google.com?

Am 8. April 2019 16:41:11 MESZ schrieb voidstarstar <notifications@github.com>:
>@Vavun Are you able to reliably reproduce this?
>
>Did you reinstall the Push Notification Tester app after installing the
>new version of GmsCore?
>
>I see a `SERVICE_NOT_AVAILABLE` error.  I think the issue is probably
>related to this.
>
>I'm also wondering if the device was not registered yet.  Did Google
>Cloud Messaging say it was connected before you opened the Push
>Notification Tester app?
>
>Is the Google Cloud Messaging -> Advanced -> Confirm new apps setting
>enabled or disabled?
>
>I could be wrong, but I don't think this is caused by this PR.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/microg/android_packages_apps_GmsCore/pull/732#issuecomment-480860648

-- 

Nanolx:
@Vavun I've enabled 'Issues' over at my fork, but as stated earlier: I don't experience any issue with 0.2.6.15180-dirty-241

Vavun:
> Are you able to reliably reproduce this?

> I can't reproduce this either. It works perfect for me with build 241

Thank you very much, I also can't reproduce this problem.
But that's wierd that he has problem with latest builds and has no problem with 167 build.

> I'm also wondering if the device was not registered yet. Did Google Cloud Messaging say it was connected before you opened the Push Notification Tester app?
> 
> Is the Google Cloud Messaging -> Advanced -> Confirm new apps setting enabled or disabled?

GCM enabled and Confirm new apps disabled in both cases.

@voidstarstar @Nanolx I will give your responses to user with this issue. Thanks.

julienschmidt:
Ping @mar-v-in. This PR seems ready to merge, too.
If so, could we also have a new release? This, #624 and #635 should fix a lot of things for many users.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -166,6 +168,26 @@ public void onCreate() {
...
         powerManager = (PowerManager) getSystemService(POWER_SERVICE);
+        if(mDeviceIdleController == null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            try {
+                java.lang.reflect.Method method = Class.forName("android.os.ServiceManager").getMethod("getService", String.class);
+                java.lang.reflect.Field field = Context.class.getField("DEVICE_IDLE_CONTROLLER");
+                IBinder binder = (IBinder) method.invoke(null, field.get(null));
+                if(binder != null)
+                    mDeviceIdleController = IDeviceIdleController.Stub.asInterface(binder);
+            } catch (NoSuchFieldException e) {
+                e.printStackTrace();
```
mar-v-in: You shouldn't use `e.printStackTrace()` as it logs to stdout which shouldn't be used on Android. Use `Log.w(TAG, e)` instead.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -485,6 +507,24 @@ private void handleAppMessage(DataMessageStanza msg) {
             for (ResolveInfo resolveInfo : infos) {
                 logd("Target: " + resolveInfo);
                 Intent targetIntent = new Intent(intent);
+                if(mDeviceIdleController != null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                    try {
+                        java.lang.reflect.Method method = android.os.UserHandle.class.getMethod("getUserId", int.class);
+                        int userId = (int) method.invoke(null, getPackageManager().getApplicationInfo(msg.category, 0).uid);
+                        logd("Adding app " + msg.category + " for userId " + userId + " to the temp whitelist");
+                        mDeviceIdleController.addPowerSaveTempWhitelistApp(resolveInfo.activityInfo.packageName, 10000, userId, "GCM Push");
```
mar-v-in: As the only reason for importing the hidden AIDL from AOSP, is to call a single method, why not use reflection for the invocation (as you use reflection anyway) and thus get rid of the whole AIDL?

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -166,6 +168,26 @@ public void onCreate() {
         heartbeatIntent = PendingIntent.getService(this, 0, new Intent(ACTION_HEARTBEAT, null, this, McsService.class), 0);
         alarmManager = (AlarmManager) getSystemService(ALARM_SERVICE);
         powerManager = (PowerManager) getSystemService(POWER_SERVICE);
+        if(mDeviceIdleController == null && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            try {
+                java.lang.reflect.Method method = Class.forName("android.os.ServiceManager").getMethod("getService", String.class);
```
mar-v-in: Don't use fully qualified name but import instead...

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
ale5000-git: This is my fault since this PR is based on my commit and my knowledge of reflection is minimal.

ale5000-git:
@mar-v-in: It would be nice to have #741 before new release for the Sailfish users.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
voidstarstar: Isn't using an AIDL generally preferred over reflection?  The reflection used here is only done because `@hide` was used in the Android API.  Shouldn't we limit our use of reflection to only scenarios that absolutely require it?  I can swap the AIDL for reflection, but I just wanted to make sure this is really what is desired.

(As a side note: I think we might even be able to [avoid using reflection completely](https://stackoverflow.com/questions/7888191/how-do-i-build-the-android-sdk-with-hidden-and-internal-apis-available), but it would require compiling against a modified `android.jar` instead of the standard SDK.  This would make the code a bit cleaner, but the build process a bit more complicated.  So far it doesn't seem like microG needs to access the hidden API very often, but if that changes as more features are implemented, this is probably one of the rare cases that it would be acceptable since we need to mimic Google.)

mar-v-in: @voidstarstar the problem with importing the AIDL directly, is that it is less likely to survive an android update than reflection. Remember that hidden APIs might be changed without prior notice even in minor updates. Also, once you have an incompatibility between android versions, it's way easier to handle it with reflection than having two AIDL files for the same interface name.

mar-v-in: If you want to explore the path of using an extended standard library (modified `android.jar`) check out the `unifiednlp-compat` module and its usage in `unifiednlp-base`. This is likely to work here as well, but you have to be careful with cross version compatibility again.

voidstarstar:
@mar-v-in Thanks for the feedback.  I made changes based on the review above.  Please let me know if I missed anything or if these should be squashed.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -485,6 +487,25 @@ private void handleAppMessage(DataMessageStanza msg) {
...
+                                .getMethod("getService", String.class).invoke(null, field.get(null));
+                        if (binder != null) {
+                            Object deviceIdleController = Class.forName("android.os.IDeviceIdleController$Stub")
+                                    .getMethod("asInterface", IBinder.class).invoke(null, binder);
+                            int userId = (int) UserHandle.class.getMethod("getUserId", int.class)
+                                    .invoke(null, getPackageManager().getApplicationInfo(msg.category, 0).uid);
+                            logd("Adding app " + msg.category + " for userId " + userId + " to the temp whitelist");
+                            deviceIdleController.getClass()
+                                    .getMethod("addPowerSaveTempWhitelistApp", String.class, long.class, int.class, String.class)
+                                    .invoke(deviceIdleController, resolveInfo.activityInfo.packageName, 10000, userId, "GCM Push");
```
mar-v-in: It's `resolveInfo.activityInfo.packageName` here and `msg.category` three lines above. Better be consistent.

Or, maybe even add a `String packageName = msg.category;` and use that instead to make it more clear what it is. Also `msg.category` should be the same as `resolveInfo.activityInfo.packageName` anyway...

voidstarstar: Good point.  Consistency would be important.  I made the change you suggested.

voidstarstar: I just noticed that `resolveInfo.activityInfo.packageName` is used [a few lines below here](https://github.com/microg/android_packages_apps_GmsCore/blob/ee1910acc42bcf5b88d6e1154ae3c160fe1fdc2d/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L510).  Should I just make this line use the `String packageName` too?

If so, should it use `resolveInfo.activityInfo.packageName` instead of `msg.category`?  That might make more sense due to the scope.

mar-v-in: I guess for the last occurence, it might make sense to use `resolveInfo.activityInfo.packageName`, as it also uses `resolveInfo.activityInfo.name`.

If you move the `String packageName =` further up (in the first line of that method), you could replace all occurences of `msg.category` with `packageName`. Even better, as I bet it isn't super obvious that `msg.category` is defined as the package name.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -485,6 +487,26 @@ private void handleAppMessage(DataMessageStanza msg) {
             for (ResolveInfo resolveInfo : infos) {
                 logd("Target: " + resolveInfo);
                 Intent targetIntent = new Intent(intent);
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
mar-v-in: Another minor: it would make sense to use the `app.wakeForDelivery` setting as a restriction whether to add the app to the temp whitelist or not, no?

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
voidstarstar: Ok, I left the last occurrence unchanged and I changed all occurrences of `msg.category` to `packageName`.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
voidstarstar: That sounds reasonable to me.  The current description of this option sounds like what the whitelist is intended to do.

> Start the app in background to receive the incoming push messages. Disabling this might delay message delivery up to the moment the app is started or cause complete message loss.

Currently it seems like this option was only used to send the broadcast to [an app that was stopped](https://developer.android.com/reference/android/content/Intent.html#FLAG_INCLUDE_STOPPED_PACKAGES).

ale5000-git:
@voidstarstar
@mar-v-in

> Looking up a class via reflection is, by magnitude, more expensive.

https://stackoverflow.com/a/435568/490687

This is why I have placed some code in onCreate(), the latest code do everything in handleAppMessage().
What do you think?

voidstarstar:
It's definitely much more expensive on the Oracle JVM and I can only assume it would be at least as much of a performance hit on Dalvik/ART.

I think calling `Method.invoke()` will still need to be done in `handleAppMessage()` in order to find the `userId` and to actually call the `addPowerSaveTempWhitelistApp()` method, but the rest can probably be done once in `onCreate()` by reusing the `Method` Objects.

voidstarstar:
I created a new commit https://github.com/voidstarstar/android_packages_apps_GmsCore/commit/627d8abe9f3a778f1a9cf835780fa8e5076d6c93 that I think will move as much reflection as possible to `onCreate()`.  There are 3 fields initialized in `onCreate()` using reflection and just 2 calls to `invoke()` in `handleAppMessage()`.

If this looks good, should I open a new PR for this?

mar-v-in:
I don't think the performance penalty of using a bit more reflection is relevant here. But prefetching certainly doesn't hurt either.

Just open a new PR (as this one is merged already and cannot be reused) with a single commit with your changes.

Tentos:
I have successfully installed microg with temporary whitelisting. Great work!

I would like to suggest two documentation changes [on the installation wiki page](https://github.com/microg/android_packages_apps_GmsCore/wiki/Installation):

- It is necessary to install microg as a system app. This is quite obvious because Android does not grant the right for temporary whitelisting otherwise. (Still, I had to look at the logcat and see the `SecurityException`s to realize this.)
- Additionally, I had to add an XML with the following content to `/system/etc/permissions/` (named e.g. `microg.xml`):
```
<?xml version="1.0" encoding="utf-8"?>  
<permissions>
  <allow-in-power-save package="com.google.android.gms" />
</permissions>
```
The last bullet point was described by @ccaapton [here](https://github.com/microg/android_packages_apps_GmsCore/issues/338#issuecomment-399103776).
When the XML was missing, I again got a `SecurityException`:
```
12-22 23:16:37.699 26505 26542 W GmsGcmMcsSvc: java.lang.reflect.InvocationTargetException
12-22 23:16:37.699 26505 26542 W GmsGcmMcsSvc: 	at java.lang.reflect.Method.invoke(Native Method)
...
12-22 23:16:37.699 26505 26542 W GmsGcmMcsSvc: Caused by: java.lang.SecurityException: Calling app u0a73 is not on whitelist
12-22 23:16:37.699 26505 26542 W GmsGcmMcsSvc: 	at android.os.Parcel.readException(Parcel.java:1692)
```
10073 is indeed the ID of my microg installation. I had manually disabled battery optimization for microg before, but it was necessary to whitelist it on a system level by the XML to not get the `SecurityException`, i.e., to get a working temporary whitelisting.

I am obviously open to do the necessary changes in the wiki, I just wanted to ask whether my suggestions are correct (and welcome :-) )

drrossum:
I can confirm that this solved a long standing delayed-notifications issue for me as well, on Android 9.

@Tentos, thanks for pointing this out!

HolisticDeveloper:
I'm not that familiar with Doze as I'm a neophyte when it comes to Android development. Would the changes in this PR have any impact for an app that is always running in the foreground on an always-on device (no battery)? We have a driver assistance application running on custom hardware, so we don't have to worry about the device "sleeping" or different apps being in focus.

PR #741, opened 2019-03-26
Add Google account button to microG settings, fix #413

### LINKED ISSUE #413: Can't add google account on alternative android implementation.

ale5000-git:
Add Google account button to microG settings, fix #413

---
This change is [Reviewable](https://reviewable.io/reviews/microg/android_packages_apps_gmscore/741)


ale5000-git:
@Nanolx
@Vavun 

I have added the bare minimum for the Google account thing.
I hope it get merged before the new official version of microG.

PR #744, opened 2019-04-02
Fix app unregistration in Android 8.0+ and prevent a crash

### LINKED ISSUE #743: Getting crash sometimes on GCM page
Usualy I get this crash when doing something on Google Cloud Messaging page.
Don't know what the reason
Maybe @voidstarstar could be so kind to figure out the reason

This is nanolx fork, but I got the same on vanilla

```
04-01 20:46:48.429 I/ActivityManager(1332): START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.google.android.gms/org.microg.gms.ui.SettingsActivityLauncher bnds=[158,523][293,718]} from uid 10145
04-01 20:46:49.985 I/ActivityManager(1332): START u0 {cmp=com.google.android.gms/org.microg.gms.ui.GcmFragment$AsActivity} from uid 10033
04-01 20:46:50.057 V/ResourceType(2524): Too many attribute references, stopped at: 0x01010034
04-01 20:46:50.057 V/ResourceType(2524): Too many attribute references, stopped at: 0x01010034
04-01 20:46:50.325 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 556, previously 557
04-01 20:46:50.326 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 556, previously 557
04-01 20:46:50.326 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 234, previously 262
04-01 20:46:50.326 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 463, previously 467
04-01 20:46:50.326 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 291, previously 557
04-01 20:46:50.326 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 130, previously 528
04-01 20:46:50.349 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 440, previously 485
04-01 20:46:50.350 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 440, previously 485
04-01 20:46:50.351 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 266, previously 273
04-01 20:46:50.351 V/ResourceType(2524): ResTable_typeSpec entry count inconsistent: given 198, previously 681
04-01 20:46:50.390 I/zygote64(2524): Do full code cache collection, code=2014KB, data=1297KB
04-01 20:46:50.399 I/zygote64(2524): After code cache collection, code=1999KB, data=1172KB
04-01 20:46:50.620 I/Choreographer(2524): Skipped 36 frames!  The application may be doing too much work on its main thread.
04-01 20:46:50.687 V/chatty  (1332): uid=1000(system) batterystats-wo expire 6 lines
04-01 20:46:50.749 V/ANDR-PERF-MPCTL(502): Invalid profile no. 0, total profiles 0 only
04-01 20:46:52.187 V/ANDR-PERF-MPCTL(502): Invalid profile no. 0, total profiles 0 only
04-01 20:46:55.404 D/AndroidRuntime(2524): Shutting down VM
04-01 20:46:55.412 E/AndroidRuntime(2524): FATAL EXCEPTION: main
04-01 20:46:55.412 E/AndroidRuntime(2524): Process: com.google.android.gms, PID: 2524
04-01 20:46:55.412 E/AndroidRuntime(2524): java.lang.IllegalStateException: Cannot call this method while RecyclerView is computing a layout or scrolling android.support.v7.widget.RecyclerView{7b700d2 VFED.V... .F...... 0,0-720,1042 #7f080062 app:id/list}, adapter:com.takisoft.fix.support.v7.preference.PreferenceGroupAdapter@f9785a3, layout:android.support.v7.widget.LinearLayoutManager@d4dc0a0, context:android.view.ContextThemeWrapper@68bf459
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView.assertNotInLayoutOrScroll(Unknown Source:31)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$RecyclerViewDataObserver.onItemRangeChanged(Unknown Source:3)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$AdapterDataObservable.notifyItemRangeChanged(Unknown Source:18)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$Adapter.notifyItemChanged(Unknown Source:3)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.preference.PreferenceGroupAdapter.onPreferenceChange(Unknown Source:9)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.preference.Preference.notifyChanged(Unknown Source:6)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.preference.Preference.setSummary(Unknown Source:18)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at org.microg.gms.ui.GcmFragment$GcmAppPreference.updateViewDetails(Unknown Source:69)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at org.microg.gms.ui.GcmFragment$GcmAppPreference.onBindViewHolder(Unknown Source:0)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.preference.PreferenceGroupAdapter.onBindViewHolder(Unknown Source:4)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.preference.PreferenceGroupAdapter.onBindViewHolder(Unknown Source:2)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$Adapter.onBindViewHolder(Unknown Source:0)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$Adapter.bindViewHolder(Unknown Source:29)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(Unknown Source:40)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(Unknown Source:513)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.GapWorker.prefetchPositionWithDeadline(Unknown Source:14)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.GapWorker.flushTaskWithDeadline(Unknown Source:15)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.GapWorker.flushTasksWithDeadline(Unknown Source:22)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.GapWorker.prefetch(Unknown Source:3)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.support.v7.widget.GapWorker.run(Unknown Source:79)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.os.Handler.handleCallback(Handler.java:790)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.os.Handler.dispatchMessage(Handler.java:99)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.os.Looper.loop(Looper.java:164)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at android.app.ActivityThread.main(ActivityThread.java:6494)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at java.lang.reflect.Method.invoke(Native Method)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:440)
04-01 20:46:55.412 E/AndroidRuntime(2524): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
04-01 20:46:55.423 W/ActivityManager(1332):   Force finishing activity com.google.android.gms/org.microg.gms.ui.GcmFragment$AsActivity
04-01 20:46:55.436 I/ActivityManager(1332): Showing crash dialog for package com.google.android.gms u0
04-01 20:46:55.482 V/chatty  (1332): uid=1000(system) android.ui expire 5 lines
04-01 20:46:55.495 V/chatty  (1332): uid=1000(system) RenderThread expire 4 lines
04-01 20:46:55.927 W/ActivityManager(1332): Activity pause timeout for ActivityRecord{bab6ba3 u0 com.google.android.gms/org.microg.gms.ui.GcmFragment$AsActivity t14625 f}
04-01 20:47:00.483 D/AutofillManagerService(1332): Close system dialogs
04-01 20:47:00.485 W/ActivityManager(1332):   Force finishing activity com.google.android.gms/org.microg.gms.ui.GcmFragment$AsActivity
04-01 20:47:00.508 W/ActivityManager(1332):   Force finishing activity com.google.android.gms/org.microg.gms.ui.SettingsActivityLauncher
04-01 20:47:00.612 I/ActivityManager(1332): Killing 2524:com.google.android.gms/u0a33 (adj 100): crash
04-01 20:47:00.623 I/cr_ChildProcessService(3858): Destroying ChildProcessService pid=3858
04-01 20:47:00.633 W/ActivityManager(1332): Unable to start service Intent { act=com.android.launcher3.WINDOW_OVERLAY dat=app://ru.whatau.cpl:10145?v=7&cv=9 pkg=com.google.android.apps.nexuslauncher } U=0: not found
04-01 20:47:00.633 W/Looper  (1332): Dispatch took 149ms on android.ui, h=Handler (com.android.server.am.AppErrorDialog$1) {af50b48} cb=null msg=7
04-01 20:47:00.636 I//system/bin/webview_zygote64(3858): System.exit called, status: 0
04-01 20:47:00.636 I/AndroidRuntime(3858): VM exiting with result code 0, cleanup skipped.
04-01 20:47:00.651 I/ActivityManager(1332): Killing 3858:com.android.webview:sandboxed_process0/u0a33i0 (adj 0): isolated not needed
04-01 20:47:00.651 I/ActivityManager(1332): Killing 3174:android.ext.services/u0a24 (adj 906): empty #31
04-01 20:47:00.653 W/AppOps  (1332): Finishing op nesting under-run: uid 1000 pkg android code 24 time=0 duration=0 nesting=0
04-01 20:47:00.657 E/SurfaceFlinger(514): Failed to find layer (Application Error: com.google.android.gms#0) in layer parent (no-parent).
04-01 20:47:00.707 D/KernelUidCpuTimeReader(1332): Removing uids 99000-99000
04-01 20:47:00.763 D/ConnectivityService(1332): ConnectivityService NetworkRequestInfo binderDied(NetworkRequest [ LISTEN id=14, [ Capabilities: INTERNET&NOT_RESTRICTED&TRUSTED] ], android.os.BinderProxy@deb62bf)
04-01 20:47:00.764 W/ActivityManager(1332): setHasOverlayUi called on unknown pid: 2524
04-01 20:47:01.208 W/OpenGLRenderer(2351): Incorrectly called buildLayer on View: ShortcutAndWidgetContainer, destroying layer...
04-01 20:47:03.282 V/ANDR-PERF-MPCTL(502): Invalid profile no. 0, total profiles 0 only
```

voidstarstar:
A crash occurs if the GCM settings of an app are viewed, the app is uninstalled, and then the settings are viewed again.

This can be replicated as follows:

* Install an app that uses GCM.
* Launch it once so it registers.
* Check app registration status in microG and also tap on it to open the settings page.
* Uninstall the app.
* Go back to microG, which should still be on the settings page for the uninstalled app.
* MicroG instantly crashes.

The crash would occur since the app no longer exists in the database, but the `updateAppDetails` method assumes it does and does not have a null check.

This PR will finish the `Activity` of the `GcmAppFragment` without updating the details if the app was uninstalled while microG was in the background.

In addition, this PR will fix unregistering an app from GCM on Android 8.0+.  Apps will now unregister when fully uninstalled or when their data is cleared.  This fixes #743.

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/gcm/UnregisterReceiver.java
+++ b/play-services-core/src/main/java/org/microg/gms/gcm/UnregisterReceiver.java
@@ -8,7 +8,10 @@ import android.util.Log;
 import java.util.List;
 
 import static android.content.Intent.ACTION_PACKAGE_REMOVED;
+import static android.content.Intent.ACTION_PACKAGE_DATA_CLEARED;
+import static android.content.Intent.ACTION_PACKAGE_FULLY_REMOVED;
 import static android.content.Intent.EXTRA_DATA_REMOVED;
+import static android.content.Intent.EXTRA_REPLACING;
 
 public class UnregisterReceiver extends BroadcastReceiver {
     private static final String TAG = "GmsGcmUnregisterRcvr";
@@ -16,10 +19,13 @@ public class UnregisterReceiver extends BroadcastReceiver {
     @Override
     public void onReceive(final Context context, Intent intent) {
         Log.d(TAG, "Package changed: " + intent);
-        if (ACTION_PACKAGE_REMOVED.contains(intent.getAction()) && intent.getBooleanExtra(EXTRA_DATA_REMOVED, false)) {
+        if ((ACTION_PACKAGE_REMOVED.contains(intent.getAction()) && intent.getBooleanExtra(EXTRA_DATA_REMOVED, false) &&
+                !intent.getBooleanExtra(EXTRA_REPLACING, false)) ||
+                ACTION_PACKAGE_FULLY_REMOVED.contains(intent.getAction()) ||
+                ACTION_PACKAGE_DATA_CLEARED.contains(intent.getAction())) {
             final GcmDatabase database = new GcmDatabase(context);
             final String packageName = intent.getData().getSchemeSpecificPart();
-            Log.d(TAG, "Package removed: " + packageName);
+            Log.d(TAG, "Package removed or data cleared: " + packageName);
             final GcmDatabase.App app = database.getApp(packageName);
             if (app != null) {
                 new Thread(new Runnable() {
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -250,7 +250,8 @@
 
         <receiver android:name="org.microg.gms.gcm.UnregisterReceiver">
             <intent-filter>
-                <action android:name="android.intent.action.PACKAGE_ADDED"/>
+                <action android:name="android.intent.action.PACKAGE_DATA_CLEARED"/>
+                <action android:name="android.intent.action.PACKAGE_FULLY_REMOVED"/>
                 <action android:name="android.intent.action.PACKAGE_REMOVED"/>
 
                 <data android:scheme="package"/>
--- a/play-services-core/src/main/java/org/microg/gms/ui/GcmAppFragment.java
+++ b/play-services-core/src/main/java/org/microg/gms/ui/GcmAppFragment.java
@@ -99,6 +99,10 @@ public class GcmAppFragment extends ResourceSettingsFragment {
 
     private void updateAppDetails() {
         GcmDatabase.App app = database.getApp(packageName);
+        if (app == null) {
+            getActivity().finish();
+            return;
+        }
         PreferenceScreen root = getPreferenceScreen();
 
         SwitchPreference wakeForDelivery = (SwitchPreference) root.findPreference(PREF_WAKE_FOR_DELIVERY);


ale5000-git:
@voidstarstar
The `if (database != null)` check isn't better to be placed inside `updateAppDetails()`?

So it will fix also other possible crashes.

voidstarstar:
@ale5000-git That's a good point.  I'll force push a commit to fix that.  It won't fix the "ghost entry" described in #743, but it should at least prevent it from crashing.

Nanolx:
I guess the issue #743 "Ghost icon" is related to [this peace of code](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/java/org/microg/gms/ui/GcmFragment.java#L195).

commit  ef9e9b0921aac40d2484d2e4e7244e1b73c0450b message (which adds the setIcon part): GCM: fix display of uninstalled apps..., so I'd say the "Ghost icon" is intentional.

Nanolx:
One more thing, trying with some app I get this in the logcat:

`04-02 22:46:28.916  1358  1378 W BroadcastQueue: Background execution not allowed: receiving Intent { act=android.intent.action.PACKAGE_REMOVED dat=package:com.example.app flg=0x4000010 (has extras) } to com.google.android.gms/org.microg.gms.gcm.UnregisterReceiver`

happening to `com.google.android.gms/org.microg.nlp.PackageReceiver` aswell, this [Stackoverflow Q](https://stackoverflow.com/questions/46221057/awareness-api-android-o-using-broadcastreceiver/46390123#46390123) seems to be related (going to sleep soon, so can't check right now).

voidstarstar:
Good catch about the "ghost entry".  I didn't realize this was intentional.  For some reason when I uninstall an app, the entry just disappears, so I assumed the "ghost entry" was a bug I couldn't replicate.  This probably means that my PR breaks whatever feature the "ghost entry" was intended to provide.

Regarding the logcat, is that caused by this PR?  Are you on Android 8.0?

Nanolx:
I'm on Android 9.0 and the log is not related to your PR, but to the "Ghost icon" referenced (sorry, wasn't clear enough).

voidstarstar:
Thanks for clarifying.

Interesting.  I might be misunderstanding, but it sounds like broadcasting an implicit `Intent` is [effectively banned](https://developer.android.com/about/versions/oreo/background#broadcasts) in Android O (API 26) and above.  So maybe the `UnregisterReceiver` is unable to receive the `android.intent.action.PACKAGE_REMOVED` broadcast that it's [listening for](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/AndroidManifest.xml#L254) and that causes the "ghost entry" to remain listed?

Maybe we need to use [Context.registerReceiver()](https://developer.android.com/reference/android/content/Context.html#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter)) instead of relying on the manifest.

Nanolx:
Yes, that's what it looks like (and what I meant).

Am 2. April 2019 23:38:46 MESZ schrieb voidstarstar <notifications@github.com>:
>Thanks for clarifying.
>
>Interesting.  I might be misunderstanding, but it sounds like
>broadcasting an implicit `Intent` is [effectively
>banned](https://developer.android.com/about/versions/oreo/background#broadcasts)
>in Android O (API 26) and above.  So maybe the `UnregisterReceiver` is
>unable to receive the `android.intent.action.PACKAGE_REMOVED` broadcast
>that it's [listening
>for](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/AndroidManifest.xml#L254)
>and that causes the "ghost entry" to remain listed?
>
>Maybe we need to use
>[Context.registerReceiver()](https://developer.android.com/reference/android/content/Context.html#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter))
>instead of relying on the manifest.
>
>-- 
>You are receiving this because you commented.
>https://github.com/microg/android_packages_apps_GmsCore/pull/744#issuecomment-479214952

-- 

ale5000-git:
The app is automatically removed from the list when it is uninstalled (but it wasn't in old versions of microG).
So the code fix the display when the user has this stituation from previous microG version.
Probably it can happen also if an app is uninstalled when microG is crashed (I'm not sure).

Nanolx:
It also happens on Oreo+ due to the background intent allowed.

Am 3. April 2019 11:41:51 MESZ schrieb ale5000 <notifications@github.com>:
>The app is automatically removed from the list when it is uninstalled
>(but it wasn't in old versions of microG).
>So the code fix the display when the user has this stituation from
>previous microG version.
>Probably it can happen also if an app is uninstalled when microG is
>crashed (I'm not sure).
>
>-- 
>You are receiving this because you commented.
>https://github.com/microg/android_packages_apps_GmsCore/pull/744#issuecomment-479417139

-- 

voidstarstar:
OK, great.  That code should be rendered harmless after this patch at least.

Regarding the Oreo+ bug, with a brief search, I've seen people use see many possible solutions, each with upsides and downsides.  I've ordered some of them from worst to best (in my opinion):

1. Use a [Context.registerReceiver()](https://developer.android.com/reference/android/content/Context.html#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter)) to register the `UnregisterReceiver` to run in the main activity thread.
    * This seems relatively simple to implement.
    * I think this will only receive the Intent while the microG app is running.  If microG is killed for any reason by the OS, it will go out of sync.
    * This requires that the `ACTION_PACKAGE_REMOVED` broadcast is sticky, but I'm not sure if this is true.
2. Make the `UnregisterReceiver` start on boot and self register for the Intent similar to what is described [here](https://medium.com/@debuggingisfun/android-handle-implicit-intents-in-post-oreo-world-6bf51854cf65).
    * This also seems relatively simple to implement.
    * I'm not sure if this will actually work.
3. Create a new job for the [JobScheduler](https://developer.android.com/reference/android/app/job/JobScheduler).
    * This is a bit more complicated.
    * The jobs may be arbitrarily delayed, so real-time results are not possible.
4. Use the new (API 26) [`PackageManager.getChangedPackages()`](https://developer.android.com/reference/android/content/pm/PackageManager.html#getChangedPackages(int)).
    * This was created at the same API level that `ACTION_PACKAGE_REMOVED` was banned, so it seems like this might be an intended replacement.
    * This does polling, so the app won't unregister for GCM immediately.
5. Use [`ACTION_PACKAGE_FULLY_REMOVED`](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_FULLY_REMOVED) instead of `ACTION_PACKAGE_REMOVED`.  This flag is almost the same, except `EXTRA_DATA_REMOVED` is true (which is [already being done](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/java/org/microg/gms/gcm/UnregisterReceiver.java#L19)) and `EXTRA_REPLACING` is false (which is not done, but probably should have been).
    * This seems like a very easy solution.
    * This requires API level 14 but microG has a [minSdkVersion of 9](https://github.com/microg/android_packages_apps_GmsCore/blob/fa403aa99a101120209ae1f4e6d963aa9afe2e1b/build.gradle#L40).  I think this means we will need to still handle both Intents.

Unless anyone has another solution, I would personally favor option 5.  The `UnregisterReceiver` also filtered for [ACTION_PACKAGE_ADDED](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/AndroidManifest.xml#L253), but I'm not sure why.  It doesn't look like this was ever used, so would this be safe to remove?

Vavun:
Then it should also be triggered on  ACTION_PACKAGE_DATA_CLEARED

ale5000-git:
I think the best is handling ACTION_PACKAGE_FULLY_REMOVED and ACTION_PACKAGE_REMOVED and ACTION_PACKAGE_DATA_CLEARED at the same time since also the latter require the app to re-register itself.

voidstarstar:
I like the idea about adding `ACTION_PACKAGE_DATA_CLEARED`.  I assume the official Play Services unregisters the app when the app data is cleared too?

Vavun:
>  I assume the official Play Services unregisters the app when the app data is cleared too?

official Play Services provided on android from 6.0 to 9.0 uses this list of intent actoins to trigger `GmsIntentOperationService`
```
            <intent-filter>
                <action android:name="android.intent.action.PACKAGE_ADDED"/>
                <action android:name="android.intent.action.PACKAGE_CHANGED"/>
                <action android:name="android.intent.action.PACKAGE_DATA_CLEARED"/>
                <action android:name="android.intent.action.PACKAGE_FULLY_REMOVED"/>
                <action android:name="android.intent.action.PACKAGE_REMOVED"/>
                <action android:name="android.intent.action.PACKAGE_REPLACED"/>
                <data android:scheme="package"/>
            </intent-filter>
```

voidstarstar:
I made some changes based on the discussion above.  I tested this on a pre-Android 8.0 device and it now unregisters apps when the data is cleared.  Unfortunately I don't own a device that supports Android 8.0+, so I can't test if the "ghost entry" disappears.  I pushed the changes and also created a [temporary release for testing purposes](https://github.com/voidstarstar/android_packages_apps_GmsCore/releases/tag/fix_crash_after_app_unregisters_patch).

Nanolx:
This fixes the issue on Pie (so I assume on Oreo as well), the app now properly gets unregistered after uninstallation and disappears in the microG GCM list, log now shows (as expected):

```
04-03 23:05:20.298  8071  8071 D GmsGcmUnregisterRcvr: Package changed: Intent { act=android.intent.action.PACKAGE_FULLY_REMOVED dat=package:com.example.app flg=0x5000010 cmp=com.google.android.gms/org.microg.gms.gcm.UnregisterReceiver (has extras) }
04-03 23:05:20.298  8071  8071 D GmsGcmUnregisterRcvr: Package removed: com.example.app
```

Great. :)

voidstarstar:
There's another related UX issue that I think could be improved.  This probably deserves its own issue, however, but I'll mention it here since it's very relevant.

Currently I think the following occurs:
1. If a user fully uninstalls an app or clears the data of an app, then it is removed from the `GcmDatabase` and it no longer appears in the GCM list.  If the GCM settings of an app were open, then that activity is finished to prevent a crash.  The next time the app is opened (or reinstalled and opened), it can register for GCM again.
2. If a user opens the GCM settings of an app and then clicks `Unregister`, the app is not removed from the `GcmDatabase` and it still appears in the GCM list as a "ghost entry".  The next time the app is opened, it can **not** register for GCM again, even if `Allow registration` is selected in the GCM settings of the app.

Does the `Allow registration` option have any effect at all or is it currently completely broken?

Scenario 2 seems like a legitimate use for the "ghost entry" since a user may want to both unregister the app and prevent it from registering in the future, but not want to completely uninstall it.  The problem is that if the user ever wants to register it in the future, it does not appear possible to me without uninstalling or clearing the data first to remove it from the `GcmDatabase`.

**Suggested solution:**
Perhaps a better UX solution would be to never remove an app from the `GcmDatabase` unless it is fully uninstalled.  In the cases where the app has the data cleared or is manually unregistered, it will unregister and leave a "ghost entry".  If a "ghost entry" is present, then the next time the app requests to register, the `Allow registration` option will be respected.

Nanolx:
I never thought about Scenario 2, nor have I hit it in real life usage, I'll test it the next days.

Ah, and now that I come to think of it: there's a similar non-working receiver in UnifiedNlp (`org.microg.nlp.PackageReceiver`) on Oreo.

voidstarstar:
I might be wrong about how Scenario 2 is handled in the database, but I think it's covered by the code [here](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/java/org/microg/gms/ui/GcmFragment.java#L204-L205) and [here](https://github.com/microg/android_packages_apps_GmsCore/blob/40974f806d79979b4cd130d70d0b30a55f464201/play-services-core/src/main/java/org/microg/gms/ui/GcmAppFragment.java#L138-L141).  After the user clicks `Unregister`, the entry remains in the list but with a permanent `Not registered` status.

voidstarstar:
I force-pushed to fix a mistake.  There was a missing `return` statement in my code that was introduced when I moved the fix to `updateAppDetails()`.  The [temporary release](https://github.com/voidstarstar/android_packages_apps_GmsCore/releases/tag/fix_crash_after_app_unregisters_patch) has been updated as well.  I tested and the crash is now prevented again.  Sorry about that.

@Nanolx You may want to add this [missing `return`](https://github.com/microg/android_packages_apps_GmsCore/compare/3da0e5d91e481f8ac7a550da5d0f191fce133fc6..cc117488d115d67eef9c6b3177a29af70aaace7b) statement to your fork.

ale5000-git:
@voidstarstar
I think it may be better to change also the log message from `Package removed` to `Package removed or data cleared`.

Vavun:
> and now that I come to think of it: there's a similar non-working receiver in UnifiedNlp

Maybe that is not receiver problem ?
https://developer.android.com/about/versions/pie/android-9.0-migration#tya
This receiver trying to start background service that is not allowed to run

So, I think that is global microG problem since target sdk of your fork is 28 now (latest vanilla still 27)

Vavun:
Anyway. Is it possible to implement this ?
And this for Pie
https://developer.android.com/reference/android/Manifest.permission.html#FOREGROUND_SERVICE

voidstarstar:
@Vavun Are you suggesting to make microG run during Direct Boot mode?  If so, how does microG currently behave on Android 7.0+ devices before the user unlocks the device?  You might want to make a separate issue for it as it's not really related to anything here.

Also, the [PackageReceiver](https://github.com/microg/android_packages_apps_UnifiedNlp/blob/82479b79c76353f532e0c6edd0d1dee8d49c48f4/unifiednlp-base/src/main/java/org/microg/nlp/PackageReceiver.java) filters [5 different Intents](https://github.com/microg/android_packages_apps_UnifiedNlp/blob/82479b79c76353f532e0c6edd0d1dee8d49c48f4/unifiednlp-base/src/main/AndroidManifest.xml#L112-L116).  It doesn't look like these are part of the implicit broadcast exceptions, so unfortunately a different solution might be needed.

ale5000-git:
@voidstarstar 
Since `PackageReceiver` realod backends, I think it make sense to add `android.intent.action.PACKAGE_DATA_CLEARED` and `android.intent.action.PACKAGE_FULLY_REMOVED` als here.

voidstarstar:
@ale5000-git To be honest, I haven't looked at UnifiedNlp much, so I'm not sure when `PackageReceiver` is required to run.  On what conditions should it be reloading the backends?

Is it fine if this only runs for `android.intent.action.PACKAGE_DATA_CLEARED` and `android.intent.action.PACKAGE_FULLY_REMOVED` but does not run for any of the Intents currently in the manifest?

For reference:

[`ACTION_PACKAGE_CHANGED`](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_CHANGED)
[`ACTION_PACKAGE_REMOVED`](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_REMOVED)
[`ACTION_PACKAGE_REPLACED`](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_REPLACED)
[`ACTION_PACKAGE_RESTARTED`](https://developer.android.com/reference/android/content/Intent.html#ACTION_PACKAGE_RESTARTED)

Also, I see no mention of `ACTION_PACKAGE_UPGRADED` or `android.intent.action.PACKAGE_UPGRADED` in the Android API or anywhere else online for that matter.  Am I missing something or does this Intent not even exist?  Is it maybe a typo?

play-services-core/src/main/java/org/microg/gms/gcm/UnregisterReceiver.java
```diff
@@ -8,18 +8,24 @@
...
 
 public class UnregisterReceiver extends BroadcastReceiver {
     private static final String TAG = "GmsGcmUnregisterRcvr";
 
     @Override
     public void onReceive(final Context context, Intent intent) {
         Log.d(TAG, "Package changed: " + intent);
-        if (ACTION_PACKAGE_REMOVED.contains(intent.getAction()) && intent.getBooleanExtra(EXTRA_DATA_REMOVED, false)) {
+        if ((ACTION_PACKAGE_REMOVED.contains(intent.getAction()) && intent.getBooleanExtra(EXTRA_DATA_REMOVED, false) &&
+                intent.getBooleanExtra(EXTRA_REPLACING, true)) ||
```
mar-v-in: Use `!intent.getBooleanExtra(EXTRA_REPLACING, false)` instead of `intent.getBooleanExtra(EXTRA_REPLACING, true)`

voidstarstar: Done

PR #761, opened 2019-04-18
Include GCM message ID in app message intent

mh0rst:
This fixes a NullPointerException when receiving a notification in react-native apps, like Mattermost (https://github.com/mattermost/mattermost-mobile)

The used framework expects the message ID to be present in the app message intent, leading to a NullPointerException when it isn't:
Here is the check: https://github.com/wix/react-native-notifications/blob/master/android/src/main/java/com/wix/reactnativenotifications/core/notification/PushNotification.java#L66
This returns null: https://github.com/wix/react-native-notifications/blob/master/android/src/main/java/com/wix/reactnativenotifications/core/notification/PushNotification.java#L47
And finally, the NPE: https://github.com/wix/react-native-notifications/blob/master/android/src/main/java/com/wix/reactnativenotifications/gcm/FcmInstanceIdListenerService.java#L29


mar-v-in:
Thanks!

PR #766, opened 2019-04-20
build.gradle improve gitVersionBase calculation

Nanolx:
Currently the most recent tag is taken into account when calculating gitVersionBase. While this works as advertised for upstream, which only has v[0-9]* release-tags, this may cause issues on forks (which may have other tags aswell).

This change in build.gradle should make the build system happy in both upstream and fork repos, by switching from `git describe --tags --abbrev=0` to `git tag --list 'v[0-9]*' --sort=-v:refname`.

### DIFF
--- a/play-services-core/build.gradle
+++ b/play-services-core/build.gradle
@@ -50,7 +50,7 @@ def execResult(...args) {
 
 def gmsVersion = "13.2.80"
 def gmsVersionCode = Integer.parseInt(gmsVersion.replaceAll('\\.', ''))
-def gitVersionBase = execResult('git', 'describe', '--tags', '--abbrev=0').substring(1)
+def gitVersionBase = execResult('git', 'describe', '--tags', '--abbrev=0', '--match=v[0-9]*').substring(1)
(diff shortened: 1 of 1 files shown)


voidstarstar:
I ran into the same problem and was going to push a solution, but yours is better.

I don't think this will work properly if microG starts having pre-releases with suffixes (e.g. v2.6.13280-beta2), but this can be fixed by using the `versionsort.prereleaseSuffix` or `versionsort.suffix` config variables.  So far microG hasn't had any pre-releases, so this will probably not be a problem.

mar-v-in:
This doesn't work as expected if anyone wants to build something else than the very latest version. If you checkout an older version, `git describe --tags` will return that version while `git tag --list` always returns all known tags, so the change would always return the newest version.

voidstarstar:
That's a good point.  What if we just use `git describe --tags --abbrev=0 --match=v[0-9]*`?

Nanolx:
Yes, I haven't thought about that, because of the nature of microG a build of an earlier version is rather unlikely in a real-world setup (i. e. anything but hunting down a regression). I've changed it to `git tag describe --abbrev=0 --match=v[0-9]*`.

mar-v-in:
Thanks!

PR #804, opened 2019-06-02
Add org.microg.tools.CondLog() for conditional logging (fix #251)

### LINKED ISSUE #251: sensitive information logged?
i have a google account configured. i am looking at my logcat, and i see stuff in the log that seems to be sensitive. in particular, i've seen suspect entries with these tags:
- GmsAuthManager
- GmsAuthManagerSvc
- GmsHttpFormClient

am i right? is there a way to disable this logging entries?

Nanolx:
Add new CondLog() [Conditional Logging] function to `play-services-core`. It only prints the Log if the desired Log.Level isLoggable(). Took some inspiration from [this gist](https://gist.github.com/kristopherjohnson/11219895).

Also includes a test commit within `AuthManagerServiceImpl`.

@mar-v-in let me know if you think this is suitable for #251 

cc-ing @ale5000-git @Lanchon

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
+++ b/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
@@ -37,6 +37,7 @@ import com.google.android.gms.auth.TokenData;
 import com.google.android.gms.common.api.Scope;
 
 import org.microg.gms.common.PackageUtils;
+import org.microg.tools.CondLog;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -79,7 +80,7 @@ public class AuthManagerServiceImpl extends IAuthManagerService.Stub {
         packageName = PackageUtils.getAndCheckCallingPackage(context, packageName, extras.getInt(KEY_CALLER_UID, 0), extras.getInt(KEY_CALLER_PID, 0));
         boolean notify = extras.getBoolean(KEY_HANDLE_NOTIFICATION, false);
 
-        Log.d(TAG, "getToken: account:" + accountName + " scope:" + scope + " extras:" + extras + ", notify: " + notify);
+        CondLog.print(TAG, Log.DEBUG, "getToken: account:" + accountName + " scope:" + scope + " extras:" + extras + ", notify: " + notify);
 
         /*
          * TODO: This scope seems to be invalid (according to https://developers.google.com/oauthplayground/),
--- a/play-services-core/src/main/java/org/microg/tools/CondLog.java
+++ b/play-services-core/src/main/java/org/microg/tools/CondLog.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (C) 2019 microG Project Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.microg.tools;
+
+import android.util.Log;
+
+public class CondLog
+{
+
+	private CondLog()
+	{
+	}
+
+	public static boolean isLoggable(String tag, final int level)
+	{
+		// Note: Log.isLoggable() will only return true for levels of INFO and above by
+		// default. To override this behavior, set system properties as described in the
+		// documentation for Log.isLoggable(), or add appropriate filtering code here.
+
+		// Log.isLoggable() will throw an exception if the length of the tag is greater than
+		// 23 characters, so trim it if necessary to avoid the exception.
+		if (tag.length() > 23)
+		{
+			tag = tag.substring(0, 22);
+		}
+
+		return Log.isLoggable(tag, level);
+	}
+
+	/**
+	 * Call Log.*(tag, msg, tr) if loggable
+	 * 
+	 * @param tag
+	 *            Used to identify the source of a log message. It usually identifies the class
+	 *            or activity where the log call occurs.
+	 * @param level
+	 *            Requested log level
+	 * @param msg
+	 *            The message you would like logged.
+	 * @param tr
+	 */
+	public static void print(final String tag, final int level, final String msg, final Throwable tr)
+	{
+		if (isLoggable(tag, level))
+		{
+			switch (level) {
+				case Log.ASSERT:
+					Log.wtf(tag, msg, tr);
+					break;
+				case Log.VERBOSE:
+					Log.v(tag, msg, tr);
+					break;
+				case Log.DEBUG:
+					Log.d(tag, msg, tr);
+					break;
+				case Log.INFO:
+					Log.i(tag, msg, tr);
+					break;
+				case Log.WARN:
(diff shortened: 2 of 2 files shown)


ale5000-git:
Very good idea, but I propose 2 DEBUG levels:
- DEBUG (standard with all messages)
- DEBUG_SAFE same as DEBUG but without auth info (that may compromise account if published online)

Nanolx:
microG has different log levels already, so we could either do more fine granulatio/re-classification or check whether we're on a debug build, instead of a relase build with something like:

```
diff --git a/play-services-core/src/main/java/org/microg/tools/CondLog.java b/play-services-core/src/main/java/org/microg/tools/CondLog.java
index 36c0cb5..be20cbb 100644
--- a/play-services-core/src/main/java/org/microg/tools/CondLog.java
+++ b/play-services-core/src/main/java/org/microg/tools/CondLog.java
@@ -42,7 +42,11 @@ public class CondLog
 			tag = tag.substring(0, 22);
 		}
 
-		return Log.isLoggable(tag, level);
+		if (! BuildConfig.DEBUG && level = Log.DEBUG) {
+			return false
+		} else {
+			return Log.isLoggable(tag, level);
+		}
 	}
 
 	/**
```

ale5000-git:
There still can be the case when someone on a debug build need to post the log publicly.
So I think we need a more fine granulation/reclassification of log messages (at least of the ones containing sensitive informations or session ids that may allow session hijacking).

Nanolx:
See the last commit if it more fits your needs, I've simply added DEBUG_SAFE as a new, custom log level, which equals to 42 (because it's the meaning of life).

Additionally isLoggable() will always return false for DEBUG_SAFE if we're on a release build.

Lanchon:
logs should never log sensitive information. this applies to every system. i shouldnt be exposed to risks if i choose DEBUG logs; this is a bug. i shouldnt need to know that a DEBUG_SAFE level exists not to be in danger!

DEBUG should be safe, and if needed, a DEBUG_WITH_SENSITIVE_INFO or something similar could be added. (DEBUG_UNSAFE is not a good name; it doesnt protect users from phishing attacks from others posing as giving help.)

more correctly, an independent LOG_SENSITIVE_INFO setting could be added which triggers logging of unsafe stuff at the proper level which probably shouldnt be DEBUG. a WARNING entry should be logged whenever this LOG_SENSITIVE_INFO mode is enabled. even if a DEBUG_WITH_SENSITIVE_INFO level is implemented, it should log a warning when selected. security is a serious thing.

additionally, if DEBUG currently produces sensitive output, then id STRONGLY suggest removing that level completely. maybe replace it with VERBOSE or SAFE_DEBUG or something similar.

the rationale would be avoiding this scenario where someone honest asks for a DEBUG log and is sent a tainted DEBUG from an old version of microg.

Nanolx:
You don't need to know about anything about DEBUG_SAFE (which maybe should be DEBUG_SENSITIVE instead?), this pull request would only allow logging of sensitive information in debug builds, ordinary users grab their apk from F-Droid or Github, so they are always on release builds.

Otherwise I agree with you. Yes, microG currently logs stuff, you don't want it to.

This pull request should lay the basis to clean stuff up in user builds, but still leave the door open in debug builds.

Am 3. Juni 2019 15:53:26 MESZ schrieb Lanchon <notifications@github.com>:
>logs should never log sensitive information. this applies to every
>system. i shouldnt be exposed to risks if i choose DEBUG logs; this is
>a bug. i shouldnt need to know that a DEBUG_SAFE level exists not to be
>in danger!
>
>DEBUG should be safe, and if needed, a DEBUG_WITH_SENSITIVE_INFO or
>something similar could be added. (DEBUG_UNSAFE is not a good name; it
>doesnt protect users from phishing attacks from others posing as giving
>help.)
>
>more correctly, an independent LOG_SENSITIVE_INFO setting could be
>added which triggers logging of unsafe stuff at the proper level which
>probably shouldnt be DEBUG. a WARNING entry should be logged whenever
>this LOG_SENSITIVE_INFO mode is enabled. even if a
>DEBUG_WITH_SENSITIVE_INFO level is implemented, it should log a warning
>when selected. security is a serious thing.
>
>additionally, if DEBUG currently produces sensitive output, then id
>STRONGLY suggest removing that level completely. maybe replace it with
>VERBOSE or SAFE_DEBUG or something similar.
>
>the rationale would be avoiding this scenario where someone honest asks
>for a DEBUG log and is sent a tainted DEBUG from an old version of
>microg.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/microg/android_packages_apps_GmsCore/pull/804#issuecomment-498266034

-- 

voidstarstar:
If there's going to be 2 logging functions for each logging level, I agree with @Lanchon that the one that includes sensitive data should contain additional verbosity.  In other words, we should use `DEBUG_SENSITIVE_DATA` and `DEBUG` instead of `DEBUG_SAFE` and `DEBUG`.  We should assume that anything unsafe will be explicitly (and verbosely) marked.

I would also suggest phrasing this with a preference of "sensitive data" over "sensitive info" to avoid confusion with the `INFO` logging level.

A global flag, off by default, should be as verbose and obvious as possible.  Probably a good idea to put the words "DANGEROUS" and/or "I_KNOW_WHAT_IM_DOING" in there for good measure.

> additionally, if DEBUG currently produces sensitive output, then id STRONGLY suggest removing that level completely. maybe replace it with VERBOSE or SAFE_DEBUG or something similar.
>
> the rationale would be avoiding this scenario where someone honest asks for a DEBUG log and is sent a tainted DEBUG from an old version of microg.

Even if we change future versions of microG to not use `DEBUG`, how will that prevent the old versions from using it?  This seems like a pointless change unless we somehow have all users always change their default logging levels before they paste a logcat, which seems unreasonable to me.  We could have only affected users change their logging level, but at that point, we might as well just tell users to upgrade to the latest microG before posting logcats (we shouldn't care about fixing bugs in old versions of microG anyway).

Nanolx:
I've renamed DEBUG_SAFE to DEBUG_SENSITIVE to be more clear what's going on.

Before going further we should wait for what @mar-v-in has to say regarding the logging.

ale5000-git:
@voidstarstar: We must always tell users to upgrade to the latest microG before posting logcat otherwise we will have a lot of issuesa for already fixed bugs.

@Nanolx: I t seems nice but how can be switched on/off DEBUG_SENSITIVE in debug builds?

Nanolx:
@ale5000-git in the current state you can't, because it's always enabled in debug builds and always disabled in release builds. It would be possible to either

a) add an optional compile time variable (override-able from `local.properties`)
b) add a system property (like microg.debug.sensitive using something like [this reflection](https://wadehuang36.github.io/2018/06/14/android-use-system-properties-as-debug-options.html))

ale5000-git:
For sensitive info I think it is safer a compile time variable, also it would be nice to avoid calling Log.isLoggable for DEBUG_SENSITIVE on DEBUG builds.

ale5000-git:
Like this:
```
		if (level == DEBUG_SENSITIVE) {
			return BuildConfig.DEBUG && DEBUG_SENSITIVE_IS_ENABLED;
		} else {
			return Log.isLoggable(tag, level);
		}
```

Nanolx:
Yes, that's probably the best if we go with a).

Nanolx:
@mar-v-in any interest in this? If not, I'm going to close it.

mar-v-in:
I personally prefer a system that by default removes sensitive information from log or replaces it with stars *but keeps the log entries*. And allows users to enable logging the sensitive information when needed. However, I'm not sure how to easily realize such system without significant additional burden during development.

ILoveOpenSourceApplications:
@mar-v-in, some of these PRs are years old. Shouldn't we merge the ones requiring only minor changes and close those that are no longer significant or would require major rework?

PR #805, opened 2019-06-02
Create values-cy/strings.xml

Cymrodor:


mar-v-in:
Create translations via Weblate: https://hosted.weblate.org/projects/microg/

PR #821, opened 2019-06-06
Improve building into AOSP Pie ROM

chris42:
Following commits improve building within a ROM as described on https://github.com/microg/android_packages_apps_GmsCore/wiki/Building

Depending on the ROM, there might be more changes needed (e.g. toolchain versions). However this are structural changes.
I tested this on Sony AOSP Pie.

Reasons for changes:
- 4b1aec8: GmsCore.apk should be in system/priv-app
- 2a2eacc: Removed ANDROID_BUILD_TOP as deprecated; Previous linking would not work with build on multiple threads
- 51a5ec3: Sony AOSP Pie ROM won't boot without the privapp-whitelisting, others seem to have same issue
- 3911c35: Newer gradle versions limit Heap to 512MB, however 1GB is needed
- ab34ef2: proguard warnings would break compile

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -34,6 +34,7 @@ $(gmscore_root)/$(gmscore_dir)/$(gmscore_apk):
 	cd $(gmscore_root)/$(gmscore_dir) && JAVA_TOOL_OPTIONS="$(JAVA_TOOL_OPTIONS) -Dfile.encoding=UTF8" ../gradlew assembleRelease
 
 LOCAL_CERTIFICATE := platform
+LOCAL_PRIVILEGED_MODULE := true
 LOCAL_SRC_FILES := $(gmscore_dir)/$(gmscore_apk)
 LOCAL_MODULE_CLASS := APPS
 LOCAL_MODULE_SUFFIX := $(COMMON_ANDROID_PACKAGE_SUFFIX)
--- a/Android.mk
+++ b/Android.mk
@@ -21,17 +21,20 @@ LOCAL_PACKAGE_NAME := GmsCore
 
 gmscore_root  := $(LOCAL_PATH)
 gmscore_dir   := play-services-core
-gmscore_out   := $(TARGET_COMMON_OUT_ROOT)/obj/APPS/$(LOCAL_MODULE)_intermediates
+gmscore_out   := out/target/common/obj/APPS/$(LOCAL_MODULE)_intermediates
 gmscore_build := $(gmscore_root)/$(gmscore_dir)/build
 gmscore_apk   := build/outputs/apk/release/play-services-core-release-unsigned.apk
 
-$(gmscore_root)/$(gmscore_dir)/$(gmscore_apk):
-	rm -Rf $(gmscore_build)
-	mkdir -p $(ANDROID_BUILD_TOP)/$(gmscore_out)
-	ln -s $(ANDROID_BUILD_TOP)/$(gmscore_out) $(ANDROID_BUILD_TOP)/$(gmscore_build)
-	echo "sdk.dir=$(ANDROID_HOME)" > $(gmscore_root)/local.properties
-	cd $(gmscore_root) && git submodule update --recursive --init
-	cd $(gmscore_root)/$(gmscore_dir) && JAVA_TOOL_OPTIONS="$(JAVA_TOOL_OPTIONS) -Dfile.encoding=UTF8" ../gradlew assembleRelease
+.PHONY: preps_gms
+preps_gms:
+	rm -rf $(gmscore_build); \
+	mkdir -p $(gmscore_out); \
+	ln -s ../../../../$(gmscore_out) $(gmscore_build)
+
+$(gmscore_root)/$(gmscore_dir)/$(gmscore_apk): preps_gms
+	echo "sdk.dir=$(ANDROID_HOME)" > $(gmscore_root)/local.properties; \
+	cd $(gmscore_root) && git submodule update --recursive --init; \
+	cd $(gmscore_dir) && JAVA_TOOL_OPTIONS="$(JAVA_TOOL_OPTIONS) -Dfile.encoding=UTF8" ../gradlew assembleRelease
 
 LOCAL_CERTIFICATE := platform
 LOCAL_PRIVILEGED_MODULE := true
--- a/Android.mk
+++ b/Android.mk
@@ -13,6 +13,15 @@
 # limitations under the License.
 
 LOCAL_PATH:= $(call my-dir)
+
+include $(CLEAR_VARS)
+LOCAL_MODULE := privapp-permissions-com.google.android.gms.xml
+LOCAL_MODULE_TAGS := optional
+LOCAL_MODULE_CLASS := ETC
+LOCAL_MODULE_PATH := $(TARGET_OUT_ETC)/permissions
+LOCAL_SRC_FILES := permissions/$(LOCAL_MODULE)
+include $(BUILD_PREBUILT)
+
 include $(CLEAR_VARS)
 
 LOCAL_MODULE := GmsCore
@@ -41,5 +50,6 @@ LOCAL_PRIVILEGED_MODULE := true
 LOCAL_SRC_FILES := $(gmscore_dir)/$(gmscore_apk)
 LOCAL_MODULE_CLASS := APPS
 LOCAL_MODULE_SUFFIX := $(COMMON_ANDROID_PACKAGE_SUFFIX)
+LOCAL_REQUIRED_MODULES := privapp-permissions-com.google.android.gms.xml
 
 include $(BUILD_PREBUILT)
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,2 @@
 android.useDeprecatedNdk=true
+org.gradle.jvmargs=-Xmx1g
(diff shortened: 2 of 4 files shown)


pgera:
>GmsCore.apk should be in system/priv-app
>Sony AOSP Pie ROM won't boot without the privapp-whitelisting, others seem to have same issue

Are these necessary if you are signing with the platform key ?

chris42:
> Are these necessary if you are signing with the platform key ?

Android.mk had always `LOCAL_CERTIFICATE := platform`. Errors still occur. If you have any other information, please elaborate.

mar-v-in:
I actually think that it's not worth the effort to continue maintaining the building as part of a ROM through Android.mk.
It's not really hard to build the APK with gradle and put into into the Android build system as a prebuilt. With https://github.com/lineageos4microg/android_prebuilts_prebuiltapks there is also a repo containing the latest version of all the relevant APKs as prebuilt, so you don't have to build them yourself.

chris42:
I agree, that using the prebuild is easier, however then the wiki entry should be removed to include it into a build, as that will not work correctly with the current code.
Would then close the pull request?

pgera:
The current method of building outside and including a prebuilt works, but building into the ROM would streamline the process and provide better QA. For instance, the lineageos4microg prebuilts only track released binaries, which may break (as it happened recently). So you may need to build from git head, which needs this additional step of building and signing outside of AOSP. If you use something like https://github.com/dan-v/rattlesnakeos-stack, you can get nicer tests/QA if it can just be included in the mainfest directly.

mar-v-in:
I'd prefer to do QA before release to end-users.

Developers and early testers can always update early without baking into the ROM (you can just install the update from the `.apk` without updating the ROM if your ROM has support built-in). I can provide pre-signed nightly `.apk` for devs to test if needed, but only outside of F-Droid repo.

Everything that is in a ROM should ideally *not* be the git master head, as it more likely to be unstable.

pgera:
>Everything that is in a ROM should ideally not be the git master head, as it more likely to be unstable.

Yes, this is true, and I'm not advocating changing that. If you are doing your own builds in an automated fashion though, it is easier to change the tracking from release to a different hotfix branch or to master and go back to release later. Basically, these are all one line changes in the manifest. Ideally, you should only track release, but if something breaks, you'll know right away. In this model, there is no reliance on upstream's signing keys and no provision to track upstream signed binaries directly. Only the rom builder has control over what is installed on the device, which seems better for security.

pgera:
A side benefit is also staying closer to AOSP in general since lineageos lags behind significantly. On a new AOSP major version release, pixels are the only phones that are able to use it. So relying on lineage4microg's prebuilts and patches are anyway not useful directly without other changes.

chris42:
@pgera When I started to build it with my ROM, I had similar thoughts, however it is quite time consuming to make this work and keep up-to-date. It basically breaks with every major android version.

The setup using prebuilds can be completely detached from LineageOS, but has the advantage, that you can use the original apks with e.g. fdroid signatures during the ROM build, hence update them via fdroid afterwards. You can check my repository [https://github.com/chris42/android_prebuilts_prebuiltapks](https://github.com/chris42/android_prebuilts_prebuiltapks)
I include it in my current AOSP builds every month, without lineageOS.

@mar-v-in It might be an option to create such a repository within the microg project to offer easy integration into AOSP or other ROM builds, have QA in your hands and be on point with every release.
@Nanolx and @ale5000-git could also give a lot of input here, as they do a lot of the things (permissions, etc.) in their zips.

chris42:
@pgera Tested my new repository and it works nicely when included in a ROM. It is modular, so you can decide in your device.mk what you want to include. Microg shows all green checkmarks, only the location providers are unchecked.

@mar-v-in Is there a way to enable the location backends automatically in microg?

pgera:
@chris42 , did you test your repo with aosp 10 ? I have been using the lineageos4microg prebuilt repo for aosp 9, and it works fine. With aosp 10, it still works, but location permissions have different behaviour. In 10, even if you grant location permission to the backends to access location any time, they lose the permission unless the microg location backends page is open. On logcat, I see messages like:
```
12-08 13:31:22.762  1281  8788 W ActivityManager: Appop Denial: Accessing service org.fitchfamily.android.dejavu/.BackendService from pid=16423, uid=10063 requires appop COARSE_LOCATION
12-08 13:31:22.764  1281  8788 W ActivityManager: Appop Denial: Accessing service org.microg.nlp.backend.apple/.BackendService from pid=16423, uid=10063 requires appop COARSE_LOCATION
12-08 13:31:22.767  1281  8788 W ActivityManager: Appop Denial: Accessing service org.microg.nlp.backend.ichnaea/.BackendService from pid=16423, uid=10063 requires appop COARSE_LOCATION
12-08 13:31:22.769  1281  8788 W ActivityManager: Appop Denial: Accessing service org.openbmap.unifiedNlp/.services.RadiocellsLocationService from pid=16423, uid=10063 requires appop COARSE_LOCATION
```

I saw that your repo has a much wider set of permissions compared to the lineage one. I'm trying to find the minimal additional set of permissions needed to make it work with aosp 10.

chris42:
Sadly no, Sony is not yet done with AOSP 10, hence I can't get my test device to Android 10 to test and update my prebuilt repo.

chris42:
Think this is outdated. If needed I can help with prebuilts.

bobobo1618:
A couple of issues I see haven't been raised:

- Without verifiable (and verifed) builds, I have no way of ensuring that the APKs were built from committed/reviewed code. For a regular, unprivileged app, the assurances provided by F-Droid are enough but for a privileged app _with signature-spoofing permissions_, I think more scrutiny is warranted.

- Generating "prebuilts" separately from the ROM is inherently more expensive than compiling them as part of the ROM, which I suspect will lead to more people choosing prebuilts over compiling themselves.

Due to these issues, I'd really like to compile MicroG as part of the AOSP build process. It looks like this is the first update to `Android.mk` in two years so it doesn't seem to be especially difficult to maintain.

chris42:
@bobobo1618 Thanks for the input. Maybe my short opinion on the two points:
- Verifiable builds of course would be nice, but building yourself without it is not inherently more secure, it most times just feels more secure. To have it more secure you would need to analyze each change in the code before compiling, otherwise you could just import security issues via code. Very little people do that and sometimes it is not even possible (lacking skill), so for most that is not a way to go.
- I don't understand this. Using prebuilts is way easier and speedier in the ROM build. That is the reason why I switched to prebuilts, as maintenance probably dropped tenfold. Using the official builds allows me to update the apps (e.g. via fdroid) without a new ROM build.

Nevertheless if you want to build it during ROM build feel free to use the changes I proposed.
I personally would vote to have a prebuilt repository officially by @mar-v-in , but right now it seems his priorities are not on microg. :-(

chris42:
@pgera 
I actually now have some Android 10 running. Regarding location permissions: Right now I am not installing the backends during build, as it seems, that then the permissions are wrong (haven't looked into that in more detail). If I install them via fdroid, they seem to work fine.

pgera:
The issue for locations is at https://github.com/microg/android_packages_apps_GmsCore/issues/986 . It has been confirmed by multiple people. So it will likely need some changes for a proper fix. I also see some pending PRs for locations. I haven't been following the threads too closely though.

bobobo1618:
Apologies in advance for the rant.

> Verifiable builds of course would be nice, but building yourself without it is not inherently more secure, it most times just feels more secure.

A prebuilt APK could contain anything that the person with the signing key feels like publishing. It can't be (and more importantly, won't be) easily inspected. It's about the least secure thing you could possibly do to your device, it's even worse than `curl https://www.key-holder.com/stuff.sh | sudo bash`, where at least you could inspect what code is being run.

Building from public source in inherently more secure because you know that even if you don't choose to audit the code, you know that you _could_ audit the code, which acts as a deterrent against malicious behaviour. In addition, if the project has more than one maintainer, you can trust that they will keep track of changes and raise a fuss if something malicious happens. This means that instead of just trusting a single key-holder, you can trust a group of maintainers.

> I don't understand this. Using prebuilts is way easier and speedier in the ROM build. That is the reason why I switched to prebuilts, as maintenance probably dropped tenfold. 

If you're happy to use somebody else's prebuilts, yes, it's easy. I was describing the situation where you're not happy to use somebody else's prebuilts so you build your own prebuilts and integrate them with the AOSP build system. This setup is a lot more cumbersome than just integrating the source.

> Using the official builds allows me to update the apps (e.g. via fdroid) without a new ROM build.

This is a totally valid point but it comes back to the first problem: it requires that you trust the person or people who hold the signing key.

To add some background to what might seem borderline paranoia, one of the things that most appeals to me about MicroG is that it removes the backdoor Google has into most Android devices in the form of Play Services, which allows it to install and run arbitrary, unverifiable and uninspectable code. MicroG, in theory, gives access to Google's APIs without this backdoor.

However in practice, MicroG is routinely distributed and installed as an unverifiable APK. Should the holder(s) of the signing keys wish or be compelled, they have the power to sign and push out a malicious update which compromises my device.

This means that essentially, Google's backdoor is just being replaced with MicroG's backdoor.

chris42:
I understand your point and made mine, hence I will not expand on security. I just want to prevent people wasting time on ROM build setups, because they feel that is the secure way to do.

You can copy my repository and exchange the apks with your own ones if needed. 
This - or setting up a new repository on your own / compiling apks yourself - is still easier than compiling during ROM build, having same level of security.
During ROM build you get into lots of trouble with changed Android guidelines, different Android, SDK, gradle and whatnot versions. 

Trust me I did that for a year or so and I am trying to save you some time here.

PR #843, opened 2019-06-19
Correct Pie patch to prevent .orig files

chris42:
Hi, 
was annoyed by getting .orig files created, as index was off. This should fix it.

Best
  Chris


chris42:
This is idle for >6 month, I guess it is not needed

PR #933, opened 2019-08-26
Add Chinese-Simplified translation

BYZYB:


mar-v-in:
Please update translations via Weblate: https://hosted.weblate.org/projects/microg/

PR #958, opened 2019-10-18
possible SQL wear-api bug

crised:
I think the SQL column name should be `androidId` instead of `nodeId`.


ILoveOpenSourceApplications:
This PR is outdated and should be closed.

ILoveOpenSourceApplications:
@ale5000-git, this PR is outdated and should be closed, right?

ale5000-git:
The PR is certainly outdated since the file was moved here: https://github.com/microg/GmsCore/blob/master/play-services-wearable/core/src/main/java/org/microg/gms/wearable/ConfigurationDatabaseHelper.java
But I don't know if the change is correct or not.

@mar-v-in
Could you please check it?

PR #1004, opened 2020-01-11
Update AndroidManifest.xml to implement background location permission

eplinux:
In Android 10, a new permission is needed to access location in background.

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -71,6 +71,7 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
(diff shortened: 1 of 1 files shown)


p4t44:
@mar-v-in can you please merge this asap? UnifiedNLP is currently not functional on Android 10 in the background.

ploink:
@mar-v-in I agree with @p4t44. 
Some more details why we need it:
- I forked [Local-GSM-Backend](https://github.com/ploink/Local-GSM-Backend) and the only way I could get network location to work reliably with Osmand is to remove the COARSE_LOCATION permission requirement from the &lt;service&gt; tag in [the manifest](https://github.com/ploink/Local-GSM-Backend/blob/master/app/src/main/AndroidManifest.xml) This means that microG, or any other app does not need any permission whatsoever to talk to the backend and get a location. As soon as this patch is merged and  released I can close this security hole.

- I have several apps that use Fused Location. Only when I take microG settings to the foreground they can briefly get a location, but short time after it has gone to the background, fused location stops working. I think this background location permission would fix that.

mar-v-in:
This has been part of the big refactoring.

PR #1028, opened 2020-03-13
Fix improper apps not using C2D_MESSAGE permission

SakiiCode:
There can be cases with (poorly written?) apps that they declare the <code>packageName&nbsp;+&nbsp;".permission.C2D_MESSAGE"</code> permission but don't actually use it with the <code>\<uses&#8209;permission\></code> tag. 

That way they don't get the push notification just errors with 
```
Permission Denial: receiving Intent { act=com.google.android.c2dm.intent.RECEIVE flg=0x30
pkg=xxxxxxxxx cmp=xxxxxxxxx/com.google.firebase.iid.FirebaseInstanceIdReceiver
(has extras) } to xxxxxxxxx/com.google.firebase.iid.FirebaseInstanceIdReceiver
require xxxxxxxxx.permission.C2D_MESSAGE due to sender com.google.android.gms (uid 10131)
```

I hope this is a good way of solving it

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
+++ b/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
@@ -61,6 +61,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.net.Socket;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


mar-v-in:
74213474b58ede10b219c13ec17a3fe88077d77e should also solve the issue (but not expose the push message to third parties just because the app didn't declare using it).

SakiiCode:
Not sure why my version exposed the push message to third parties, but your commit works so it's fine.

PR #1088, opened 2020-07-16
Debloat MicroG

Mattis6666:
This removes maps, safetynet and other useless shit. I also removed all permissions other than contacts access which I wasn't able to find.

I tested login and usage with vanced 14.21 (too lazy to resign split apks, thus old .apk). I did not test notifications.

MicroG crashes when clicking on the Google Safetynet tab or UnifiedNlp, rest works as expected.


Mattis6666:
ignore this lol

PR #1092, opened 2020-07-19
Fix GCM heartbeat interval for roaming connections with manual interval

delthas:
When using GCM on a roaming connection, the heartbeat interval is set
to: `networkRoaming  * 6000`. It should instead be
`networkRoaming * 60000` because we're converting from a number of
minutes (stored in the properties) to a number of milliseconds, like
it's done for regular mobile and wifi connections.


mar-v-in:
Thanks, good catch :)

PR #1116, opened 2020-08-03
Exposure notifications - reading from the stream in a loop

kbobrowski:
Hi, `stream.read(byte[] buffer)` does not make an effort to read exactly `buffer.size` bytes, it just tries to read at least 1 byte, and may return with a smaller number of bytes read:

https://docs.oracle.com/javase/7/docs/api/java/util/zip/ZipInputStream.html#read(byte[],%20int,%20int)

This was a bug also present in original Google implementation, which resulted in header validation error on some devices.


mar-v-in:
Thanks!

PR #1153, opened 2020-08-30
Added missing parcel field 8 for polylineoptions. Added gradle compile time option for selecting Mapbox map style.

draekko:
Was having issues with various apps and especially with com.trusty.ty.satellite (satellite tracking app) where the app would only show a white area. After debugging and playing with logcat i found it was failing with the hardcoded styles since i was using my own token key i did not have them available to me. So i added a BuildConfig switch to toggle between default hardcoded ones, the default Mapbox styles, or adding the Uri for a custom style.  The app was also complaining about field 8 for PolylineOptions being unknown so added that in as well. The app still doesn't work with markers showing their information but i guess that's a future battle.

### DIFF
--- a/play-services-api/src/main/java/com/google/android/gms/maps/model/PolylineOptions.java
+++ b/play-services-api/src/main/java/com/google/android/gms/maps/model/PolylineOptions.java
@@ -45,6 +45,8 @@ public class PolylineOptions extends AutoSafeParcelable {
     private boolean visible = true;
     @SafeParceled(7)
     private boolean geodesic = false;
+    @SafeParceled(8)
+    private boolean clickable = false;
 
     public PolylineOptions() {
     }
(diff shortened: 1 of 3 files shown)


draekko:
Update with new patch for adding support for InfoWindow to show/hide the title/snippet with mapbox.

ljl-covid:
"Removed exposure notification"? What am I missing here?

draekko:
Crap was tired yesterday updated the wrong repo!

draekko:
Those EN removal changes changes were supposed to be for my personal local branch only!

Ok removing them locally by pushing a reset and they seem to be gone now here as well, whew!

dimaryaz:
Sorry, this isn't quite related to your change, but do you know where the token key is coming from? How is MapBox supposed to work? If I build MicroG myself, do I somehow end up with a different token than what's in the official builds?

dimaryaz:
To answer my own question: it should be set using "mapbox.key" in "local.properties". And, if you set it to the key from the official build, suddenly everything works! Tried it with Uber, Lyft, and Yelp, and the map shows up now.

Why is the official key not mentioned anywhere? Is it supposed to be secret or something? It's not _that_ hard to extract it...

ale5000-git:
@draekko
All of this seems to be already implemented by other in the current microG.
Could you please check if something is missing?

draekko:
> @draekko All of this seems to be already implemented by other people in the current microG. Could you please check if something is missing?

Dude that was 4 years ago. I don't have that stuff on hand anymore.

ale5000-git:
I will close it for now as it does seems everything is already supported as of now; if there is something missing just open a new one.

Issue #1158, opened 2020-09-04
Exposure Notification (Corona) Scanner Implementation


tommes0815:
**ROM**:
lineage-14.1 unofficial build - Android 7.1.2

**GmsCore**:
dirty build, commit d641ca7e7e9a147f65120c5cb40216c11744b702

**device**:
Sony Z5 compact (E5823)

**Corona app:**
CWA 1.3.0.37

Hi, I am having some issues running the EN API. First I noticed that after running EN for a while the preferences app did not show any collected IDs. So I debugged a bit. My device seems to scan for 30mins and then stops. It only continues scanning after disabling and re-enabling EN or rebooting (basically after the `ScannerService` is restarted). I added some debug output printing `seenAdvertisements` to the callback function [onScanResult](https://github.com/microg/android_packages_apps_GmsCore/blob/fd6d915f0ad1b4b35e17e0b5d844c6fc4a5efcf8/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ScannerService.kt#L69). I see those debug prints for 30min and then stopping. They only restart after restarting the `ScannerService`.

Since nobody else noticed while testing, it might be a problem with my device/bluetooth chip. Nonetheless, it might be worth finding the issue or changing the implementation in case it affects other people. Since there is no error message it might go unnoticed and basically breaking the EN scanning functionality.

While searching for the reason of the bug I also had a look on the Google implementation / description. Please correct me if I am wrong, but I think your current implementation constantly / all the time scans for other devices? The EN system, at least as envisioned from Google/Apple, should actually only scan for a couple of seconds in an interval of a couple of minutes (when the device is woken or every 5min). Do you have a reason for the continues scanning other than to have a quick first release? I guess it is also not very battery friendly (ok, there is a "battery optimization" item on your todo list). Probably my problem would also be solved with a system, which scans only every now and then (where the ScannerService is restarted).

rudestorm:
Not sure about this, but I thought this is part of the battery optimization process mentioned [here](https://github.com/microg/android_packages_apps_GmsCore/issues/1057#issuecomment-668276901)

mar-v-in:
- I agree that scanning continuously is not optimal and that will certainly be optimized later on, however I also don't see a big battery impact on my device at least.
- Do you have "battery optimizations" ignored for microG Services Core according to Self-Check?
[...]

PR #1161, opened 2020-09-08
EN: using scanning loop instead of continuous scanning

tommes0815:
* converting ScannerService to a lifecycleService
* running a simple loop starting / stopping the bluetooth scanner
* introducing SCANNING_INTERVAL as a constant for the loop period
* introducing SCANNING_TIME as a constant for the scanning duration per loop run

This is my first time using kotlin and I am also not very experienced programming for Android, so I probably did a lot of mistakes. I basically used the same concepts of the `AdvertiserService` for the `ScannerService`. The `AdvertiserService` class has a `suspendCoroutine`. I am not exactly sure for what it is used and did not include it in the `ScannerService`. I am also not sure if `ScannerService` needs the `onBind` function?

Probably there are many smarter ways of implementing it, so feel free to ignore it or redo it later on. With theses changes the EN API gets closer to the Google/Apple concept and it also solves #1158 for me. In case you want to stick to continuous scanning, you can set `SCANNING_TIME = SCANNING_INTERVAL` and still solve   #1158.

### DIFF
--- a/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/Constants.kt
+++ b/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/Constants.kt
@@ -11,6 +11,10 @@ import java.util.*
 const val TAG = "ExposureNotification"
 val SERVICE_UUID = ParcelUuid(UUID.fromString("0000FD6F-0000-1000-8000-00805F9B34FB"))
 
+const val SCANNING_INTERVAL = 5 * 60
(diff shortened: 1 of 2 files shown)


mar-v-in (commented):
In your implementation, the scan loop can be executed multiple times, because stopScanLoop does not necessarily stop the scan loop (if you are currently in any of the delay phases) but sets looping to false and restartScan will start a new scan loop setting looping to true again. This will probably cause a bunch of issues...

I'll go with an attempt not making use of suspendable functions first, but this code is going to change anyway once we start supporting "ExposureWindow" mode.

tommes0815:
Thank you for the review. Since you committed your implementation already, we can drop this

PR #1214, opened 2020-10-06
Mapbox: implement predefined user location

vincentvidal:
Mapbox predefined user location display is used by some applications to
automatically show user location on maps.

The two required methods from GoogleMap API are isMyLocationEnabled and
setMyLocationEnabled, which are implemented here by using
LocationComponent class.

The initialization must be performed when the map style is loaded.


vincentvidal:
In Tripadvisor and Park4Night for example, the user location is not displayed. This does not happen in Uber for example. The reason is that these two apps use Mapbox predefined user location display utilities, which is not implemented in GoogleMap.kt
The implementation proposed here works at least for these two apps.
The code should be carefully reviewed though: I had never used kotlin before..

mar-v-in:
Merged and cleaned up in 981301e9c6e236765d53389b3a7fdb689b352a80

PR #1215, opened 2020-10-08
EN: Fix overflow that prevents matching

cgrigis:
In the `finish*Matching()` methods of `ExposureDatabase.kt`, due to an overflow on `Int`, the value compared to `oldestRpi` becomes negative, and the condition is always `true`, ignoring the key.

### DIFF
--- a/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt
+++ b/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt
@@ -339,7 +339,7 @@ class ExposureDatabase private constructor(private val context: Context) : SQLit
         val keys = listSingleDiagnosisKeysPendingSearch(tid, database)
         val oldestRpi = oldestRpi
         for (key in keys) {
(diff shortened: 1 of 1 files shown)


mar-v-in:
`ROLLING_WINDOW_LENGTH_MS` is of type Long since 6afcca0396be456995fa9f960ec1b581f778a942, thus the int to long conversion you made explicit should have been implicit before. I'm not against this change to make it explicit, just skeptical how it can have any impact .

cgrigis:
> `ROLLING_WINDOW_LENGTH_MS` is of type Long since [6afcca0](https://github.com/microg/android_packages_apps_GmsCore/commit/6afcca0396be456995fa9f960ec1b581f778a942), thus the int to long conversion you made explicit should have been implicit before. I'm not against this change to make it explicit, just skeptical how it can have any impact .

Aha, indeed, I did not notice that this part changed after `v0.2.12.203315`, which is the version I have my code based on. On that version it has a dramatic impact. :-)
Regarding explicitness, I believe other places in that file have an explicit `.toLong()` even though multiplied by `ROLLING_WINDOW_LENGTH_MS` (such as in `findMeasuredExposures()`), probably done before the change, but I agree that is no longer necessary here.

mar-v-in:
Merged manually via 4e3c03cfaa45b298f04abba97c534a9224262666

PR #1222, opened 2020-10-10
Add patch for Android 11 (R)

kdrag0n:
This has been tested to work with a third-party app as part of my ProtonAOSP ROM. Thanks to @gudenau for their [Android 10 patch](https://github.com/microg/android_packages_apps_GmsCore/pull/957), which I used as a base.

Note that Google has been deprecating dessert names and version letters since Android 10 in favor of using version numbers in more places, so maybe the patch file should be renamed to `android_frameworks_base-11.patch`.

### DIFF
--- a/patches/android_frameworks_base-R.patch
+++ b/patches/android_frameworks_base-R.patch
@@ -1,0 +1,177 @@
+From 67d661ce0253b099255eef2690f1437df41a8094 Mon Sep 17 00:00:00 2001
+From: Danny Lin <danny@kdrag0n.dev>
+Date: Mon, 5 Oct 2020 12:36:35 -0700
+Subject: [PATCH] Add support for app signature spoofing
+
+This is needed by microG GmsCore to pretend to be the official Google
+Play Services package, because client apps check the package signature
+to make sure it matches Google's official certificate.
+
+This was forward-ported from the Android 10 patch by gudenau:
+https://github.com/microg/android_packages_apps_GmsCore/pull/957
+
+Changes made for Android 11:
+  - Updated PackageInfo calls
+  - Added new permission to public API surface, needed for
+    PermissionController which is now an updatable APEX on 11
+  - Added a dummy permission group to allow users to manage the
+    permission through the PermissionController UI
+    (by Vachounet <vachounet@live.fr>)
+  - Updated location provider comment for conciseness
+
+Change-Id: Ied7d6ce0b83a2d2345c3abba0429998d86494a88
+---
+ api/current.txt                               |  2 ++
+ core/res/AndroidManifest.xml                  | 15 ++++++++++++
+ core/res/res/values/config.xml                |  2 ++
+ core/res/res/values/strings.xml               | 12 ++++++++++
+ non-updatable-api/current.txt                 |  2 ++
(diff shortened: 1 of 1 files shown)


patches/android_frameworks_base-R.patch
```diff
@@ -0,0 +1,177 @@
...
++    <permission-group android:name="android.permission-group.FAKE_PACKAGE"
++        android:label="@string/permgrouplab_fake_package_signature"
++        android:description="@string/permgroupdesc_fake_package_signature"
++        android:request="@string/permgrouprequest_fake_package_signature"
++        android:priority="100" />
++
++    <!-- Allows an application to change the package signature as
++         seen by applications -->
++    <permission android:name="android.permission.FAKE_PACKAGE_SIGNATURE"
++        android:permissionGroup="android.permission-group.UNDEFINED"
```
chris42, line 70: Line 70 needs to be 
`android:permissionGroup="android.permission-group.FAKE_PACKAGE"`
otherwise strings will not be correct in GUI.

chris42:
Also I found, that within the app permissions the signature spoofing permission for microg is correctly shown. However in the overall list of the permission manager 0 apps are shown for signature spoofing.

machiav3lli:
This attitude could also be seen on some ROMs were patched while building which leaves the permission management passively accepting all requests. While with some other patches(there's been couple) you needed to prompt the permission dialog manually through adb shell as the app wasn't patched to handle it. So what you discribe is fully dependant on the signature spoofing patch of the rom not of microG.

chris42:
@machiav3lli Could you elaborate? I do not understand what you are trying to say.

machiav3lli:
@chris42 gladly. In my experience I noticed that A10 ROMs differ on the matter of showing if an app has sigSpoofing. There's ROMs that behave the way you discribed it, while others show that there's an app requesting the permission but can't grant it manualy. Only through adb shell the request dialog could be called: 

`adb shell am start com.android.vending/.GrantFakeSignatureOermissionActivity`

How they're patched is the only variable I could find then(unfortunately, didn't document the diffrrences then). And still didn't test enough number of A11 ROMs to say if it also applies, but we can assume that it's the same.

chris42:
I am not sure how that relates to what I described before. I just noted, that in the permission overview signature spoofing is not showing any apps, while in the app specific it is shown. This is just the list of permissions and is independent of if it has been granted or not.
Regarding how it is patched. Exactly above with this patch in this PR.

I would suspect it has to do with linking of the created signature spoofing category.

kdrag0n:
Corresponding PermissionController patch, not included because it's outside of frameworks/base: https://github.com/ProtonAOSP/android_packages_apps_PermissionController/commit/f7a946c37afe83e57cea03ccd00086ccba2dbdb5

With this patch:

- Requesting the permission in microG opens the runtime permission prompt
- Permission manager says "Spoof package signature: 1 of 1 apps allowed"
- Revoking the permission in permission manager works
- Nothing is granted or ignored silently

chris42:
@mar-v-in Would it make sense to add this and the last patch to the repository?

mar-v-in:
I moved this into https://github.com/microg/patches

PR #1234, opened 2020-10-19
Kotlin Interfaces, Constants, Abstraction, & Enum conversion

Doomsdayrs:
I converted the Enums, Constants, Abstract classes, and Interfaces to kotlin

Does it compile?: Yes
Does it function?: Possibly

Changes:
- Interfaces converted
- Abstract classes converted
- Constant lists converted
- Enums converted
- FuseLocationProviderAPI class due to some incompatibilities between the interloop, using anonymous class
- Removed AsyncTask from GcmListenerService and used kotlin GlobalScope instead, since AsyncTasks are deprecated
- Updated gradle
- Updated gradle dependencies
- Swapped some dependencies to the KTX version


Doomsdayrs:
Pretty sure I did not touch that activity class in the android manifest  hmm

Doomsdayrs:
Maybe I'll help with converting the gradle scripts to KTS

PR #1264, opened 2020-11-11
Update MapsConstants.java

xxxfreez:


ale5000-git:
@xxxfreez: Why do you remove constants?

PR #1270, opened 2020-11-16
Add equals() and hashCode() to Status

cgrigis:
While integrating some microG UI modules bundled with an app, I ran into a situation where sometimes (possibly due to particular sequences of intents/activities), there were various instances of `Status.SUCCESS` around, resulting in failed calls even though they were successful.

This adds a proper `equals()` to remove this possibility.


mar-v-in:
d038fd82cfd43c89dc8efe746e7776ba49a53cd6 is a more appropriate solution to this issue :)

cgrigis:
Indeed, missed the `isSuccess()`, much better :)

PR #1271, opened 2020-11-16
EN: fix AEM version check

cgrigis:
According to the [spec](https://covid19-static.cdn-apple.com/applications/covid19/current/static/contact-tracing/pdf/ExposureNotification-BluetoothSpecificationv1.2.pdf), the AEM version is encoded in bits 7..4 of the first byte, so only those should be checked for validity (I've seen advertisements coming from GAEN with a first byte of e.g. 0x5C).


mar-v-in:
Bits 3..2 contain the tx power confidence needed for [ExposureWindow.getCalibrationConfidence()](https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/ExposureWindow#getCalibrationConfidence()) API. I already have the required changes locally and will upload later today.

mar-v-in:
This should be covered by 4c2ef04364691b04aace378415b9a85c011829d3.

PR #1275, opened 2020-11-19
Always include git commit count in version code

chirayudesai:
* New commit, new build, new version code
* Makes it easy to ship updates

Change-Id: I8f40fad2b176b1f2bba07abaa5ddf0dd8da6689b


chirayudesai:
I did this downstream in CalyxOS since we wanted to ship an interim update with some minor changes (Mapbox API key)

Not sure if you'd want this, but figured I might as well send a PR to find out :)

chirayudesai:
> Starting build for version 0.2.13.204215-16 (e37345a) (204215042)...

From the travis logs

mar-v-in:
The reason why I opted to not do this is that it means that if we have any rebases that change the commit count (e.g. when on a development branch) that results in version code downgrades which are not handled very nice in Android. The current way we have one version code for all interim versions between two releases, which works reasonably fine in most cases (although I see it doesn't work well if you want to release interim versions).

Issue #1052, opened 2020-04-30
Gmail not working and force closing


Pokemetti:
Hello, I use microg with Nanodroid on LineageOS 16 with microg (from lineage.microg.org) as well as stock TouchWiz Marshmallow on my S5 mini. On both phones Gmail force closes as soon as I try to start the app and occasionally force closes in the background. Gmail worked just fine before but since a few weeks it's been like that. The Gmail app is totally unusable like that, but there aren't any alternative apps I know of. I tried clearing data, giving it all permissions, didn't work. It seems to be related to some update. I'll maybe try since which version it doesn't work.

Pokemetti:
I think that the latest working release is 2020.03.29.306383213.release

NolfNolfer:
I'm having this problem too.

It started after I tried to flash a new firmware and a rom update. I reverted back to the old firmware and restored the backup, but even after restoring an backup of the Gmail app it still kept crashing.

NolfNolfer:
> I think that the latest working release is 2020.03.29.306383213.release

BTW, tried that, it doesn't work for me that version.

breversa:
I found solutions to several of my problems here : https://gitlab.com/Nanolx/NanoDroid#issues
You may be lucky and find the solution to yours too. :-)

Pokemetti:
Sadly, seems to not be addressed there

Axolord:
Can confirm, does not work for me either.
As a workaround for now I would promote using another mail client like fairemail for example, although it would be great having gmail itself fixed.

microG version: 0.2.10.200475-dirty via Nanodroid
OS: Havoc 3.3 with Android 10
Devices: Redmi Note 5 Pro

Techwolf:
I upgraded to Android 10 via a fresh install of Havoc-OS 3.7 onto a Motorola G7 power. After installing gmail, it was doing the same as this issue, about three very quick flashes of the small window of app crash/forced closed. Using Titanuium Backup, I install an old backup, both app+data. After opening gmail, close it out and looked at notices, grant account access permission to gmail. Re-opened gmail and it just worked without issue. Gmail version 2019.04.14.246198419

Pokemetti:
Well, as I stated, even version 2020.03.29.306383213 works which is newer than what you mentioned, so that is to be expected.

Mrnofish:
After starting the app, I get a flashing "GMail keeps stopping" dialog.

Lineage with MicroG 20201014
[...]

PR #1283, opened 2020-11-26
Remove the arg restriction from Account ContentProvider get_accounts

dimaryaz:
It appears to fix https://github.com/microg/GmsCore/issues/1052


mar-v-in:
Thanks for investigating this issue and proposing a patch.

Your proposed patch would have allowed Google Apps to read details about non-Google accounts your device (which they shouldn't be allowed to).

With c0324710f3cd647c3706f0a1b222a1871406bf8d this is not possible but also the API will no longer throw an exception when trying to access accounts one was not supposed to but instead return an empty list.

dimaryaz:
Ooh, got it. Thanks!

Issue #1208, opened 2020-10-04
Import and export exposure notification data 


rugk:
## Use case

I have used Google Play Services before.
Now I want to switch to microG.
Because of the Corona/COVID-19-situation I do not want to loose my data.

## How to

What I can currently do:
Go to the COVID screen:
![corona-google](https://user-images.githubusercontent.com/11966684/95024810-c9251600-0685-11eb-9d4f-e9f2e1e5d439.png)

Export the data:
![corona-google-export](https://user-images.githubusercontent.com/11966684/95024819-d215e780-0685-11eb-8c16-b3f03f280f50.png)

**Problem:** After you press that menu entry, you get literally **no** feedback. I have **no idea** where that data is saved. (checking the usual internal storage I could not find anything.)

## Intended solution

Anyway, what we need in microG's implementation is just an import feature for those.
For whatever reason, Google Play Services do not offer any import feature though – maybe so one cannot abuse it and generate fake results. (though AFAIK this is already cryptographically assured, as you cannot easily generate those hashes or respectively only random ones hmm)
Anyway, would be great if we could switch to microG but keeping the 14days history, so one does not loose data that could result in a missed COVID notification.

For completeness you may of course also implement an export feature – I guess if you have an import feature that will be easy. So you'd have feature parity with Google's implementation in this regard then.

mar-v-in:
@rugk do you have any details on the export format? Does it actually contain the history of collected ids or just the history of risk calculations? From the description it sounds like the latter.

rugk:
> do you have any details on the export format?

No, because as said I could not even find the file/where it exports those. :see_no_evil: (Any help appreciated… :see_no_evil: )

rugk:
BTW of course a silly, but technically possible solution to this problem is also: Have another phone with you with Google Play Services for >= 14days (or it may be possible to only do that at the end of the 14 days) and then upgrade your main phone to microG. Afterwards, keep your other phone for 14 days too and then you can also upgrade the other phone to microG. 
If the other phone did not get any notification in these 14 days then it's fine, hehe…

Of course, this is a little weird and not everyone has two phones at their disposal, so this is only a workaround,

mh-:
Sorry to say that, but it’s only a report, not the actual RPI data. 
To get the actual RPI (and your own TEK) data, you need to have access to the GMS files on your device, i.e. root access. 
[...]

PR #1291, opened 2020-11-29
EN: Add database export feature

tlambertz:
This pull requests adds an database-export feature to the exposure-notification preference screen.

I am using MicroG on an unrooted device, so cannot read the database otherwise. By having this export feature, I can use the [corona warn companion app](https://github.com/mh-/corona-warn-companion-android) to create an analysis without root.

Your image title

The export is done using `Intent.ACTION_SEND`. Unfortunately, a `FileProvider` cannot directly share files in the `/databases` folder, so I copy it from `/databases` to `/files` first.
This method allows any file-browser app to capture the SEND intent and write the database to a user specified directory, or (with a companion pull request) open it directly in the corona companion app.

### DIFF
--- a/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt
+++ b/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt
@@ -8,13 +8,16 @@ package org.microg.gms.nearby.exposurenotification
 import android.annotation.TargetApi
 import android.content.ContentValues
 import android.content.Context
+import android.content.Intent
 import android.database.sqlite.SQLiteCursor
 import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteDatabase.CONFLICT_IGNORE
 import android.database.sqlite.SQLiteOpenHelper
+import android.net.Uri
 import android.os.Parcel
 import android.os.Parcelable
 import android.util.Log
+import androidx.core.content.FileProvider
 import com.google.android.gms.nearby.exposurenotification.CalibrationConfidence
 import com.google.android.gms.nearby.exposurenotification.DiagnosisKeysDataMapping
 import com.google.android.gms.nearby.exposurenotification.ExposureConfiguration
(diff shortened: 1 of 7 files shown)


play-services-nearby-core-ui/src/main/res/xml/filepaths.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths>
+    <files-path path="." name="files" />
```
mar-v-in, line 3: Exposing everything in the /files folder is bad practice. Instead, please only expose a specific folder for these exports.

play-services-nearby-core-ui/src/main/res/xml/filepaths.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mar-v-in, line 1: The filename of this file is a global resource name across microg. Thus, the name should be picked such that it is tied to the exposure notification or nearby module that it's used with instead of something very generic like "filepaths".

mar-v-in (commented):
The database file includes the users private TEKs, which I don't think are needed for this kind of export. I'd prefer to only export the collected IDs (in table advertisements), as this is the only thing CWA companion will look at. The export option should then also be moved inside the "Collected IDs" screen to also make it more clear what exactly is exported.

The copied file currently stays in the temporary location forever. As the file can become as large as several hundred MB, it should be cleaned up after being exported. Also, as this is a temporary file, it would probably be better to make use of the cache folder that can be easily cleaned up from within system settings.

The file provider and export logic should probably be in play-services-nearby-core, only the UI aspects (option in settings, starting activity for export) should be in play-services-nearby-core-ui.

tlambertz:
Thank you for the quick review! I have cleaned it up according to your suggestions.

The code is now copying only the advertisements table for export. I agree that this is more privacy friendly, and it fits my use-case perfectly. But this will reduce the usability for folks wanting a backup feature. Though given that is is not automatic, it is of questionable use for that anyway.

What parts of the table grow to 100+MB? Shouldn't a few advertisements be fairly small? Or do you store all of the "known infected" codes as well?

The main feature I want from CWA-companion is the timeline-view for each individual risk encounter. I am not sure if MicroG is a suitable place to implement this, given that it has to interact with the national databases?

rugk:
> But this will reduce the usability for folks wanting a backup feature

I doubt many people would use/consider it as such/one.
The use case for #1208 for was mainly being able to switch between phones / systems (StockRom CustomRom or CustomRom to another CustomRom or so) and keeping the data. There you plan in advantage and backup the files to later import them. (Of course, i know the import would still be missing here.)

ljl-covid:
Indeed I think the TEKs should also be exported for potential future importing. Having exposure notifications enabled basically "locks you in" to the phone and ROM you're using it: if you want to move to a different ROM or start using another phone entirely (and the previous phone will get formatted to be used by someone else), you'd want to still have the TEKs in case you get COVID and get tested in the 14 days that follow, which is not exactly something that can be foreseen.

So for now while there is no import feature, I guess it doesn't matter, but in terms of #1208 it would be beneficial to have it. Privacy-wise, if the phone ends up in the wrong hands, the presence of an "export" feature for the TEKs is unlikely to be the thing that causes a bit privacy breach...

play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt
```diff
@@ -901,6 +904,53 @@ class ExposureDatabase private constructor(private val context: Context) : SQLit
...
+            }
+
+            // Open a sharesheet
+            if (fileUri != null) {
+                // Grant temporary read permission to the content URI
+                val sendIntent: Intent = Intent().apply {
+                    action = Intent.ACTION_SEND
+                    putExtra(Intent.EXTRA_STREAM, fileUri)
+                    addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                    type = "application/vnd.sqlite3"
```
mar-v-in, line 946: Can we use a more specific file type such that https://github.com/mh-/corona-warn-companion-android/pull/97 can specifically listen for that file type? Suggestion: `application/vnd.microg.exposure+sqlite3`

mar-v-in: @tlambertz have seen this? any opinion?

tlambertz: I do not have a strong opinion on this one. Using the default sqlite3 type would allow a user to open the database with an sqlite3 browser directly.. though I am not aware of any app that only listens to this specific type. I'll change it.

mar-v-in:
Merged (4158e719996da307a22f8d25f63727358eac91a6, 0a9d4869c7903d86c9e00033a48c5191d6a2e70d). Thanks!

PR #1299, opened 2020-12-02
EN: ensure token is non-null in provideDiagnosisKeys()

cgrigis:
This is a suggested fix for #1300. There might be better ways. :)


mar-v-in:
I added `@Nullable` annotation in 2b09a5e530fc9733b41063ce8fcf88007a896d84 which caused the Kotlin compiler to find this error as well. I also handled another issue that can occur when `provideDiagnosisKeys()` is invoked before `setDiagnosisKeysDataMapping()` in ExposureWindow mode.

cgrigis:
Thank you!

PR #1302, opened 2020-12-07
Add Japanese Translation

yuk7:
I wrote a Japanese Translation.
Can you tell me if there are any necessary corrections?

I'm a microG user and I'm very grateful for your work.


mar-v-in:
Thanks!

Issue #1179, opened 2020-09-17
Using microG EN as library


cgrigis:
I am attempting to use microG as a library for the SwissCovid app as has been [suggested](https://github.com/microg/android_packages_apps_GmsCore/issues/1057#issuecomment-685972245) by @mar-v-in.

My apologies in advance if I am doing something silly, just starting up with microG...

What I have done so far is remove the dependencies on GMS in my project, put the relevant modules in (`play-services-core` etc.), implement `provideDiagnosisKeys()` in the client lib, and minor fixes until the app compiles. Now the app starts running, but as soon as it tries to start EN (sending `com.google.android.gms.nearby.exposurenotification.START` Intent), it fails to bind in `MultiConnectionKeeper.this.context.bindService()` (returns `false`), eventually returning `API_UNAVAILABLE` to the app. I have tried to change GMS_PACKAGE_NAME to `org.microg.gms` or the app package, with no success.

From logcat:
```
2020-09-17 15:14:18.645 4340-5972/? W/ActivityManager: Unable to start service Intent { act=com.google.android.gms.nearby.exposurenotification.START pkg=org.microg.gms } U=0: not found
```

My questions are:

* Any hints on what I should be changing to make it work?
* Can the app as such even run on a phone with Google GMS installed (which is what I am testing on right now)? Should I rename all com.google.android.gms to avoid conflicts?
* Am I missing something to enable microG, some sort of init()?

Thank you in advance for any help!

cgrigis:
OK, some progress.
* I had to change `GMS_PACKAGE_NAME` to the **full** package name of the app :man_facepalming: [wears brown paper bag over head].
* After that, the service was binding, but would return `FAILED_UNAUTHORIZED` in `ExposureNotificationService.handleServiceRequest()`. I fixed this by using `checkCallingOrSelfPermission()` instead of `checkCallingPermission()`; not sure yet if that is the proper way (is this because, unlike normal usage of microG, in this case the caller and the service are in the same process?)
* After a few more dependencies fixes, it looks like the app is running.

Now on to see if it does what it should be doing. :)

mar-v-in:
Just FYI, the following modules should be enough for your usecase:
```
play-services-base-api
play-services-base-core
play-services-base
play-services-basement
play-services-nearby-api
play-services-nearby-core-proto
play-services-nearby-core
play-services-nearby
play-services-tasks
```

cgrigis:
[...]

PR #1304, opened 2020-12-09
EN: Declare EN_SETTINGS intent filter also in play-services-nearby-core-ui

cgrigis:
Still part of #1179: the DP3T-SDK checks the availability of EN on the phone by checking whether the `ACTION_EXPOSURE_NOTIFICATION_SETTINGS` intent resolves to an activity (it does not send the intent, just checks whether it resolves).

So far, I have worked around this by declaring the intent filter in the DP3T project, as it is not declared in the microG libraries I use (it's in `play-services-core`).

I was wondering if that filter could also be declared in `play-services-nearby-core-ui`, with a lower priority? As I understand the way priorities work (but I could be  wrong), it should still result in the `SettingsActivity` be called when both are available.

Does it make sense, or is it a bad idea? :)
Or should the SDK just check EN availability differently?

### DIFF
--- a/play-services-nearby-core-ui/src/main/AndroidManifest.xml
+++ b/play-services-nearby-core-ui/src/main/AndroidManifest.xml
@@ -18,6 +18,11 @@
             <intent-filter>
                 <action android:name="org.microg.gms.nearby.exposurenotification.CONFIRM" />
             </intent-filter>
+            <intent-filter android:priority="-100">
+                <action android:name="com.google.android.gms.settings.EXPOSURE_NOTIFICATION_SETTINGS" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
         </activity>
     </application>
 </manifest>


mar-v-in:
If we add this intent filter, we have to be able to handle it properly. Just adding it like this, even with low priority, can cause issues down the road.

But I also think that the behavior of DP3T SDK is not very good either. Resolving activities is generally discouraged and is made "harder" on Android 11 (see <https://medium.com/androiddevelopers/package-visibility-in-android-11-cc857f221cd9>). IMO the best way to check if EN is available is to call [ExposureNotificationClient.isEnabled()](https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/ExposureNotificationClient#isEnabled()), which either returns true or false (information you'd probably want to have anyway) or throws an ApiException when the service is either not installed or not available to the calling app.

mar-v-in:
I guess the best thing to do here is to implement the intent using a lower priority and without exporting it. The launched activity should display the microG EN settings UI. I'll look into doing this before the next release.

mar-v-in:
This is now realized as 3dad397dd150c2f1587ee0dbbee448c6ede81ee1.

cgrigis:
> If we add this intent filter, we have to be able to handle it properly. Just adding it like this, even with low priority, can cause issues down the road.

Indeed, I was not very comfortable with it, thanks for the confirmation.

> But I also think that the behavior of DP3T SDK is not very good either. Resolving activities is generally discouraged and is made "harder" on Android 11 (see https://medium.com/androiddevelopers/package-visibility-in-android-11-cc857f221cd9). IMO the best way to check if EN is available is to call [ExposureNotificationClient.isEnabled()](https://developers.google.com/android/reference/com/google/android/gms/nearby/exposurenotification/ExposureNotificationClient#isEnabled()), which either returns true or false (information you'd probably want to have anyway) or throws an ApiException when the service is either not installed or not available to the calling app.

That makes sense...

> I guess the best thing to do here is to implement the intent using a lower priority and without exporting it (thus restricting it to apps bundling microG). The launched activity should display the microG EN settings UI. I'll look into doing this before the next release.
> This is now realized as [3dad397](https://github.com/microg/GmsCore/commit/3dad397dd150c2f1587ee0dbbee448c6ede81ee1).

Oh! Actually I did not mean to have this included if it did not make sense (I did not mind having to add it on the DP3T side), just wondering "out loud"... but thank you for including it!! :bow:

mar-v-in:
I think it makes sense this way, as it allows apps that integrate the library to still call the microG EN settings screen if they had used the intent to show the settings (independent of DP3T resolving this intent to indicate existence of EN API).

Anyway, I guess I can close this.

cgrigis:
Understood. Thanks.

PR #1305, opened 2020-12-09
Update russian translation

Vavun:


mar-v-in:
Thanks :)

Vavun:
Hi @mar-v-in , I have one question.

How about renaming **Google Cloud Messaging** to **Firebase Cloud Messaging**?
Or even simplify to **Cloud Messaging**

From time to time in microG Russian community, people ask whether Firebase Cloud Messaging is supported or ask how messages will work if Google Cloud Messaging is no longer supported by Google.

I mean, that such names are sometimes introduces new users confused.

At that moment I'm preparing a Belarusian translation and I would like to make it easier to understand how microG works. For any languages.

Thanks for your work ☺️

mar-v-in:
I renamed it to just "Cloud Messaging" with b842af98faea5ede7e63774b78a8b93253972e5c.

Generally, I think that instead of using brand names, we should shift towards a more descriptive name of the feature. Like just "push notifications". As we now have descriptive texts at the bottom of each category, those could point out what Google service is used (if any) to realize that feature.

Vavun:
> 
> 
> I renamed it to just "Cloud Messaging" with [b842af9](https://github.com/microg/GmsCore/commit/b842af98faea5ede7e63774b78a8b93253972e5c).
> 
> Generally, I think that instead of using brand names, we should shift towards a more descriptive name of the feature. Like just "push notifications". As we now have descriptive texts at the bottom of each category, those could point out what Google service is used (if any) to realize that feature.

Thanks for the answer. According to your recommendations, I have simplified the translation a little.
https://github.com/microg/GmsCore/pull/1311

Issue #1307, opened 2020-12-10
[EN] Signatures of diagnosis keys received from backend are not verified


cgrigis:
The exposure key files provided by servers are [signed](https://developers.google.com/android/exposure-notifications/exposure-key-file-format#file-sig) using a PHA-specific private key, whose public counterpart is provided to Google for allowlisted apps.

As microG does not have the PHA-specific public signature keys, the exposure key files are currently not verified before matching is attempted. This has of course security implications, as processing data received from untrusted sources without validating it first can be abused.

I was wondering about how to add this feature in a way that is compatible with microG.
Obtaining the PHA public keys, being public :-), should be possible without compromising security (TBC, I am trying to obtain the backend key for Switzerland).
Assuming we have those public keys, we can verify the signatures, by matching the verification key to use with the package name of the app providing the keyfiles (this seems to be the way GAEN does it).

In order to handle the case of custom apps and backends, as well as the case of missing keys, we could have the option to verify keyfile signatures accessible to the user (i.e. enable/disable). In case the feature is disabled, no verification takes place and keyfile handling proceeds as it does now. If the feature is enabled, keyfiles are verified, and matching only happens if the signature is valid.

Another approach I was thinking of would be to have the app somehow provide to the EN implementation the public key(s) needed to verify the keyfiles: if the app does not provide any key, the processing is the same as now; if keys are provided, only the keyfiles with verified signature are processed. I am however not clear on how the app could provide this to the EN implementation...

I have started a [branch](https://github.com/microg/GmsCore/compare/master...c4dt:verify-keyfiles-signatures) to explore how keyfile verification could work.

What are your thoughts?

mar-v-in:
I assume that apps using EN API will get their keys from trusted servers using TLS encrypted connections and/or otherwise verify authenticity of keys before passing them to microG to match them with collected IDs. I don't see why it would be necessary to do this within microG.

As you say this has security implications, can you tell what kind of abuse you are thinking of that is prevented by this signature?

cgrigis:
While, as you say, the apps will get their keys from the back-end using TLS, this only guarantees the key files are provided by the correct source. In case of a breach of the back-end server, an attacker could plant malicious files, which will be delivered unnoticed through TLS.

[...]

PR #1391, opened 2021-02-09
[EN] Verify keyfiles signatures

cgrigis:
See #1307

### DIFF
--- a/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureNotificationServiceImpl.kt
+++ b/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureNotificationServiceImpl.kt
@@ -48,14 +48,14 @@ class ExposureNotificationServiceImpl(private val context: Context, private val 
     // Table of back-end public keys, used to verify the signature of the diagnosed TEKs.
     // The table is indexed by package names.
     private val backendPubKeyForPackage = mapOf<String, String>(
-            Pair("ch.admin.bag.dp3t.dev",
-                    "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEsFcEnOPY4AOAKkpv9HSdW2BrhUCWwL15Hpqu5zHaWy1Wno2KR8G6dYJ8QO0uZu1M6j8z6NGXFVZcpw7tYeXAqQ=="),
-            Pair("ch.admin.bag.dp3t.test",
-                    "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEsFcEnOPY4AOAKkpv9HSdW2BrhUCWwL15Hpqu5zHaWy1Wno2KR8G6dYJ8QO0uZu1M6j8z6NGXFVZcpw7tYeXAqQ=="),
-            Pair("ch.admin.bag.dp3t.abnahme",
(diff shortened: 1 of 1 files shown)


ale5000-git:
@cgrigis: Does it work also for [Corona Warn App](https://github.com/corona-warn-app/cwa-app-android)?

cgrigis:
> @cgrigis: Does it work also for [Corona Warn App](https://github.com/corona-warn-app/cwa-app-android)?

@ale5000-git Not currently, because we don't have the public key from the German authorities. If anybody can obtain this, it will be trivial to add support. :)

ale5000-git:
Corona Warn App support the transnational database that cover many European countries together, I'm interested in this rather than the German one.

cgrigis:
Ah, my mistake. Nonetheless, we need whatever public key allows for their signature verification. Currently I only have the one from the Swiss PHA for SwissCovid — `ch.admin.bag.dp3t`.

cgrigis:
@ale5000-git Do you know of a contact who could provide us with the public key for Corona Warn App?

ale5000-git:
@cgrigis

Apparently there is
- `de.rki.coronawarnapp-dev` for testing
- `de.rki.coronawarnapp` for production

Some info here:
https://github.com/corona-warn-app/cwa-server/blob/main/docs/DISTRIBUTION.md
https://github.com/corona-warn-app/cwa-server/tree/main/docker-compose-test-secrets

This is a bit confusing for me, does it include what you need?

cgrigis:
@ale5000-git 

Thank you for the pointers, that is useful information.

> https://github.com/corona-warn-app/cwa-server/tree/main/docker-compose-test-secrets
> 
> This is a bit confusing for me, does it include what you need?

Unfortunately (and understandably), the above contains the data for dev/test. What we need is the public key corresponding to the private key used by TSI to sign the key material sent by the production back-end...

play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureNotificationServiceImpl.kt
```diff
@@ -25,18 +26,53 @@ import org.json.JSONObject
...
+        } catch (e: Exception) {
+            Log.w(TAG, "Failed to retrieve back-end public key for ${packageName}: " + e.message)
+            null
+        }
+    }
+
+    // Table of supported signature algorithms for the diagnosed TEKs.
+    // The table is indexed by ASN.1 OIDs as specified in https://tools.ietf.org/html/rfc5758#section-3.2
+    private val sigAlgoForOid = mapOf<String, Signature>(
+            Pair("1.2.840.10045.4.3.2", Signature.getInstance("SHA256withECDSA")),
```
mar-v-in, line 72: Either map these to the name string (e.g. "SHA256withECDSA") and only run `Signature.getInstance()` when needed or map to a function that creates instances of Signature (by surrounding the invocation of `Signature.getInstance()` with curly braces).

PS: I very much like the notation of kotlin infix to for map creation, like
```kotlin
val sigAlgoForOid = mapOf<String, Function0<Signature>>(
  "1.2.840.10045.4.3.2" to { Signature.getInstance("SHA256withECDSA") },
  "1.2.840.10045.4.3.4" to { Signature.getInstance("SHA512withECDSA") }
)
```

cgrigis: I changed it as you suggested (indeed, the `... to ...` notation is quite nice, changed my other map accordingly :) ).

It is a trade-off though, exchanging the unused instances in memory for a new instantiation for every signature verification, right?

PR #1399, opened 2021-02-20
Hide legend if no data is to be displayed

### LINKED ISSUE #1383: "0 - -1" IDs per hour
**Describe the bug**
The ID collection chart shows "0 - -1 IDs per hour" in its legend when no data is available.

image

**To Reproduce**
Delete all microg exposure notification data, then open the new ID chart.

**Expected behavior**
Maybe this part of the legend could be hidden. There's really no great solution otherwise; while "0 to -1" is technically true for all boxes that are colored this way, it doesn't really make sense and "0 - 0" would probably be better.

**System**
Android Version: 5.0
Custom ROM: emulator

fynngodau:
This fixes #1383 by not drawing the legend if the highest datapoint has a value of less than zero. It looks like this then:

image


Wangtt185 (commented):
1

Wangtt185 (commented):
1

Issue #1441, opened 2021-04-10
Grid of collected contact Ids does not show details / wrong row header


di-sc:
**Bug description**
As shown in the picture below there are contacts collected (**4**) but these are not marked in the grid below.
I would expect that the contacts are marked in the calendar grid in a different color.

Additionally or related to the issue: instead of showing the month in the row header the current minutes are shown.

**To Reproduce**
1. Install the CCTG
2. Have some contacts with people who also have installed the CCTG or CWA
3. Open the "Begegnungen" view

**Expected behavior**
Show detailed information of collected Ids in the calendar grid
Show a correct row header

**Screenshots**
![image](https://user-images.githubusercontent.com/54937900/114271085-f1fcde00-9a0f-11eb-989d-b641fb0371ed.png)

**System**
ZTE Blade L8
Android Version: 9
Custom ROM: ?

fynngodau:
Just 4 collected IDs is definitely too low in case you were at any place at which you'd expect to find someone who also has the app.

Here's the code the matrix graph uses to determine the format that it uses for the column description:

https://github.com/microg/GmsCore/blob/91b0f6893f9848354af7654b3cebc842615bcd7f/play-services-nearby-core-ui/src/main/kotlin/org/microg/gms/nearby/core/ui/DotChartView.kt#L41-L46

From the documentation, it looks like `DATE_FORMAT` is supposed to be one of these values (or maybe also something similar that is appropriate) depending on your locale: `mm/dd/yyyy`, `dd/mm/yyyy`, `yyyy/mm/dd`

This is then used with `SimpleDateFormat` by microG (unless the string is empty), which is incorrect, because `mm` in `SimpleDateFormat` means "Minute in hour", whereas `MM` means "Month in year". Nice find @di-sc!

fynngodau:
I opened PR #1443 to address the issue I described above.

I'm wondering what is up with the chart in the screenshot, as normally, the first and last box are moved according to the current time (I noticed that the screenshot was recorded in the middle of the day). Compare it to this screenshot to see what I mean (12-hour time format): 
![image](https://user-images.githubusercontent.com/16943720/108603224-083ed400-73a7-11eb-8a40-3c73285ce1ea.png)

di-sc:
> Just 4 collected IDs is definitely too low in case you were at any place at which you'd expect to find someone who also has the app. This might mean that the app is not collecting IDs correctly.

[...]

PR #1443, opened 2021-04-10
Fix date format being used incorrectly

fynngodau:
Addresses #1441. 

The issue detailed in https://github.com/microg/GmsCore/issues/1441#issuecomment-817174607 is fixed in this PR because it replaces `'m'` with `'M'`.

However I'm not sure under which circumstances the operating system actually provides a `DATE_FORMAT` value. In the emulator nothing was provided. I also think that this string does not provide a great format because it always separates the digits with `/`, which is rather American and not fitting German, for example, where usually `.` is used.

### DIFF
--- a/play-services-nearby-core-ui/src/main/kotlin/org/microg/gms/nearby/core/ui/DotChartView.kt
+++ b/play-services-nearby-core-ui/src/main/kotlin/org/microg/gms/nearby/core/ui/DotChartView.kt
@@ -10,14 +10,12 @@ import android.annotation.TargetApi
 import android.content.Context
 import android.graphics.*
 import android.os.Build
-import android.provider.Settings
-import android.text.TextUtils
+import android.text.format.DateFormat
 import android.util.AttributeSet
 import android.view.MotionEvent
 import android.view.View
 import org.microg.gms.nearby.exposurenotification.ExposureScanSummary
 import org.microg.gms.ui.resolveColor
-import java.text.SimpleDateFormat
 import java.util.*
 import kotlin.math.max
 import kotlin.math.min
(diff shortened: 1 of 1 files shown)


play-services-nearby-core-ui/src/main/kotlin/org/microg/gms/nearby/core/ui/DotChartView.kt
```diff
@@ -38,11 +38,11 @@ class DotChartView : View {
             val now = System.currentTimeMillis()
             val min = now - 14 * 24 * 60 * 60 * 1000L
             val date = Date(min)
-            val format = Settings.System.getString(context.contentResolver, Settings.System.DATE_FORMAT);
+            val format = Settings.System.getString(context.contentResolver, Settings.System.DATE_FORMAT)
```
di-sc, line 41: Why using the DATE_FORMAT at all?
It seems to be outdated and not compatible with the SimpleDateFormat (because of the missused 'm').
getMediumDateFormat is already in the context's locale.

fynngodau: As described I don't know and am not happy with the output it provides either. We'll need to ask @mar-v-in who wrote this code.

mar-v-in: I don't remember exactly, but there was something with users being able to pick the date format manually in system settings on some Android versions which would not be reflected by locale based formatting. I guess I'd be fine with just using the getMediumDateFormat and ignore system DATE_FORMAT.

fynngodau: I've updated the PR accordingly.

mar-v-in:
Thanks!

PR #1452, opened 2021-04-17
Handle GMS intent for launching account settings

Fs00:
This small change prevents Google apps from crashing due to an ActivityNotFoundException when the user clicks on "Manage Google account" button. The corresponding intent is now handled by AccountSettingsActivity (currently used as a placeholder for account settings).


mar-v-in:
Thanks.

shahirtj:
لماذا كل هذا التعقيد

PR #1464, opened 2021-04-29
Omit RPIs in log

fynngodau:
As discussed, we shouldn't log current RPIs to the system log. In case they are needed, they can still be accessed through the user interface.


fynngodau:
Thanks @mar-v-in, added to PR.

PR #1483, opened 2021-05-28
Fix some push issues by adjusting heartbeat interval

grote:
When visiting settings UI and one is not checked in, this will also cause a check-in to be performed (if enabled). In case internet is missing when check-in is attempted, it will be retried once internet is back.

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/ui/SettingsFragment.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/ui/SettingsFragment.kt
@@ -5,6 +5,7 @@
 
 package org.microg.gms.ui
 
+import android.content.Intent
 import android.content.Context
 import android.os.Bundle
 import androidx.lifecycle.lifecycleScope
@@ -13,6 +14,8 @@ import androidx.preference.Preference
 import com.google.android.gms.R
 import org.microg.gms.checkin.getCheckinServiceInfo
 import org.microg.gms.gcm.GcmDatabase
+import org.microg.gms.gcm.McsConstants.ACTION_RECONNECT
+import org.microg.gms.gcm.TriggerReceiver
 import org.microg.gms.gcm.getGcmServiceInfo
 import org.microg.gms.safetynet.getSafetyNetServiceInfo
 import org.microg.nlp.client.UnifiedLocationClient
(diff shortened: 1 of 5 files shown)


play-services-core/src/main/kotlin/org/microg/gms/ui/SettingsFragment.kt
```diff
@@ -56,24 +60,30 @@ class SettingsFragment : ResourceSettingsFragment() {
...
 
     private suspend fun updateDetails() {
-        if (getGcmServiceInfo(requireContext()).configuration.enabled) {
+        val context = requireContext()
+        val gcmServiceInfo = getGcmServiceInfo(context)
+        if (gcmServiceInfo.configuration.enabled) {
             val database = GcmDatabase(context)
             val regCount = database.registrationList.size
+            // check if we are connected as we should be and re-connect if not
+            if (!McsService.isConnected(context)) {
```
grote, line 69: I think `McsService` is in the persistent process, so we can't/shouldn't do this check here like this, right?

grote: I changed this to use the `gcmServiceInfo` directly.

chirayudesai:
I've been running this for a few months now, and it's working well.

Push notifications would be delayed sometimes before but now they're pretty much instantaneous.

eggplanting:
> I've been running this for a few months now, and it's working well.
> 
> Push notifications would be delayed sometimes before but now they're pretty much instantaneous.

Tested your updated APK posted here:
https://www.reddit.com/r/CalyxOS/comments/pkvx81/microg_push_notification_fixes_test_apks/

Only with them push notifications seem to work first time to me, without any issues.

I guess many people would appreciate if the fixes will be integrated in the next stable release.

Thank you all for making this possible!

emkangail (approved):

play-services-core/src/main/kotlin/org/microg/gms/ui/SettingsFragment.kt
emkangail: Thanks

ale5000-git:
@grote: The CI now report a compilation error.

grote:
Looks like @mar-v-in merged master in here. I can rebase this on master and resolve issues.

mar-v-in:
@grote that would be amazing :)

PR #1485, opened 2021-06-03
Refactor settings access to use a SettingsProvider

grote:
to enable safe usage of settings no matter which process is getting/setting them.
Previously, different processes were accessing settings in an unsafe way and the warn methods were throwing Runtime exceptions which went largely unnoticed, but happened, especially on a fresh start of the OS.

This is not yet touching SafetyNet settings which at present are only accessed from UiThread.

Change-Id: Ie4134e7be2a7ca4a373790f45fbcbd09bf02ad86


Oizaro:
@grote

`Permission Denial: reading org.microg.gms.settings.SettingsProvider uri content://org.microg.gms.settings/auth from pid=9904, uid=10067 requires the provider be exported, or grantUriPermission()`

PR #1486, opened 2021-06-03
172yevwvsbwu

Sergo19911991:
vggggg


Sergo19911991:
Uy

Sergo19911991:
Hayward e

Sergo19911991:
Huawei p40 lite

ale5000-git:
I'm not sure if it is easy to do but the Google enforcement is only active for new ROMs, so instead of faking the device it is probably easier to fae the ROM date to something like 2017/2018.

ale5000-git:
@Sergo19911991: microG GmsCore can now fake different device.

PR #1487, opened 2021-06-04
Make the application persistent

chirayudesai:
* Should help with not having services killed
* System app only, but shouldn't cause any errors if it's a user-app

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -114,6 +114,7 @@
         android:fullBackupOnly="true"
         android:extractNativeLibs="false"
         android:forceQueryable="true"
+        android:persistent="true"
         android:icon="@mipmap/ic_core_service_app"
         android:label="@string/gms_app_name"
         android:theme="@style/Theme.AppCompat.DayNight">


chirayudesai:
Untested, would like to hear thoughts though. Might not be needed after all, but I still figured I'd put this up.

mar-v-in:
AFAIR, for many purposes, this flag does not add a lot beyond the battery optimization exempt. It also should only affect system apps. But it's unlikely to make anything worse, so I'll just merge it for now.

mar-v-in:
Turns out persistent apps can't be updated at runtime when installed on /system (even if the app on /system had no persistent flag). They still should be updatable using staged updates or by updating the apk on /system, but those really are not user-friendly alternatives, so I'm going to revert this.

chirayudesai:
Sorry about that, I can confirm it doesn't adb install even if the app on /system has the persistent flag)

`Failure [INSTALL_FAILED_INVALID_APK: Package com.google.android.gms is a persistent app. Persistent apps are not updateable.]`

Will explicitly test updates before making any such changes.

mar-v-in:
FYI, you can still update by using staged updates as described here: https://github.com/microg/GmsCore/issues/1503#issuecomment-872097477

Issue #1519, opened 2021-07-15
Granting the permissions required for EN does not work on Android 11


cgrigis:
**Describe the bug**
Granting the permissions required for EN does not work on Android 11

**To Reproduce**
1. Install an EN-using app such as CCTG or LibreCHovid
2. Go through the onboarding
3. When the "Turn on Exposure Notifications" popup shows, click on "GRANT"
4. Nothing happens

**Expected behavior**
The permissions for the app show up and the user can grant permissions.

**System**
Android Version: 11
Custom ROM: LineageOS 18.1

**Additional context**
The problem has been verified with both CCTG and LibreCHovid on LineageOS 18.1, and for LibreCHovid on a stock Android 11 (Samsung A40). For some reason, CCTG does not exhibit the problem on stock Android 11.

When the problem happens, the device log shows:

```
107-15 13:13:22.761  3562  3562 E GrantPermissionsActivity: Apps targeting 30 must have foreground permission before requesting background and must request background on its own.
```

Not sure if it is related to #1440

fynngodau:
Further reading:

* https://developer.android.com/about/versions/11/privacy/location#background-location
* https://developer.android.com/training/location/permissions#request-background-location

ale5000-git:
The permission granting is fixed inside microG, about the other apps you should wait new versions of themself.

lgommans:
I see the same behavior as described by OP, but on stock Samsung without microG. (I'd love µG but also I want the stock camera and ui so... afaict there is no way to patch sig spoofing onto stock android since android 10 :/) It seems that this problem is not (only) on µG's side.

fynngodau:
For CCTG, please update to 2.8.0.0 from our repo: https://bubu1.eu/cctg

ale5000-git:
Is it working fine for everyone?
Can we close this issue?

residuum:
Not sure, if it is a problem with CCTG or GmsCore, but exposure notification is not working using the latest build from LineageOS for MicroG (Lineage build: lineage_FP3-userdebug 11 RQ3A.210805.001.A1 eng.root.20210903.215601 dev-keys, microG Services Code Version 0.2.22.212658) and CCTG (version 2.8.0.0).

When I try to turn on contact tracing, a modal dialog appears to activate exposure notification. When I try to give the additional rights to microG Services Core, nothing happens in the UI. In logcat I see result 5 for GrantPermissionsActivity.

Here is an excerpt from logcat:

[...]

PR #1520, opened 2021-07-15
EN: Fix permission granting on Android 11

cgrigis:
An attempted fix for #1519, inspired by the fix for https://github.com/home-assistant/android/issues/1039

It is not ideal as it requires a double operation by the user: first granting the foreground location, then the background :-/

Tested on LineageOS 17.1, LineageOS 18.1 and stock Android 11 (Samsung A40).


mar-v-in:
Thanks.

PR #1521, opened 2021-07-16
SettingsProvider: Remove logs

chirayudesai:
* This is really spammy!

Change-Id: If1b9cbaa50bf8af37ca7146c6f510ad6875bab35


chirayudesai:
Cc: @grote

grote:
yeah, if everything works as expected they can be removed, but their spammyness might also hint at some bugs that access settings too frequently.

RobertoWa:
what about a on/off switch for logging?

mar-v-in:
Thanks.

PR #1523, opened 2021-07-21
Remove an outdated and misleading log message

Fs00:
This PR removes an outdated log message that was left by mistake after commit aee45c2.


mar-v-in:
Thanks.

PR #1534, opened 2021-08-18
Add Stadia to known Google packages

chirayudesai:
Change-Id: Ie70cde85e398dc5eed13fc2294f78088661a743a


chirayudesai:
Need to test further to see if this actually works, but this was making it stuck early on.

mar-v-in:
Thanks. Definitely doesn't hurt.

PR #1541, opened 2021-08-24
Add the app which was denied push permission to log message

Fs00:
This small change enriches the log message to see exactly which app was denied push permission.


jairus1988:
implement

PR #1582, opened 2021-10-13
Work around requireContext() crashes

grote:
Work around rare crashes that can happen when fragments are not attached to a context while co-routines are still executing but requiring a context.

Change-Id: Ie6c7cee50014b59c25384d3bf9a122081b9917fc


fynngodau:
This sounds related to https://codeberg.org/corona-contact-tracing-germany/cwa-android/issues/134.

PR #1605, opened 2021-11-03
Notice push connection errors faster and fix re-connects

grote:
Draft, because based on #1483

Please see individual commit messages for details.


play-services-core/src/main/java/org/microg/gms/gcm/McsInputStream.java
```diff
@@ -57,7 +56,7 @@
     private int streamId = 0;
     private long lastMsgTime = 0;
 
-    private boolean closed = false;
+    private volatile boolean closed = false;
```
grote, line 59: This gets accessed from more than one thread, so at the very least it needs to be `volatile`.

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -434,11 +444,14 @@ private synchronized void connect() {
         try {
             closeAll();
             ConnectivityManager cm = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
-            activeNetworkPref = GcmPrefs.get(this).getNetworkPrefForInfo(cm.getActiveNetworkInfo());
-            if (!GcmPrefs.get(this).isEnabledFor(cm.getActiveNetworkInfo())) {
+            NetworkInfo activeNetworkInfo = cm.getActiveNetworkInfo();
+            activeNetworkPref = GcmPrefs.get(this).getNetworkPrefForInfo(activeNetworkInfo);
+            if (!GcmPrefs.get(this).isEnabledFor(activeNetworkInfo)) {
```
grote, line 449: Here, I am just pulling out `activeNetworkInfo` into a variable.

PR #1606, opened 2021-11-08
request BLUETOOTH_SCAN and ADVERTISE permissions on Android 12

### LINKED ISSUE #1591: Exposure Notifications need permission changes for Android 12
Something in Android 12 changed with how you access the bluetooth scanning functionality[1] and the migration works poorly currently:

https://codeberg.org/corona-contact-tracing-germany/cwa-android/issues/169

Haven't yet done much investogation,

[1]: https://www.xda-developers.com/android-12-location-scan-nearby-bluetooth-devices/

Bubu:
This also adds a warning notification when the app doesn't have the
required permission after an OS update.

Closes #1591 

Tested to work on Android 12 while using it as a library in CCTG. I tested the upgrade case by initially not requesting the nearby permission at activation time. At that point we should have the same state as someone who activated it with a previous version. The notification shows, can be clicked, leading to the microG EN settings page, which now shows a new indicator for this problem. Clicking that asks for the correct permission and the indicator and notification disappear.

image

As noted in #1591 I really want to get rid of/improve the notification access permissions prompt mess, as users who aren't intimately familiar with the android permission system still regularly struggle to activate EN through microG. But we can leave that for a later point as this update here doesn't make it *much* worse than it already is.

N.B. Removing permissions that have initially been granted... just leads to madness it seems (as far as the app is concerned the permission is still granted, even after a phone restart... not even the security exceptions of the ble services trigger but something goes wrong deeper in the stack, so... I don't think there's anything we can do to help people who do that here...)


Bubu:
I added a commit that improves on the background location requesting as implemented by here: https://github.com/microg/GmsCore/pull/1520

Instead of directly jumping to the settings screen (which is what the "Request background location access" amounts to... :facepalm:) and leaving the user there without a clue what to do, we update the label, icon and button with new text explaining what the user needs to do next.

image

Yes, this flow is annoying... but I believe this improves the situation by quite a bit.

I liked @fynngodau's [toot](https://fosstodon.org/@utf8equalsX/104359649537615235) inside a comment as that's basically the best summary of the problem and solution here. :-)

Bubu:
There was a logic error in one of the permission checks, fixed now.

ale5000-git:
@Bubu
I just found this: https://codeberg.org/corona-contact-tracing-germany/GmsCore/commit/8f386dabdbd716c8f5ae0ac6603a3fa9bcc83fb6
It probably need to be merged also.

play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/AdvertiserService.kt
```diff
@@ -156,7 +160,11 @@ class AdvertiserService : LifecycleService() {
                         .setTxPowerLevel(ADVERTISE_TX_POWER_LOW)
                         .setConnectable(false)
                         .build()
-                advertiser.startAdvertising(settings, data, callback)
+                try {
+                    advertiser.startAdvertising(settings, data, callback)
+                } catch (e: SecurityException) {
+                    Log.e(TAG, "Couldn't start advertising: Need android.permission.BLUETOOTH_ADVERTISE permission.", )
```
ale5000-git, line 166: On SDK < 26 this permission doesn't exist, it just a log message but I think it is better change it and just report "Couldn't start advertising"

Bubu: Thanks, fixed!

play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/AdvertiserService.kt
```diff
@@ -204,7 +212,11 @@ class AdvertiserService : LifecycleService() {
...
         if (Build.VERSION.SDK_INT >= 26) {
             wantStartAdvertising = true
-            advertiser?.stopAdvertisingSet(setCallback as AdvertisingSetCallback)
+            try {
+                advertiser?.stopAdvertisingSet(setCallback as AdvertisingSetCallback)
+            } catch (e: SecurityException) {
+                Log.i(TAG, "Tried calling stopAdvertisingSet without android.permission.BLUETOOTH_ADVERTISE permission.", )
+            }
         } else {
             advertiser?.stopAdvertising(callback)
```
ale5000-git, line 221: I'm not sure since I don't really know that code but this case is the only one not try catched, is this right?

Bubu: I think there was a reason I didn't add this here but looking at it again I think it doesn't hurt to do the same as on the newer api here and just log a message :+1: .

Bubu: Added.

chilliger:
@mar-v-in I think this is critical. I was not aware that the app did not work until I look into it.

PR #1613, opened 2021-11-17
Re-add daily RPI histogram database query

fynngodau:
With the database call from https://github.com/microg/GmsCore/blob/6e176cceed88214d9c797e74af8e30e98fd0e87a/play-services-nearby-core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureDatabase.kt#L676-L685 readded, we can (more easily) add the old bar chart back as a card inside CCTG:

image


fynngodau:
This should still be modified somehow to allow for timezone support

fynngodau:
We probably won't need it this way and rather implement it in a timezone-aware way. And extension functions are cooler anyhow :slightly_smiling_face:

PR #1626, opened 2021-12-27
[ideia] Google maps and Android Auto

DavidFFerreira:
Since huawei has been ban, and problably other brands, and the vanced works perfecly in those phones, why not make an android auto and google maps for it? (for vanced)


ale5000-git:
Everything should already be fixed on the latest version (microG can now spoof another device).

PR #1634, opened 2022-01-10
Add missing fields to ENF database

### LINKED ISSUE #1633: ENF: reportType and daysSinceOnsetOfSymptoms get lost
When the two fields `reportType` and `daysSinceOnsetOfSymptoms` were introduced, no additional fields and/or logic was added to `ExposureDatabase`. Since the TemporaryExposureKeys are inserted into the database and then retrieved again when `findAllMeasuredExposures` is called, the two fields are thus lost and will not be returned to the application.

Elsensee:
At first I thought this might need a bigger solution but looking at it again, it really only needs changes in the Database class.

Fixes #1633.


mar-v-in:
Thanks a lot. Totally forgot about those two fields :)

ale5000-git:
@mar-v-in: Maybe I'm missing something but why it adjust table also from current DB_VERSION (11)?

Elsensee:
@ale5000-git `5 until 11` is excluding 11.

ale5000-git:
Here: https://github.com/Elsensee/GmsCore/commit/e84452aeaf5b8809aad06ae6a66bc3af478d72a3#diff-dc9624f7cabd388a7e4d57a44a0737b8e66d17ce2803ffb5b37f07499eb7e71e

it use:
```
if (oldVersion in 5 until 7) {
     Log.d(TAG, "Altering tables for version >= 7")

private const val DB_VERSION = 8
```

So if it is like you said (I cannot check now) then that one is wrong.
and also log entries are wrong:
```
>= 7
>= 11
```
or am I missing something?

Ping @mar-v-in

ale5000-git:
Some logs are < while others are >=, it is confusing.

Elsensee:
It is not. The version was changed to 8 without schema changes. There were only lines moved around for, probably, better readability or consistency. The changes were made earlier. The changes happened in this commit: https://github.com/microG/GmsCore/commit/4c2ef04364691b04aace378415b9a85c011829d3#diff-dc9624f7cabd388a7e4d57a44a0737b8e66d17ce2803ffb5b37f07499eb7e71e

Issue #1655, opened 2022-01-30
ENF: RSSI 0 dBm


transistortim:
Hey,

I'm using microG included in CCTG. On 2022-01-27 in the evening, I upgraded CCTG to 2.14.1.2, which according to the info screen uses microG version 18214816000.
As can be seen in the plot below, after the update there are many encounters with an RSSI of 0 dBm in exposure.db. In the days before the upgrade everything seems fine.

Looking at recent changes to the ENF code, I stumbled upon https://github.com/microg/GmsCore/commit/110157dab8097bc6c6589c950483731f4175320d and marked durations of 0 as `x` and durations unequal 0 as `o`. The plot indicates only encounters with a duration of 0 are stored with a valid RSSI.
![2022-01-30 exposuredb](https://user-images.githubusercontent.com/26510865/151720070-54378d20-fdce-4653-9237-67fe9293af98.png)

Device: Motorola Moto G7 Play (channel) with stock Android 10 (Google Play services are installed)

Can anybody reproduce this? Python code to create the plot:
```Python
import datetime

import matplotlib.pyplot as plt
import matplotlib.dates as md
from matplotlib.lines import Line2D
import sqlite3


file_name = "2022-01-30 exposure.db"

plt.figure()
plt.ylabel("RSSI / dBm")
ax=plt.gca()
xfmt = md.DateFormatter("%Y-%m-%d %H:%M:%S")
ax.xaxis.set_major_formatter(xfmt)
plt.grid()

with sqlite3.connect(file_name) as db_conn:
    data = db_conn.execute("SELECT timestamp, rssi, duration FROM advertisements ORDER BY timestamp;")
    for entry in data:
        x = datetime.datetime.fromtimestamp(int(entry[0]/1000))
        y = entry[1]
        if entry[2] == 0:
            plt.plot(x, y, marker="x")
        else:
            plt.plot(x, y, marker="o")

legend_elements = [Line2D([], [], linestyle="None", color="k", marker="x", label="duration == 0"),
                   Line2D([], [], linestyle="None", color="k", marker="o", label="duration $\\neq$ 0",)]
ax.legend(handles=legend_elements)
plt.show()

```

transistortim:
In my opinion the error is caused by the introduction of MIN in https://github.com/microg/GmsCore/commit/110157dab8097bc6c6589c950483731f4175320d.
We have `rssi = (a + r*b)/c`.
`c` is `MAX(duration, t-timestamp)`.
[...]

PR #1656, opened 2022-01-31
ENF: Fix average RSSI calculation

transistortim:
Correctly discard new measurements when the reported timestamp is older
than the timestamp in the database.
Fixes https://github.com/microg/GmsCore/issues/1655.


Elsensee:
Also, the first RSSI value is always discarded, because `MINIMUM_EXPOSURE_DURATION_MS` is 0 and there is `rssi * timestamp` (=0).
Maybe that should be set to 250? (Advertisements are sent every 250ms, so I would assume that this is the minimal value)

mar-v-in:
Indeed the first value is discarded, but that doesn't really matter. What matters is the time between the scan periods (3-5 minutes), where we do not have any RSSI data (as we are not scanning) so we are approximating it based on the first occurrence in the following scan period. This is similar to the behavior in GMS. (IIRC they approximate based on the previous scan period, which would mean they effectively discard the last measured rssi value.)

PR #1681, opened 2022-03-03
Use new method for lifting idle restrictions temporarily when receiving PUSH messages

grote:
The new method only gets used on Android 12 and above.

See: https: //cs.android.com/android/_/android/platform/frameworks/base/+/1aa89de1cfec5b7c1197d2f6c2b33dffda8bd3f5

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
+++ b/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
@@ -16,6 +16,7 @@
 
 package org.microg.gms.gcm;
 
+import android.annotation.SuppressLint;
 import android.app.AlarmManager;
 import android.app.PendingIntent;
 import android.app.Service;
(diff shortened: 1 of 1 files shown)


chirayudesai:
FTR, the issue was:

`W GmsGcmMcsSvc: java.lang.NoSuchMethodException: android.os.IDeviceIdleController$Stub$Proxy.addPowerSaveTempWhitelistApp [class java.lang.String, long, int, class java.lang.String]`

gustlik:
Hi, when do you plan to add it to new release? I am wondering if there is additional "beta" branch with that patch included. 

I have that issue with signal in android 12 on few devices and it is problematic. That is the reason why i am looking for solution. thank you.

ale5000-git:
@gustlik
You can wait the new release or try the nightly build at https://microg.org/dl/core-nightly.apk

bam2000:
Any updates when this will be implemented in the stable version ?

ale5000-git:
@bam2000
It is already in the version [0.2.28.231657](https://github.com/microg/GmsCore/releases/tag/v0.2.28.231657).

PR #1707, opened 2022-04-18
Fix EN with `targetSdk` 31

fynngodau:
This is required to build CCTG with Android 12: https://codeberg.org/corona-contact-tracing-germany/cwa-android/issues/176

### DIFF
--- a/play-services-nearby-core/src/main/AndroidManifest.xml
+++ b/play-services-nearby-core/src/main/AndroidManifest.xml
@@ -37,7 +37,8 @@
 
         <service
             android:name="org.microg.gms.nearby.exposurenotification.ExposureNotificationService"
-            android:process=":persistent">
+            android:process=":persistent"
+            android:exported="true">
             <intent-filter>
                 <action android:name="com.google.android.gms.nearby.exposurenotification.START" />
             </intent-filter>
@@ -45,7 +46,8 @@
 
         <receiver
             android:name="org.microg.gms.nearby.exposurenotification.ServiceTrigger"
-            android:process=":persistent">
+            android:process=":persistent"
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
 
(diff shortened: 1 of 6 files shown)


mar-v-in (approved):

fynngodau:
Additionally, it is now necessary to add `FLAG_IMMUTABLE` (or `FLAG_MUTABLE`) to all `PendingIntent`s. I added this to the PR.

Note that the flags of some PendingIntents were calculated using logical and, which I think resulted in 0 (`1<<30 & 1<<27`). I corrected this to use logical or.

There's one more change that I'm not sure how to upstream, since we cannot edit the manifest dynamically to either declare or not declare the location permission depending on the application's target SDK: https://codeberg.org/corona-contact-tracing-germany/GmsCore/commit/d0f0292718325965c19430433ca867f39539eb30 (This implements the [strong assertion](https://developer.android.com/guide/topics/connectivity/bluetooth/permissions#assert-never-for-location) that Bluetooth is never requested for location and therefore doesn't declare or request the permission on these API levels anymore – subject to testing on real devices.)

mar-v-in:
Regarding the location/scan permission: I'd propose to remove the related permissions entirely from nearby-core package and only have them at the package that embeds nearby-core (CCTG or microG core). Then it's up to the package to decide how they handle the permissions for their specific target sdk. We can populate the correct required permissions in `ExposureNotificationsConfirmActivity` (and potentially other place) based on target sdk and if `neverForLocation` flag is present at the `BLUETOOTH_SCAN` permission.

ale5000-git:
@fynngodau
FLAG_IMMUTABLE was added API level 23, so does the code works on Lollipop?

ale5000-git:
@mar-v-in
Since microG is a location provider probably need `android.permission.ACCESS_BACKGROUND_LOCATION` even without EN but I'm not sure.
Is it requested when compiled without EN?

fynngodau:
@ale5000 Since `FLAG_IMMUTABLE` is a constant, it is resolved to its constant value at compile time. Thus the access to this constant itself will not crash on Lollipop. I would strongly expect that the bit that represents the flag has no meaning in Lollipop and is therefore ignored, but I haven't specifically verified.

PR #1783, opened 2022-09-16
Fido: implement FidoAppIdExtension constructor

relan:
FidoAppIdExtension class should have a constructor[1] with the appId argument.

[1] https://developers.google.com/android/reference/com/google/android/gms/fido/fido2/api/common/FidoAppIdExtension#public-constructors


play-services-fido-api/src/main/java/com/google/android/gms/fido/fido2/api/common/FidoAppIdExtension.java
```diff
@@ -31,6 +31,10 @@ public String getAppId() {
         return appId;
     }
 
+    public FidoAppIdExtension(String appId) {
```
mar-v-in, line 34: The deserializer needs a no-parameter constructor (which can be private though). Please add an additional private no-parameter constructor.

mar-v-in (changes requested):

mar-v-in (approved):

PR #1801, opened 2022-10-06
Update com.android.tools.build:gradle to 7.2.2

ale5000-git:
This is only a minor update so it shouldn't cause any issue but it add support for SDK 33 (when it will be needed).


build.gradle
```diff
@@ -30,7 +30,7 @@ buildscript {
     ext.volleyVersion = '1.2.1'
     ext.wireVersion = '4.4.1'
 
-    ext.androidBuildGradleVersion = '7.2.1'
+    ext.androidBuildGradleVersion = '7.2.2'
```
elevenfive, line 33:
```suggestion
    ext.androidBuildGradleVersion = '7.3.1'
```

ale5000-git: @elevenfive
The version 7.2.2 works correctly without any other change but newer versions require other changes or it will fail build.

elevenfive: For what it's worth @ale5000-git I believe these build failures with AGP 7.3.1 due to more stringent lint are actually valid and important:

```
microg/GmsCore/play-services-base-core/src/main/kotlin/org/microg/gms/profile/ProfileManager.kt:70: Error: Implicit cast from XmlResourceParser to AutoCloseable requires API level 19 (current min is 14) [NewApi]
            else -> getProfileXml(context, profile)?.use {
```

ale5000-git: About the lint issue, it is possibly related to this: https://github.com/microg/GmsCore/issues/1816

elevenfive: I upgraded Gradle and was able to run `./gradlew assemble` (no lint in assemble of course), with this commit:

https://github.com/elevenfive/GmsCore/commit/7314e7c291e200a508fe6fdc396c98d97bdd5870

ale5000-git: Note: It is fine to update Gradle to see warnings/errors and fix things but currently I would advise against updating Gradle in production code since currently microG has more than 50 deprecation warnings, and things may break things in new versions of Gradle.

ale5000-git:
Supersed by: https://github.com/microg/GmsCore/commit/4c32a11c32bb1691ed4200611a9629c1d6041764

PR #1802, opened 2022-10-06
Fix IllegalArgumentException with JaCoCo + AGP >= 7.3.0

elevenfive:
Using GmsCore and JaCoCo with an instrumented test app built with Android Gradle Plugin version > 7.1.2 causes:

```
java.lang.Exception: java.lang.IllegalArgumentException
..at com.google.android.gms.dynamic.ObjectWrapper.unwrap(ObjectWrapper.java:43)
```

### DIFF
--- a/play-services-basement/src/main/java/com/google/android/gms/dynamic/ObjectWrapper.java
+++ b/play-services-basement/src/main/java/com/google/android/gms/dynamic/ObjectWrapper.java
@@ -34,15 +34,40 @@ public class ObjectWrapper<T> extends IObjectWrapper.Stub {
         if (obj == null) {
             return null;
         }
+
         if (obj instanceof ObjectWrapper) {
             return ((ObjectWrapper) obj).t;
         }
+
         IBinder binder = obj.asBinder();
         Field[] fields = binder.getClass().getDeclaredFields();
-        if (fields.length != 1) {
-            throw new IllegalArgumentException();
+
+        if (fields.length < 1) {
+            throw new IllegalArgumentException("No fields were found");
         }
-        Field field = fields[0];
+
+        // Ignore synthetic field(s) from JaCoCo or elsewhere
+        // https://www.jacoco.org/jacoco/trunk/doc/faq.html
+
+        @Nullable
+        Field field = null;
+
+        for (Field currentField : fields) {
+            if (currentField.isSynthetic()) {
+                continue;
+            }
+
+            if (field == null) {
+                field = currentField;
+            } else {
+                throw new IllegalArgumentException("Too many non-synthetic fields were found");
+            }
+        }
+
+        if (field == null) {
+            throw new IllegalArgumentException("No non-synthetic fields were found");
+        }
+
         if (!field.isAccessible()) {
             field.setAccessible(true);
             try {


ale5000-git:
@elevenfive
The function operate only on the first element, if it is going to allow more element it should be changed.

elevenfive:
I agree - but the idea here is to keep that behavior and simply not throw if there are more declared fields that can simply be ignored, so I believe this is safe.

ale5000-git:
We should wait a reply from @mar-v-in

mar-v-in:
I don't get why this is needed.

Does AGP > 7.1.3 add new fields to existing classes? This woild probably break tons of apps.
How do we know which field is the relevant one? The idea of this functionality is to allow wrapper objects that have exacts one field; this is when we know which field is the relevant one. 

Can you provide an example for an app affected by this? Maybe the root cause is something else that should be fixed differently.

elevenfive:
I will come up with a repro app.  How do we generate `com.google.android.gms-XXXXXXXX.apk`?  I see these APKs when running `./gradlew build` but I'm not sure which is the right one to use:
Screenshot from 2022-10-07 09-25-16

mar-v-in:
Official builds are with mapbox and with nearby.

elevenfive:
When trying to verify this fix and also see what the root cause is with the fields, I'm getting "[App] won't run without Google Play Services, which are not supported by your device."  Is there anything I need to do signing-wise?  I tried to see if the microG code is looking at the signature of the microg related APKs, but didn't see anything conclusive.  I'm trying with both local and release builds but still running into this.

mar-v-in:
What system are you running? Is microG self-check saying that everything is fine?

I still don't get what you try to do here. Is there any released app that is affected by this, which you can point me to? I haven't heard of any apps having this issue yet.

elevenfive:
Hi!  I figured it out and now have information about the root cause.  It is because we are using JaCoco and the latest R8 in AGP 7.3.0 changed some behaviors related to the build.  I have not searched extensively yet to find the exact cause but I can confirm that the difference occurs when only changing to AGP 7.3.1 from AGP 7.1.2.

I am working on an app that has containerized application support and we are using MicroG as a GMS substitute for apps when running inside our container.  We have an instrumented test with SupportMapFragment + MicroG that is failing when we try to upgrade to AGP 7.3.1 and I have narrowed it down to these JaCoCo synthetic fields.

I printed out the fields here:

```
7.1.2: private final java.lang.Object com.google.android.gms.dynamic.ObjectWrapper.zza]
7.3.1: private final java.lang.Object com.google.android.gms.dynamic.ObjectWrapper.zza, private static transient boolean[] com.google.android.gms.dynamic.ObjectWrapper.$jacocoData
```

From the JaCoCo FAQ:

```
My code uses reflection. Why does it fail when I execute it with JaCoCo?
To collect execution data JaCoCo instruments the classes under test which adds two members to the classes: A private static field $jacocoData and a private static method $jacocoInit(). Both members are marked as synthetic.

Please change your code to ignore synthetic members. This is a good practice anyways as also the Java compiler creates synthetic members in certain situation.
```

I will update this PR to more closely only allow the case of 2 fields when the 2nd is this synthetic JaCoCo field.

play-services-basement/src/main/java/com/google/android/gms/dynamic/ObjectWrapper.java
```diff
@@ -34,14 +34,24 @@ public static Object unwrap(IObjectWrapper obj) {
...
-        if (fields.length != 1) {
-            throw new IllegalArgumentException();
+
+        if (fields.length < 1) {
+            throw new IllegalArgumentException("No fields were found");
         }
+
+        // Ignore synthetic field from JaCoCo
+        // https://www.jacoco.org/jacoco/trunk/doc/faq.html
+        if (fields.length > 2 || (fields.length == 2 && !fields[1].isSynthetic())) {
```
mar-v-in, line 51: I'm not sure you can rely on this field being added as a second field.

elevenfive: Thank you for the review, that's a fair concern.  I've updated the handling to be order independent.

mar-v-in (approved):

play-services-basement/src/main/java/com/google/android/gms/dynamic/ObjectWrapper.java
ale5000-git: What about just count not-sysnthetic fileds?
So:
- Synthetic fields can be any number even 100
- Not-synthetic fieds must be 1

Quote:
> Please change your code to ignore synthetic members. This is a good practice anyways as also the Java compiler creates synthetic members in certain situation.

ale5000-git:
What about just count not-synthetic fileds?
So:
- Synthetic fields can be any number even 100
- Not-synthetic fieds must be 1

Quote:
> Please change your code to ignore synthetic members. This is a good practice anyways as also the Java compiler creates synthetic members in certain situation.

elevenfive:
> What about just counting not-synthetic fields? So:
> 
> * Synthetic fields can be any number even 100
> * Not-synthetic fields must be 1
> 
> Quote:
> 
> > Please change your code to ignore synthetic members. This is a good practice anyways as also the Java compiler creates synthetic members in certain situation.

Yes that's what I did with the latest changes. Require 1 non synthetic field and any number 0+ of synthetic fields

elevenfive:
@mar-v-in Can you share the release plans if they are known?

elevenfive:
Thank you for the review, approval, and merge!  

I tried to ballpark the release based on history but I'm having a hard time since it's pretty varied historically for this library.  Gut wise do you think the order of days / weeks / months?  A project I'm working on is blocked on updating AGP until this next release is out, so we're just curious for planning purposes.  Thanks again, no worries!

mar-v-in:
I'm aiming for this month, but probably not within the next two weeks.

Maybe using the [nightly](https://microg.org/dl/core-nightly.apk) is an option for you?

PR #1803, opened 2022-10-06
Increase build memory

elevenfive:
Fixes build issue with JVM OOM, when user's build environment is not already setting these values to an increased memory amount in `~/.gradle/gradle.properties`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 android.useAndroidX=true
 android.disableAutomaticComponentCreation=true
+org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx3072m -Xmx3072m -XX:+UseParallelGC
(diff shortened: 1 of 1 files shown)


elevenfive:
Copy pasta error, yep!  Will try locally with 2048.  Parallel GC is recommended these days, probably should keep that if no disagreement there.

ale5000-git:
Quote:
```
Currently Kotlin daemon uses the same JVM arguments as Gradle daemon. It is possible to override it in gradle.properties:
org.gradle.jvmargs=-Dkotlin.daemon.jvm.options=-Xmx1500m,-Xms=500m
but it is hard to discover.
```

So setting it with `org.gradle.jvmargs=-Xmx2048m` affect both Kotlin and Gradle (if I'm not wrong).

elevenfive:
Neat - thank you, didn't realize that!

mar-v-in (approved):

PR #1826, opened 2022-11-15
Add support for microG restricted signature spoofing

ale5000-git:
Since some ROMs only support the microG restricted signature spoofing (with a different permission) users are forced to use their custom microG version.
This change allow them to use the official microG if they want.

NOTE: The permission is already in use so we can't change the name

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -74,6 +74,7 @@
         android:protectionLevel="privileged|signature" />
 
     <uses-permission android:name="android.permission.FAKE_PACKAGE_SIGNATURE" />
+    <uses-permission android:name="android.permission.MICROG_SPOOF_SIGNATURE" />
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
--- a/play-services-core/src/main/java/org/microg/tools/selfcheck/RomSpoofSignatureChecks.java
+++ b/play-services-core/src/main/java/org/microg/tools/selfcheck/RomSpoofSignatureChecks.java
@@ -36,6 +36,7 @@ import static org.microg.tools.selfcheck.SelfCheckGroup.Result.Unknown;
 public class RomSpoofSignatureChecks implements SelfCheckGroup {
 
     public static final String FAKE_SIGNATURE_PERMISSION = "android.permission.FAKE_PACKAGE_SIGNATURE";
+    public static final String MICROG_FAKE_SIGNATURE_PERMISSION = "android.permission.MICROG_SPOOF_SIGNATURE";    // Note: it is a privileged permission (not runtime) so we cannot request it but only check
 
     @Override
     public String getGroupName(Context context) {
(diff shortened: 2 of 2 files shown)


mar-v-in:
If ROMs come with microG preinstalled, that would mean that no version of microG that is signed using a different key can be installed, no matter the permission name.

Also, the permission can be changed to privileged or otherwise restricted without renaming it, so if a ROM decides to rename it nonetheless, they introduced an incompatibility and I don't think it's on us to handle any such incompatibility introduced by ROMs.

ale5000-git:
@mar-v-in: I have verified that it is signature and privileged so installing microG in /system/priv-app works.
I think they renamed it to prevent confusion with the standard spoofing permission that is usually runtime.

With this change flashable zips can works.

Vavun:
@ale5000-git can you give examples please ?
What ROM devs made this mess ?

ale5000-git:
At the end it is just a matter of opinions and user choices, many people like to improve security in general and avoid possible unknown bugs that may arise in the future (also for non techie users).
As example there is CalyxOS.

mar-v-in:
CalyxOS has a signed system partition. One should not modify /system/priv-app.

ale5000-git:
If the bootloader isn't locked it shouldn't matter I think.
In addition I also plan to support also installing my zip as Magisk module (so the system partition isn't changed).

mar-v-in:
The CalyxOS installer locks the bootloader by default after installation.

ale5000-git:
I haven't tried but I think it can be unlocked again, also using flashable zip require a minimum of experience.

ale5000-git:
CalyxOS has a lot of improvements to security beside the bootloader lock and I don't consider the bootloader lock so important (but it is always user choice).
Also the are devices that may have older versions of CalyxOS but they are no longer supported by CalyxOS and in this way they would be able switch to the official microG.
Beside that there may be other ROMs that may choose to use the restricted permission (unlike the standard one).

The final purpose is to increase the compatibility of microG with the ROMs as much as possible.
My long plan is for my zip to support all ROMs and devices in the world.

mar-v-in:
- Most other ROMs that "restricted" the permission did so without renaming it.
- With unlocked bootloader, as soon as an untrusted person had physical access to the device, the device can't be trusted anymore (i.e. one should not enter the encryption password on such device).
- People with devices that are no longer supported by CalyxOS should probably use another ROM as they are missing out on important security updates like the one for CVE-2022-20465 (lockscreen bypass)
- CalyxOS additionally ships updates for microG independently through their F-Droid repository, so even if someone was stuck on an outdated version of CalyxOS for whatever reason, they could still install an updated version of microG via F-Droid.
- If you are modifying the system partition, you could also just modify the system to include support for the permission name that is already supported in microG.

ale5000-git:
Also look at issue 1829 where it would have been nice if he was able to try the latest nightly build of microG.

ale5000-git:
@chirayudesai

This is just my personal opinion and suggestion:
Support both permission names in the ROM, so then your microG apk doesn't need to be changed but if one want to stay with an unlocked boot loader and its own version or official version of microG can still do it.

About other apps using the same permission I don't think having permission automatically rejected (suggested change) or unsupported (as is now) make some difference.

chirayudesai:
We've just renamed it back, https://review.calyxos.org/q/topic:microg-perm-name

It'll be in CalyxOS 4.11.1 and newer.

Issue #1816, opened 2022-11-04
SafetyNetDatabase cannot be cast to java.lang.AutoCloseable


ale5000-git:
On Nougat when opening SafetyNet preferences I get: `SafetyNetDatabase cannot be cast to java.lang.AutoCloseable`

Problematic line: https://github.com/microg/GmsCore/blob/2a19ef45de5ad43bb3101f72aff477a2fc491e68/play-services-core/src/main/kotlin/org/microg/gms/ui/SafetyNetPreferencesFragment.kt#L106

Logcat:
```
01-01 05:20:05.777 12969 12969 E AndroidRuntime: FATAL EXCEPTION: main
01-01 05:20:05.777 12969 12969 E AndroidRuntime: Process: com.google.android.gms:ui, PID: 12969
01-01 05:20:05.777 12969 12969 E AndroidRuntime: java.lang.ClassCastException: org.microg.gms.safetynet.SafetyNetDatabase cannot be cast to java.lang.AutoCloseable
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at org.microg.gms.ui.SafetyNetPreferencesFragment$updateContent$1$1.invokeSuspend(SafetyNetPreferencesFragment.kt:106)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
01-01 05:20:05.777 12969 12969 E AndroidRuntime: 	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
[...]

PR #1830, opened 2022-11-18
Explicitly implement AutoCloseable in SafetyNetDatabase

StarGate01:
This PR fixes a hard crash on Android 9 when opening the SafetyNet Settings.

This is required for API < 29 , as `SQLiteOpenHelper` implements `AutoCloseable` only starting with Android Q.

See: https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper Note 2

### DIFF
--- a/play-services-safetynet-core/src/main/kotlin/org/microg/gms/safetynet/SafetyNetDatabase.kt
+++ b/play-services-safetynet-core/src/main/kotlin/org/microg/gms/safetynet/SafetyNetDatabase.kt
@@ -11,9 +11,11 @@ import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteOpenHelper
 import android.os.Build
 import android.util.Log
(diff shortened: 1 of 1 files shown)


StarGate01:
It appears the CI build task does not honor the `@RequiresApi(19)` annotation, any ideas on how to fix this?

ale5000-git:
Hi, look at my report: https://github.com/microg/GmsCore/issues/1816

I think it can be fixed simplier by using the androidx library, see here: https://stackoverflow.com/questions/73502338/android-content-res-typedarray-cannot-be-cast-to-java-lang-autocloseable/73502339#73502339

StarGate01:
Oh, thats an elegant solution, feel free to close this PR once that solution is merged.

ale5000-git:
Fixed by mar-v-in with: https://github.com/microg/GmsCore/commit/ee997ba3ac175f9d7cf04243aaf534c9a9182a69

PR #1848, opened 2022-12-13
Updates to Simplified Chinese translations

hustler-not-chatty:
- Adjust multiple translations for term "exposure".
- Several new strings translated.
- Other minor tweaks.


mar-v-in:
Thanks.

PR #1849, opened 2022-12-13
Specify targetPackage for intents from Authenticator Preferences

hustler-not-chatty:
Turns out that those intents may resolve to activities from Vanced microG, resulting a UID mismatch between resolved app (Vanced) and authenticator app (microG).

Validated by uninstalling Vanced microG, after which the AccountSettingsActivity launched as expected.

This is an attempt to eliminate the conflict.

Logcat from Settings app:
```
AccountTypePrefLoader: Refusing to launch authenticator intent becauseit exploits Settings permissions: Intent { act=com.google.android.gms.accountsettings.PRIVACY_SETTINGS flg=0x10000000 (has extras) }
```

Reference: https://cs.android.com/android/platform/superproject/+/master:packages/apps/Settings/src/com/android/settings/accounts/AccountTypePreferenceLoader.java;l=166;drc=6d97d6d7215fef247d1a90e05545cac3676f9212?q=Refusing%20to%20launch%20authenticator%20intent&ss=android

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -543,7 +543,6 @@
 
         <activity
             android:name="org.microg.gms.ui.AccountSettingsActivity"
-            android:exported="true"
             android:process=":ui"
             android:taskAffinity="org.microg.gms.settings">
             <intent-filter>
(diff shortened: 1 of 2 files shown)


mar-v-in:
Thanks. Good catch.

PR #1860, opened 2023-01-01
Make more SafetyNet-related strings translatable

Fs00:
I was about to create a PR to update Italian translations, when I noticed that a couple of SafetyNet-related UI strings weren't used anywhere and some more weren't translatable.
This PR addresses those small issues and does a bit of code clean-up (most notably, I got rid of a class that wasn't used).


mar-v-in:
Thanks.

PR #1863, opened 2023-01-06
Update Italian translations

Fs00:
This is a follow-up to #1860 to update the Italian translation and (again) slightly improve localization for the SafetyNet UI.
I've also translated the UI strings for the oss-licenses module ~~(which had a couple of unused strings btw)~~.


mar-v-in:
The "unused strings" from oss-licenses may be used by apps using it as a library, so they have to stick around for compatibility. Feel free to add a corresponding comment instead, so it is documented why they are there.

Fs00:
Problem addressed (I've also added translations for those strings), I think this is now ready.

mar-v-in:
Thanks!

Issue #1463, opened 2021-04-29
Any plans to move from MapBox to its FOSS fork MapLibre?


token0:
MapLibre project is an open source alternative to Mapbox’s proprietary 2.0 update.
Is there any plans to move?

https://github.com/MapLibre
https://twitter.com/maplibre

mar-v-in:
Probably, yes. For now we can stick with the older version we already have.

huuhaa:
@mar-v-in Now when time has flies after latest comment, are there any news about this?

Atemu:
I don't see a reason to switch to MapLibre. They haven't added any interesting features for this use-case. Their major changes were for iOS, QT and other platforms.

For our use-case, it's about the same as the old MapBox version currently in use.

ildar:
I see @token0 made a Pull Request (#1865) for this. Is it regression-free?
@mar-v-in could you please make an alpha release with that so we can test
it beforehands? Thanks
Thank you all guys for this!

PR #1865, opened 2023-01-09
Use MapLibre

fynngodau:
Fixes https://github.com/microg/GmsCore/issues/1463.


Torvin:
Hey question about this. I installed `0.2.27.223616` and according to the release notes this change is included, but when I open an app that shows a map (e.g. Uber) - I still see "mapbox" in the bottom left corner. Is this the expected behaviour or did something go wrong on my side?

mar-v-in:
That's fine. The latest release version of MapLibre we're using still has the Mapbox Logo. This is going to change with their next release.

Issue #1906, opened 2023-03-21
[com.trafi.whitelabel.bvg] crashes on show closest bike


thrillfall:
**Affected app**
Name: Jelbi
Package id: com.trafi.whitelabel.bvg

**Describe the bug**
When i click the button the show me bicycles or scooters close to my location it opens a drawer displaying the closest vehicles data then the app crashes

**To Reproduce**
1. Go to 'Main Menu'
2. Click on 'Bikes' (bicycle symbol)
3. See bicycle detail drawer
4. crash

**Expected behavior**
It should not crash on bike detail display

**Screenshots**
the app crashes shortly after this is displayed
![jelbi_crash](https://user-images.githubusercontent.com/15801468/226586283-803ee2ca-416a-43c4-aba4-a5f01a8f38b5.png)

**System**
Android Version: 12
Custom ROM:  LineageOS 19 (microG build)

**microG**
microG Core version: 0.2.27.223616
microG Self-Check results: All ticked

**Additional context**
Logs: 
```03-21 11:30:29.112 32676   497 W .whitelabel.bv: Accessing hidden method Lcom/android/org/conscrypt/ConscryptEngineSocket;->setUseSessionTickets(Z)V (max-target-q,core-platform-api, reflection, denied)
03-21 11:30:29.113 32676   497 W .whitelabel.bv: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setUseSessionTickets(Z)V (max-target-q,core-platform-api, reflection, denied)
03-21 11:30:29.113 32676   497 W .whitelabel.bv: Accessing hidden method Lcom/android/org/conscrypt/AbstractConscryptSocket;->setUseSessionTickets(Z)V (max-target-q, reflection, denied)
03-21 11:30:29.113 32676  1153 W .whitelabel.bv: Accessing hidden method Lcom/android/org/conscrypt/ConscryptEngineSocket;->setUseSessionTickets(Z)V (max-target-q,core-platform-api, reflection, denied)
03-21 11:30:29.113 32676  1153 W .whitelabel.bv: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setUseSessionTickets(Z)V (max-target-q,core-platform-api, reflection, denied)
[...]

Issue #1886, opened 2023-02-21
[se.skanetrafiken.washington] Crashes on detailed connection page, if it includes a transfer


felix-engelmann:
**Affected app**
Name: Skånetrafiken
Package id: se.skanetrafiken.washington

**Describe the bug**
The app crashes when trying to display a connection connection which includes a transfer, i.e. most likely a footpath.

**To Reproduce**
1. Go to 'Search journey' in the bottom menu
2. Search for a connection including a change, e.g. From: Lund LTH, To: Malmö C 
3. Click on any connection with two legs
4. App shows partial details and then crashes.

**Expected behavior**
The expected behavior is to show a detailed view of the connection.

**Screenshots**
Search overview working as expected
![vlcsnap-2023-02-21-13h47m11s284](https://user-images.githubusercontent.com/5196262/220348779-f8d95f5f-8a57-4296-a15a-a1b2b07bc547.png)
Loading screen of the detailed connection which tries to load a small map overview but then immediately crashes
![vlcsnap-2023-02-21-13h47m26s784](https://user-images.githubusercontent.com/5196262/220348771-7579e88f-6b36-4e47-8153-2c571ab4e498.png)

**System**
Android Version: 12
Custom ROM: /e/ OS 1.8.1-s-2023020357059-stable-FP4

**microG**
microG Core version: 0.2.27.223616-1 (4ec0d9f)-neon
microG Self-Check results: All ticked

**Additional context**
logcat shows the following output during the crash (excerpts):
```
02-19 12:32:54.794  3499  3516 D SafeParcel: Unknown field id 26 in com.google.android.gms.measurement.internal.AppMetadata, skipping.
02-19 12:32:54.794  3499  3516 D SafeParcel: Unknown field id 27 in com.google.android.gms.measurement.internal.AppMetadata, skipping.
02-19 12:32:54.794  3499  3516 D MeasurementService: f1(com.google.android.gms.measurement.internal.EventParcel@73edf52) for AppMetadata[se.skanetrafiken.washington]
02-19 12:32:54.794  3499 24962 D MeasurementService: f10(-6650651463208247115, Journey details, MainActivity, se.skanetrafiken.washington)
02-19 12:32:54.826 16413 27348 I Mbgl-EGLConfigChooser: In emulator: false
02-19 12:32:54.831 16413 16413 V Mbgl-ConnectivityReceiver: connected - true
02-19 12:32:54.836 16413 16413 D GmsMapBitmap: registerMap
[...]

PR #1920, opened 2023-04-22
Fix calling parameters of Dot for PatternItem

felix-engelmann:
This is a minor change fixing #1886. The Skanetrafiken app now is able to display footpaths.
Thanks to @raminou who found the issue that the length argument must be of type Float.

Unfortunately I was not able to test the other apps affected by this issue (cf #1906)

Screenshot of the working view:

working

Zoomed in section of the footpath:

footpath

### DIFF
--- a/play-services-maps/src/main/java/com/google/android/gms/maps/model/Dot.java
+++ b/play-services-maps/src/main/java/com/google/android/gms/maps/model/Dot.java
@@ -21,7 +21,7 @@ public final class Dot extends PatternItem {
      * Constructs a {@code Dot}.
      */
     public Dot() {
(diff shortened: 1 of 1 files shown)


thrillfall:
Thanks for providing a fix

mar-v-in:
The `length` field in `PatternItem` is a `Float` (aka nullable) by now (e.g. in 0.3.0). Can you confirm you still see issues without this change?

felix-engelmann:
The crash no longer happens, from which I assume it is fixed. The only problem remaining, is that the map is no longer zoomable/panable. I'll investigate this once I have more time.

PR #1922, opened 2023-04-26
Request not a pull

Mohammedibrahim15:
I know that you and oxrxl launched new Mmicrog . I don't know how to communicate with oxrxl but i want to tell you if you could add Huawei login feature in this microg like the feature present in vanced microg .. please can you and ixrxl put this into your considerations?


ale5000-git:
Do you request really have sense?

So why not add a button for every brand just to clutter the screen and use a full HD TV just to show all buttons?

A single button is enough for all devices, to avoid the Google ban you can just use "Device profiles".

```
Device profiles
This features is mostly useful for Huawei devices or other devices that are banned from accessing Google services. Choosing a profile of a non-banned device should allow using the device with Google services.

For now, two device profiles are included with microG (Nexus 5X with Android 8.1.0 and Moto G with LineageOS 14.1).

If you are a developer, you can easily create a device profiles based on data gathered from a device. See [this file](https://github.com/microg/GmsCore/blob/master/play-services-core/src/main/res/xml/profile_bullhead_27.xml) for a sample device profile. Once created, you can import device profiles in microG settings (under "Device registration") to test the profile.
```

PR #1948, opened 2023-06-01
Catch some exceptions caused by network errors

Uldiniad:


mar-v-in:
I seem to have done the same in 47720085820a03988e99b605a387a32f81f89542

PR #1949, opened 2023-06-01
Update gradle and dependencies

Uldiniad:


Uldiniad:
> You're removing the embedding of safe-parcel into play-services-basement. This is required to ensure the same module structure as used by original Google modules. If there is no way to implement embedding ("fat aar") with AGP 8, you'd need to copy the whole safe-parcelable library over to play-services-basement (which would also allow to provide the same API for safe parcel as provided by Google's play-services-basement)

Thanks for the info. Will attempt to addres this

Uldiniad:
> > You're removing the embedding of safe-parcel into play-services-basement. This is required to ensure the same module structure as used by original Google modules. If there is no way to implement embedding ("fat aar") with AGP 8, you'd need to copy the whole safe-parcelable library over to play-services-basement (which would also allow to provide the same API for safe parcel as provided by Google's play-services-basement)
> 
> Thanks for the info. Will attempt to addres this

Done

mar-v-in (changes requested):
I see you often adjusted the version of androidx library dependencies in non-core packages. These typically reflect the dependency required from the respective original Google library, as is typically written in a comment above those dependencies. As these are api scope dependencies and thus part of the public API of those packages, they need to match to ensure full compatibility.

You can easily check the version of those dependencies that you need to use via https://mvnrepository.com/ (e.g. for play-services-auth-base, the version of andoidx.collection must be 1.0.0 as can be seen on https://mvnrepository.com/artifact/com.google.android.gms/play-services-auth-base/18.0.9). If you want to make sure a later version of the library is effectively bundled in GmsCore, you can reference that library from the corresponding core package.

Uldiniad:
Superseded by https://github.com/microg/GmsCore/commit/4c32a11c32bb1691ed4200611a9629c1d6041764

PR #1950, opened 2023-06-01
Bump target SDK to 33

Uldiniad:


Uldiniad:
I made everything exported by default because I couldn't easily tell which components need to be and which don't. Please edit as necessary

mar-v-in:
Before 33 we had exported implicitly when there was an intent filter. As such, exported should only be true on entries that have an intent filter or where the name is in com.google namespace (those may be called directly by apps)

Uldiniad:
> Before 33 we had exported implicitly when there was an intent filter. As such, exported should only be true on entries that have an intent filter or where the name is in com.google namespace (those may be called directly by apps)

Thank you for the info. Will adress this when I get some time

Uldiniad:
On a similar vein, what's the process for choosing PendingIntent mutability flags?

Uldiniad:
> > Before 33 we had exported implicitly when there was an intent filter. As such, exported should only be true on entries that have an intent filter or where the name is in com.google namespace (those may be called directly by apps)
> 
> Thank you for the info. Will adress this when I get some time

Verified that all entries where `android:exported` was set to `true` have intent filters

Uldiniad:
> On a similar vein, what's the process for choosing PendingIntent mutability flags?

Added `FLAG_IMMUTABLE` by default except where it caused errors/crashes

grote:
Looks like this MR is not taking into account changes to the ability to schedule exact alarms: https://developer.android.com/reference/android/app/AlarmManager#setExactAndAllowWhileIdle(int,%20long,%20android.app.PendingIntent)

> Note: Starting with Build.VERSION_CODES#S, apps targeting SDK level 31 or higher need to request the SCHEDULE_EXACT_ALARM permission to use this API, unless the app is exempt from battery restrictions.

chirayudesai:
Based on discussions, it'd be better to do the target SDK bump one at at time, and test each bump as extensively as possible.

It'd also make it easier to revert / fix things as they come up.

It's 29 now, so 30 next, then 32 (as it's pretty much the same as 31), 33, and finally 34.

PR #1977, opened 2023-07-07
Update mapbox libraries

### LINKED ISSUE #969: Poor map performance with many markers/layers
The map view in the [WeShare](https://play.google.com/store/apps/details?id=com.we.weshare) app is unusably slow to respond. It seems to do lot's of stuff in the UI thread.
On a phone with play services the app works fine.

So this is most likely some of the overlay things that are slow here. The map works fine until the car POIs are loaded then it stops responding to input after scroll/zoom for a couple of seconds.

image

You can also watch the progress of the poi clustering algorithm... it takes a couple of seconds when there's a lot going on on the map :see_no_evil:

VxlerieUwU:
This fixes #969 but requires setting the minimum Android version to 5.0. It may need more testing since breaking changes may be present in org.maplibre.gl:android-sdk.

### DIFF
--- a/play-services-maps-core-mapbox/build.gradle
+++ b/play-services-maps-core-mapbox/build.gradle
@@ -21,8 +21,8 @@ dependencies {
     implementation project(':play-services-maps')
     implementation project(':play-services-base-core')
     implementation project(':play-services-location')
-    implementation("org.maplibre.gl:android-sdk:10.0.0")
-    implementation("org.maplibre.gl:android-plugin-annotation-v9:1.0.0") {
+    implementation("org.maplibre.gl:android-sdk:10.2.0")
+    implementation("org.maplibre.gl:android-plugin-annotation-v9:2.0.0") {
         exclude group: 'com.google.android.gms'
     }
     implementation 'org.maplibre.gl:android-sdk-turf:5.9.0'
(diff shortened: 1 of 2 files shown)


ale5000-git:
> but requires setting the minimum Android version to 6.0

Or `tools:overrideLibrary` can be used to ovverride the minimum SDK value.

fynngodau:
Indeed, it's great that maplibre finally has a new release for the annotations plugin :)

Current microG contains version 10.0.0 of `maplibre-native`, which already requires API 21. We therefore instruct the manifest merger to ignore this here: https://github.com/microg/GmsCore/blob/master/play-services-maps-core-mapbox/src/main/AndroidManifest.xml#L32

Since version 2.0.0 of `maplibre-plugins` aligns its minSdk with `maplibre-native`, we should add `com.mapbox.mapboxsdk.plugins.annotation` in the same place. A small note, API 21 corresponds to Android 5, not 6.

> It may need more testing since breaking changes may be present in org.maplibre.gl:android-sdk.

Fortunately this is not the case: [10.2.0](https://github.com/maplibre/maplibre-native/releases/tag/android-v10.2.0) is documented to be identical to [10.0.2](https://github.com/maplibre/maplibre-native/releases/tag/android-v10.0.2), which contains barely any changes, just like [10.0.1](https://github.com/maplibre/maplibre-native/releases/tag/android-v10.0.1).

mar-v-in:
@VxlerieUwU I think you wanted to do

```xml
<uses-sdk tools:overrideLibrary="com.mapbox.mapboxsdk,com.mapbox.mapboxsdk.plugins.annotation" />
```

VxlerieUwU:
> @VxlerieUwU I think you wanted to do
> 
> ```
> <uses-sdk tools:overrideLibrary="com.mapbox.mapboxsdk,com.mapbox.mapboxsdk.plugins.annotation" />
> ```

Thanks!

PR #1998, opened 2023-07-31
Provisioning: Location providers

t-m-w:
New provisioning extras to control Location features:
* Internet lookups (Mozilla Location Services): wifi_mls, cell_mls
* Local location learning: wifi_learning, cell_learning
* Query location from supported hotspots: wifi_moving
* Geocoder (Nominatim): nominatim_enabled

### DIFF
--- a/play-services-core/build.gradle
+++ b/play-services-core/build.gradle
@@ -33,6 +33,7 @@ dependencies {
     implementation project(':play-services-fido-core')
     implementation project(':play-services-gmscompliance-core')
     implementation project(':play-services-location-core')
+    implementation project(':play-services-location-core-base')
     implementation project(':play-services-location-core-provider')
     withNearbyImplementation project(':play-services-nearby-core')
     implementation project(':play-services-oss-licenses-core')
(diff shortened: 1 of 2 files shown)


mar-v-in:
This looks reasonable. Note though that (starting next version) all providers will be disabled by default (except when the corresponding UnifiedNlp module was enabled before the upgrade - having them enabled by default was actually by accident), so this will be more useful as an opt-in ;)

t-m-w:
> This looks reasonable. Note though that (starting next version) all providers will be disabled by default (except when the corresponding UnifiedNlp module was enabled before the upgrade - having them enabled by default was actually by accident), so this will be more useful as an opt-in ;)

Good to know, and thanks for checking this out so quickly. I was asked to make it more granular, so I just did that now.

t-m-w:
I see again what you mean about location settings being off by default, including location learning! So I added a few more settings to this, since I think users will commonly want location learning enabled.

mar-v-in:
In case anyone wonders why location learning is disabled by default: it effectively creates a local database of locations where the smartphone used GPS over the last weeks. For some people that may be problematic, so that's why it makes sense to be opt-in, so they're not surprised by that.

Issue #265, opened 2017-01-09
Store native libs uncompressed and zipaligned


ale5000-git:
If the native libs are stored in the apk uncompressed and zipaligned then they can be loaded from the apk without be pre-extracted.

It won't work on all cases but it would simplify things in most cases.

Ref: https://github.com/LineageOS/android_frameworks_base/commit/ff193d642eea7128faad837d19e347cd25212c27

mar-v-in:
I guess this is about the case when you put microG to `/system` right?

ale5000-git:
I think it apply to both system and data, it use the libraries from the apk without extract them.

Look [here](https://medium.com/@wkalicinski/smallerapk-part-8-native-libraries-open-from-apk-fc22713861ff#.d4o7omnhd).

ale5000-git:
Thanks :)

PR #2000, opened 2023-07-31
Allow to compress the libraries

ale5000-git:
The current size of microG GmsCore is 76,3 MB (64,8 MB are just uncompressed libraries).
If the libraries are compressed then the total size of microG GmsCore become about 35,8 MB.

This PR change the values of useLegacyPackaging (previously know as extractNativeLibs) to allow compressing the libraries.

I was the one (on #265) suggesting to store the libraries uncompressed and it was nice at the time that the libraries was small but now they are quite big and I think now is more beneficial to have them compressed.


CoelacanthusHex:
But this will increase the size after installation to library size + compressed library size. Because the library must be extracted from the apk before it can be loaded.

mar-v-in:
> But this will increase the size after installation to library size + compressed library size. Because the library must be extracted from the apk before it can be loaded.

Yes, but it might still be less total space used (would have to investigate) as not all libraries are extracted, but only those required for the architecture. Also download size is relevant as well, not only disk space use.

ale5000-git:
@CoelacanthusHex
As mar-v-in said, the total will be less on most devices since there are libraries for 4 architectures in microG GmsCore and in general just 1 architecture or maximum 2 are extracted.

ale5000-git:
I have tested it by compressing with a zip compressor and the result is this:
```
Original file = 76,3 MB
File with compressed libs = 35,8 MB

File with compressed libs (35,8 MB) + external arm64&arm32 libs (28,6 MB) = 64,4 MB
File with compressed libs (35,8 MB) + external arm64 libs (17,2 MB) = 53 MB
File with compressed libs (35,8 MB) + external arm32 libs (11,4 MB) = 47,2 MB

File with compressed libs (35,8 MB) + external x86_64&x86 libs (36,1 MB) = 71,9 MB
File with compressed libs (35,8 MB) + external x86_64 libs (18,1 MB) = 53,9 MB
File with compressed libs (35,8 MB) + external x86 libs (17,9 MB) = 53,7 MB
```

It should cover most cases, obviously there may be edge cases not covered here but it is mostly smaller.

PR #2019, opened 2023-08-28
safeparcel: Support short

chirayudesai:
* Fixes crash with certain USB security keys, exception:

```
Caused by: java.lang.RuntimeException: Type is not yet usable with SafeParcelReflectionUtil: short at org.microg.safeparcel.SafeParcelReflectionUtil$ClassDescriptor$FieldDescriptor.getType(SafeParcelReflectionUtil.java:521)
```

Full stacktrace: https://zifb.in/raw/Q89aPk61T4
Tested using: https://demo.yubico.com/webauthn-technical/registration


play-services-basement/src/main/java/org/microg/safeparcel/SafeParcelReflectionUtil.java
```diff
@@ -364,6 +367,14 @@ private static void readField(AutoSafeParcelable object, Parcel parcel, int head
                 descriptor.field.set(object, l);
                 break;
             }
+            case Short: {
+                short i = (short) SafeParcelReader.readInt(parcel, header);
```
mar-v-in, line 371: This works because short and int are identical in SafeParcel's byte representation, but please use `SafeParcelReader.readShort()` nonetheless.

mar-v-in (changes requested):

play-services-basement/src/main/java/org/microg/safeparcel/SafeParcelReflectionUtil.java
chirayudesai: Done.

mar-v-in: No need to cast to `short` anymore, as `readShort()` already returns a `short` :wink:

chirayudesai: Sorry about that, didn't pay enough attention the first time around.

mar-v-in (approved):

PR #2035, opened 2023-09-14
Declare com.google.android.providers.gsf.permission.READ_GSERVICES permission, fix #1366

### LINKED ISSUE #1366: READ_GSERVICES permission
I am using microG on android 9. Trying to install Google VR today and com.google.vr.apps.ornament refuses to run without permission com.google.
android.providers.gsf.permission.READ_GSERVICES

Should gsf permissions be defined by in microg?

ale5000-git:
Declare com.google.android.providers.gsf.permission.READ_GSERVICES permission, fix #1366

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -17,6 +17,10 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <permission
+        android:name="com.google.android.providers.gsf.permission.READ_GSERVICES"
+        android:label="@string/perm_gsf_read_gservices_label"
+        android:protectionLevel="normal" />
+    <permission
         android:name="com.google.android.c2dm.permission.RECEIVE"
         android:label="@string/perm_c2dm_receive_label"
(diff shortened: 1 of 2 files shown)


play-services-core/src/main/AndroidManifest.xml
```diff
@@ -16,6 +16,11 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <permission
+        android:name="com.google.android.providers.gsf.permission.READ_GSERVICES"
+        android:label="@string/perm_gsf_read_gservices_label"
+        android:permissionGroup="android.permission-group.ACCOUNTS"
```
mar-v-in, line 22: This permission group doesn't seem to exist (anymore). Maybe just not put the permission in a group - original GSF also doesn't.

mar-v-in (changes requested):
OK, let's get this into GmsCore, some people don't install GsfProxy and we already have other stuff that originally was in GSF in GMS.

mar-v-in (approved):

PR #2042, opened 2023-09-17
Improve microG settings activities

ale5000-git:
@mar-v-in

Changes:
- Separates launcher icon from main settings activity, so it can be disabled without making microG settings inaccessible since there are launchers that doesn't allow to hide Apps on their own (takes inspiration from the PR of @Nanolx).
- Add an activity-alias for Google settings compatibility (other apps are supposed to able to open Google settings).


mar-v-in:
The issue I see here, is that people that currently hide microG `SettingsActivity` from their launcher will now have it shown again (as the launcher activity is changed to `SettingsActivityLauncher`, the hiding is not preserved).

Also, can you point to apps that currently launch directly into `com.google.android.gms.app.settings.GoogleSettingsLink`?

ale5000-git:
Maybe users that currently hide microG SettingsActivity from their launcher will now have it shown again but this is a really minor issue; once this is changed it will avoid people breking microG settings only to hide them from launcher.

This app: https://play.google.com/store/apps/details?id=com.weberdo.apps.serviceinfo&hl=en
Inside "More..." there is an option to open Google settings but there may also be other apps.

mar-v-in:
> Maybe users that currently hide microG SettingsActivity from their launcher will now have it shown again but this is a really minor issue; once this is changed it will avoid people breaking microG settings only to hide them from launcher.

I think we can probably just switch the names around a little bit, such that we rename current `SettingsActivity` to e.g. `MainSettingsActivity` and `SettingsActivity` becomes an activity-alias with category launcher (what in your current PR is `SettingsActivityLauncher`). This way,. any hiding of SettingsActivity is preserved.

The reason why I think this is important is that there are probably way more people that have `SettingsActivity` correctly hidden in their launcher (including custom ROMs and launchers that hide it by default) than people that have `SettingsActivity` disabled.

ale5000-git:
@mar-v-in
Done; but having "Launcher" in the name would have made it clearer to those who want to disable it manually.

mar-v-in:
Merged manually.

Issue #2059, opened 2023-09-23
MicroG settings icon in settings is always light mode and it's out of place. 


Wojtaz0w:
**Describe the bug**
The icon of microG in device settings is almost invisible when dark theme is on. 

**To Reproduce**
1. Go to settings
2. (Don't) see the broken icon. 

**Expected behavior**
Icon respects device theme. 

**Screenshots**
![Screenshot_20230923-203435_Ustawienia.png](https://github.com/microg/GmsCore/assets/104674625/6ea67d18-094c-4bd0-a8e8-c5296e06e65a)

![Screenshot_20230923-203423_Ustawienia.png](https://github.com/microg/GmsCore/assets/104674625/57d5c95c-9e0a-4f59-bf20-5a8fa0e11cb1)

**System**
Android Version: 13 (TQ3C.230805.001.A4) 
Custom ROM: risingOS 1.3

Turbolqk:
Imo it the setting it self should be moved to the general settings menu, where google play services normally stay.

ale5000-git:
This issue was already fixed some time ago but I don't remember the exact version so I'm closing this.
In case you still have the issue with the latest version, please comment here.

PR #2061, opened 2023-09-25
Fix the icon of microG settings inside Android settings on Android < 8, fix #2057

### LINKED ISSUE #2057: microG settings icon inside Android settings is missing on Nougat
The icon of `microG Settings embedded in System Settings on SDK 23` is missing on Nougat.

Screenshot_20230922-180320

---

In the past it was like this:
microg-settings

ale5000-git:
Fix the icon of microG settings inside Android settings on Android < 8, fix #2057

This is probably something left out from this commit: https://github.com/microg/GmsCore/commit/90f8a6a50243d9fd84d1b11ee9c54e830c4e227f


Wojtaz0w:
does it fix issue created by me? #2059

ale5000-git:
Obviously no.
In the title there is "Android < 8" that mean "Android version less than 8", in this case in particular apply to:
```
Android 7.x
Android 6.x
```

Wojtaz0w:
Nevermind, i didnt read that part of title... sorry

ale5000-git:
Don't worry, on newer Android it use a different file so they are completely separate.

ale5000-git:
@mar-v-in
I have also tried to reuse the drawable from here on Nougat: https://github.com/microg/GmsCore/blob/master/play-services-core/src/main/res/drawable-v26/ic_app_settings_foreground_mono.xml

It generally works but it is too small and if I try to resize it it get disaligned and truncated.
Since the benefit of vectorial images is to render correctly at any size I conclude that the xml is broken.

mar-v-in:
I added the missing icon via 44b9173476c4f293649507703e9179d24667754d

Not sure what issue you had with the vector drawable, but resizing it worked fine for me.

ale5000-git:
@mar-v-in

With your changes it works.

Is this file still needed? https://github.com/microg/GmsCore/blob/master/play-services-core/src/main/res/drawable-anydpi-v21/microg_light_color_24.xml

PR #2068, opened 2023-10-01
Add a missing null check in GamesService

js6pak:
Fixes a warning I noticed when launching Clash Royale. It worked regardless, but it might help in other games.


mar-v-in (approved):

Wojtaz0w:
Thank you so much.

PR #2071, opened 2023-10-11
Expose device profile as content provider

fynngodau:
As a next step, fakestore can access the device profile when it queries the licensing service. For this, it can import `play-services-base-core` and use `ProfileManager.applyProfileData` with the data provided by the provider.


mar-v-in:
- I don't think this provider should be in `play-services-base/core`: there is no need to expose this for other apps that use `play-services-base/core` (apps that bundle any of the `/core` modules as dependency will also bundle `play-services-base/core`, think of use cases like CCTG). Maybe put it in `play-services-core` or create a new module `play-services-base/core/package` (we already have that structure for nearby)
- Can you add a new permission (e.g. `org.microg.gms.permission.INTERNAL`) with protection level signature that is required to access the provider? It's not really critical data to protect in this case, but we likely will want this in the future anyway for cases when the vending app accesses info from GmsCore, so better start early

fynngodau:
> Maybe put it in `play-services-core` or create a new module `play-services-base/core/package` (we already have that structure for nearby)

Thanks for the feedback, I'll adapt my PR later today.

> Can you add a new permission (e.g. `org.microg.gms.permission.INTERNAL`) with protection level signature that is required to access the provider?

I considered that the device profile information could be used by other apps as well, like Aurora, which also provides a device spoof feature. For this reason I think that we should only add the permission once we add additional providers with more sensitive data.

fynngodau:
@mar-v-in I've moved the provider to a new `play-services-core-package` module.

mar-v-in:
I don't think `play-services-core-package` makes a lot of sense. Either make it part of `play-services-core` or `play-services-base-core-package` (to highlight it belongs to code that lives in `play-services-base`, is part of the GmsCore apk package, but does not belong into `play-services-base-core` because apps that use `core` packages should not receive it)

fynngodau:
@mar-v-in Sorry, I missed that word in the upper comment. :upside_down_face: Now it should be correct.

PR #2076, opened 2023-10-22
Licensing UI

fynngodau:
Adds UI for enabling / disabling licensing, as requested in https://github.com/microg/GmsCore/pull/2069#issuecomment-1751786662.

Default: off. `SettingsProvider` is now exposed and protected via two signature protected permissions, `org.microg.gms.permission.READ_SETTINGS` and `.WRITE_SETTINGS`. (I split up reading and writing so that, for example, we can relax `.READ_SETTINGS` to `signatureOrSystem` later on in case a ROM finds use for accessing any microG settings.)

I will add the code that checks for the preference to https://github.com/microg/GmsCore/pull/2069 or a follow-up PR immediately after this PR is merged.

Screenshot_1697983660


mar-v-in (commented):
- Please add [REUSE style](https://reuse.software/tutorial/#step-2) copyright headers to newly added files
- I'd suggest to not call the setting (in code) `play`, as this could mean either Play Store or Play Services. Maybe go for `vending` instead?
- As we plan to soon have futher services in Play Store category, I don't think it makes sense to use the SwitchBar preference here, as it is not the primary action of this category (a normal switch preference should be enough)

fynngodau:
@mar-v-in I've added new commits that address your points.

play-services-base/core/src/main/AndroidManifest.xml
```diff
@@ -5,10 +5,17 @@
   -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
+    <permission android:name="org.microg.gms.permission.READ_SETTINGS"
```
mar-v-in, line 8: `:play-services-base-core` is a public library used by third party apps. Permissions must be scoped to the application (via `${applicationId}`) to not cause conflicts.

play-services-base/core/src/main/AndroidManifest.xml
```diff
@@ -5,10 +5,17 @@
...
+    <permission android:name="org.microg.gms.permission.WRITE_SETTINGS"
+        android:protectionLevel="signature" />
+
     <application>
         <provider
             android:name="org.microg.gms.settings.SettingsProvider"
             android:authorities="${applicationId}.microg.settings"
-            android:exported="false" />
+            android:exported="true"
+            android:readPermission="org.microg.gms.permission.READ_SETTINGS"
```
mar-v-in, line 18: Doesn't that imply that we need to add a `<uses-permission>` for this permission to `:play-services-base-core` itself, or we wouldn't be able to access the provider from within the app itself?

mar-v-in (changes requested):

play-services-base/core/src/main/AndroidManifest.xml
fynngodau: Per [docs](https://developer.android.com/guide/topics/providers/content-provider-basics#Permissions), this is not necessary, as

> components in the provider's application always have full read and write access, regardless of the specified permissions.

play-services-base/core/src/main/AndroidManifest.xml
mar-v-in: You also need to update the permission name used in the `<provider>`'s read/writePermission attribute. ;)

fynngodau: Of course :disappointed: 4123bcf085f4ae0a8d468b701243e4b7f9c39826

mar-v-in (approved):

Issue #1771, opened 2022-08-31
About Play Integrity API


j0110:
Hi,
It would be great to add the possibility to check for Play Integrity API directly with MicroG (as it is already done with Safety Net), to get rid of the real Play Store and just relying on MicroG structure !
Thanks a lot

paolo-caroni:
There is much MIT licensed app that do this, the problem is that microg seems to not support play integrity API yet.

paolo-caroni:
> @j0110 It is handled directly in microG but it is the library compiled in the app itself that check also for Play Store but FakeStore can spoof it almost perfectly.
> 
> There were some code changes in FakeStore that have already been merged that will improve the situation. But there isn't a new version released yet.

Sorry but no, the release of FakeStore 0.2.0 don't solve this issue, Play Integrity API dosn't run, also Play Integrity is not listed as supported on the [wiki](https://github.com/microg/GmsCore/wiki/Implementation-Status).

ale5000-git:
[Play Integrity API Checker](https://play.google.com/store/apps/details?id=gr.nikolasspyr.integritycheck) report success on my phone.

It is possible the problem is not microG.
I'm not completely sure but I think Play Integrity API is just SafeNet with some additional checks and maybe a different interface but the base checks are the same.

paolo-caroni:
Interesting, you have fakstore or google play as market?
![Uploading Screenshot_20230517-215043_Play Integrity API Checker.png…]()

paolo-caroni:
![Screenshot_20230517-215043_Play Integrity API Checker](https://github.com/microg/GmsCore/assets/2750264/90d304a9-32a1-4742-b2a9-7263cb5d0568)

j0110:
> [Play Integrity API Checker](https://play.google.com/store/apps/details?id=gr.nikolasspyr.integritycheck) report success on my phone.
> 
> It is possible that the problem is not microG. I'm not completely sure but I think Play Integrity API is just SafeNet with some additional checks and maybe a different interface but the base checks are the same.

I'm really interested !
How did you succeeded ?

Uldiniad:
One function that needs implementing https://developer.android.com/google/play/integrity/reference/com/google/android/play/core/integrity/IntegrityManager.html#requestIntegrityToken(com.google.android.play.core.integrity.IntegrityTokenRequest)

paolo-caroni:
[...]

PR #2077, opened 2023-10-26
Implement Play Integrity

### LINKED ISSUE #2050: [Play Integrity API] Binding to the service in the play store failed . This is due to having an old play store version installed on the device
Binding to the service in the play store failed . This is due to having an old play store version 

Is this fake store issue or gms core issue

js6pak:
It doesn't work yet, I'm suspecting microg's droidguard implementation, but it also doesn't make sense as there are [recent reports](https://github.com/microg/GmsCore/issues/1771#issuecomment-1672982750) of play integrity working through patched play store (I wasn't able to test it myself).
The same error (`Error retrieving information from server. DF-DFERH-01`) is returned for basically all potential problems, but at this point I'm pretty sure the issue is with the droidguard token, either with the data I pass into it or with the implementation being unable to handle the play integrity flow.
As a side note I tried updating the droidguard version in microg, but it caused even the safetynet check to fail.

TODO
- [ ] make it work
- [ ] move AIDL and other @PublicApi into `play-services-integrity`
- [ ] [express/standard api](https://developer.android.com/google/play/integrity/standard)

Closes #2050

### DIFF
--- a/vending-app/build.gradle
+++ b/vending-app/build.gradle
@@ -4,6 +4,8 @@
  */
 
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+apply plugin: 'com.squareup.wire'
 
 android {
     namespace "com.android.vending"
@@ -11,6 +13,7 @@ android {
     buildToolsVersion "$androidBuildVersionTools"
 
     defaultConfig {
+        multiDexEnabled = true
         versionName vendingAppVersionName
         versionCode vendingAppVersionCode
         minSdkVersion androidMinSdk
@@ -33,13 +36,36 @@ android {
     }
 
     compileOptions {
+        coreLibraryDesugaringEnabled true
+
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
+    }
+
+    kotlinOptions {
+        jvmTarget = 1.8
     }
 }
 
 dependencies {
     implementation project(':fake-signature')
+
+    implementation project(':play-services-droidguard')
+    implementation project(':play-services-tasks-ktx')
+
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:2.0.3")
+    implementation "androidx.annotation:annotation:$annotationVersion"
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:$lifecycleVersion"
+    implementation "androidx.lifecycle:lifecycle-service:$lifecycleVersion"
+
+    implementation "com.squareup.wire:wire-runtime:$wireVersion"
+    implementation("io.ktor:ktor-client-android:2.3.5")
+}
+
+wire {
+    kotlin {
+        javaInterop = true
+    }
 }
 
 if (file('user.gradle').exists()) {
--- a/vending-app/src/main/AndroidManifest.xml
+++ b/vending-app/src/main/AndroidManifest.xml
@@ -9,6 +9,8 @@
         android:name="com.android.vending.CHECK_LICENSE"
         android:protectionLevel="normal" />
 
+    <uses-permission android:name="android.permission.INTERNET" />
+
     <application
         android:forceQueryable="true"
         android:icon="@mipmap/ic_app"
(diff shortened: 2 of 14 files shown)


ale5000-git:
@js6pak 
You have misunderstood, I pass with original (so NOT patched) Play Store intalled in the system partition.

ale5000-git:
Both SafetyNet and Play Integrity pass with official PlayStore.

ale5000-git:
I have got it working with direct system partition modifications (without Magisk), so I don't know if it can work with Magisk.

I will try in the future but I can't try it now.

PS: This is a PR, so isn't the correct place to get help.

huwenkai26:
> 安装在系统分区中

How do you install the official play store + microG gms

huwenkai26:
> @js6pak You have misunderstood, I pass with microG GmsCore + the original (so NOT patched) Play Store intalled in the system partition.

How do you install the official play store + microG gms

BurhanBudak:
Options for this issue is either the fakeStore2PlayStore module that added a shell Play Store SDK 28.

LeVraiRoiDHyrule:
Hi, as of today, what is the best solution to get play integrity to work with MicroG ? I found this : https://github.com/daboynb/PlayIntegrityNEXT that can apparently get device fingerprints automatically. I tried it but still get no valid play integrity. I guess I need something like fakeStore or similar. What is the best solution as of today ? I would like to avoid installing a true play store that could track me. Thanks in advance for any answer and have a nice day

ale5000-git:
@LeVraiRoiDHyrule 
Currently the only way is to use microG Services + real Play Store.
Also now it is more complicated because it need a stock kernel, if you are using a different kernel you should spoof the strings to look like a stock kernel (I cannot help with this).

ale5000-git:
You can probably spoof everything but you need to find the sources of the kernel, change it, compile it and flash it on the device.
Spoof other devices isn't really needed because it only check againts blacklisted words like "lineageos".

ale5000-git:
Actually since there are infinite valid kernel strings they can't whitelist but only blacklist so it isn't hard to fix.
The only problem is that it needed to compile the kernel.
New ROMs will probably be already ok since once the developer know it will fix it, the only problem is with not maintained ROMs.

LeVraiRoiDHyrule:
> @LeVraiRoiDHyrule Currently the only way is to use microG Services + real Play Store. Also now it is more complicated because it need a stock kernel, if you are using a different kernel you should spoof the strings to look like a stock kernel (I cannot help with this).

I see, thanks for the information. Is there a modified minimal play store that would work to avoid the fully featured play store ? Is installing real play store a problem for privacy ?
I am using this microg installer so I plan on doing this : https://github.com/nift4/microg_installer_revived#how-do-i-get-the-real-play-store

Espionage724:
> You can probably spoof everything but you need to find the sources of the kernel, change it, compile it and flash it on the device. Spoof other devices isn't really needed because it only check againts blacklisted words like "lineageos".

Can I get fingerprints and all the official strings from stock OxygenOS and then add it in some files before building LineageOS? I'm curious about avoiding obvious Google blacklist checks and it seems as easy as changing some device-specific text before building; can you provide more details?

ale5000-git:
To all: Please stop all unrelated discussions.

This is a PR so only the ones that want to help or post constructive messages related to the subject should post.
Instead to get help please open a new ticket.

jakubslaby09:
any progress on this?

MoralCode:
Would it be feasible to add an option to microG settings to disable fakestore or install the real playstore (or a stripped down version) so that users have a choice between maximum privacy (at the cost of apps that require the Integrity API not working) or functioning apps? (ideally without rooting)

jakubslaby09:
> Would it be feasible to add an option to microG settings to disable fakestore or install the real playstore (or a stripped down version) so that users have a choice between maximum privacy (at the cost of apps that require the Integrity API not working) or functioning apps? (ideally without rooting)

I think it's a better idea to try to send a PI response from fakestore, like what @js6pak tried to do in 45a3732067f4a79c7b28e94425103ea70e2d2900 , but I don't have experience with microg code.

Weissnix4711:
> Would it be feasible to add an option to microG settings to disable fakestore or install the real playstore (or a stripped down version) so that users have a choice between maximum privacy (at the cost of apps that require the Integrity API not working) or functioning apps? (ideally without rooting)

Afaik the play store must be installed to priv-apps, so I doubt this would be possible. If the device is rooted, sure, but not every device will be. Also, some installation methods already provide the ability to do this, so I'm not sure it's necessary. microg_installer_revived will install any apk you want, be it real or patched. I think nanodroid also allows you the option to install the patched play store.

CoelacanthusHex:
> Would it be feasible to add an option to microG settings to disable fakestore or install the real playstore (or a stripped down version) so that users have a choice between maximum privacy (at the cost of apps that require the Integrity API not working) or functioning apps? (ideally without rooting)

There isn't any option needed. Just install only microG Service and don't install microG Companion, and then install modified Google Play (modification to make purchase feature work). It's unnecessary to install both microG Service and microG Companion.

bphd:
> > Would it be feasible to add an option to microG settings to disable fakestore or install the real playstore (or a stripped down version) so that users have a choice between maximum privacy (at the cost of apps that require the Integrity API not working) or functioning apps? (ideally without rooting)
> 
> There isn't any option needed. Just install only microG Service and don't install microG Companion, and then install modified Google Play (modification to make purchase feature work). It's unnecessary to install both microG Service and microG Companion.

So as of now you can validate PI with MGIR + PS, or not?

ale5000-git:
This is a PR, anything not related to code should not belong here.
I have locked the conversation; if anyone has a constructive comment related to this PR then he/she could always create a new ticket and link this PR.

PR #2082, opened 2023-11-01
Fido: Allow background activity launch

Uldiniad:
Fixes passkey creation

Change-Id: I49c97ad820be35dec1f32944361b4e3b617c1d0f


t-m-w:
Note that this fixes everything related to FIDO/Webauthn for Android 14 (so, not just passkey creation, but also authentication, and also security keys and device-as-security-key). These require presenting the registration or authentication activity to proceed, so they're not functional in Android 14 without this.

PR #2099, opened 2023-11-26
Location: add SNCF/Ouigo location lookup support

VxlerieUwU:
Self-explanatory. Tested with _SNCF_WIFI_INOUI.

Edit: There may be a second endpoint specifically used on Intercités night trains at [www.ombord.info/api/jsonp/position/](www.ombord.info/api/jsonp/position/) using the same SSID as daytime Intercités, though I am unable to verify that.

### DIFF
--- a/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
+++ b/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
@@ -242,6 +242,24 @@ class MovingWifiHelper(private val context: Context) {
         }
         return location
     }
+    
+    private fun parseSncf(location: Location, data: ByteArray): Location {
+            val json = JSONObject(data.decodeToString())
+            if(json.getInt("fix") == -1) throw RuntimeException("GPS not valid")
+            location.accuracy = 100f
+            location.latitude = json.getDouble("latitude")
+	    location.longitude = json.getDouble("longitude")
+            json.optDouble("speed").takeIf { !it.isNaN() }?.let {
+                location.speed = it.toFloat()
+                LocationCompat.setSpeedAccuracyMetersPerSecond(location, location.speed * 0.1f)
+            }
+            location.time = json.getLong("timestamp")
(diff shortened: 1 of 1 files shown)


play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
```diff
@@ -213,6 +213,29 @@ class MovingWifiHelper(private val context: Context) {
         }
         return location
     }
+    
+    private fun parseSncf(location: Location, data: ByteArray): Location {
+            val json = JSONObject(data.decodeToString())
+            if(json.getInt("fix") == -1) { // no fix
+            	return location
```
mar-v-in, line 220: Please throw an exception in this case. Returning the default location object might cause problematic behavior.

mar-v-in (changes requested):
Thanks for working on extending support for additional networks. After all, this can only work as a collaborative community effort, as it's impossible for me to figure out the details for all networks in the world.

Also, allowing different endpoints on wi-fi networks with the same name is indeed likely to be needed in the future. I will look into adjusting the structure of the code to make this easy (especially, we can and should remember which endpoint to query based on mac address, so we won't try all endpoints for a given name again and again). For now, it's fine to focus on the more common case.

play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
VxlerieUwU: Done!

VxlerieUwU:
> Also, allowing different endpoints on wi-fi networks with the same name is indeed likely to be needed in the future. I will look into adjusting the structure of the code to make this easy (especially, we can and should remember which endpoint to query based on mac address, so we won't try all endpoints for a given name again and again). For now, it's fine to focus on the more common case.

Actually, I think that this endpoint isn't used on french night trains anymore, I just saw it mentioned on another github project. And these endpoints do work for 60-70% of french trains.

PR #2123, opened 2023-12-19
Fix the problem that Tasks&Jamboard&Fitbit App cannot log in with Google

DaVinci9196:
1.Fix the problem that Tasks&Jamboard&Fitbit App cannot log in with Google 
2.Declare CAR_SPEED permission

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -90,6 +90,11 @@
         android:description="@string/perm_provision_description"
         android:label="@string/perm_provision_label"
         android:protectionLevel="privileged|signature" />
+    <permission
+        android:name="com.google.android.gms.permission.CAR_SPEED"
+        android:label="@string/perm_card_speed"
+        android:permissionGroup="android.permission-group.LOCATION"
+        android:protectionLevel="dangerous"/>
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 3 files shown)


play-services-core/src/main/AndroidManifest.xml
```diff
@@ -90,6 +90,11 @@
         android:description="@string/perm_provision_description"
         android:label="@string/perm_provision_label"
         android:protectionLevel="privileged|signature" />
+    <permission
+        android:name="com.google.android.gms.permission.CAR_SPEED"
```
mar-v-in, line 94: Which app/issue is fixed by adding the `CAR_SPEED` permission? What about `CAR_MILEAGE` and `CAR_FUEL` permissions?

ale5000-git:
Fitbit isn't a Google apps, also I never had any problem with it.

ale5000-git:
Fitbit isn't a Google apps, also I never had any problem with it.
Tasks correctly use android.permission.GET_ACCOUNTS.
Probably the only needed onw is Jamboard.

play-services-core/src/main/AndroidManifest.xml
DaVinci9196: Fixed the Google Maps location permission authorization pop-up window when connecting to Android Auto. The window still pops up after authorization.

ale5000-git:
@mar-v-in 
If you want I already have a commit with these car related permissions (excluding the signature ones) here: https://github.com/micro5k/GmsCoreMod/commit/6026519041794bf03b68d675860adb3a5aa94cf1.patch

ale5000-git:
How already said Fitbit and Tasks doesn't need this change.

The other parts are superceded by #2214 and #2208.

PR #2129, opened 2023-12-22
supplement and optimize potoken related services

DaVinci9196:
Optimize the PR "completing PoTokenService functions",  thanks for the suggestion

### DIFF
--- a/play-services-api/src/main/aidl/com/google/android/gms/potokens/internal/IPoTokensService.aidl
+++ b/play-services-api/src/main/aidl/com/google/android/gms/potokens/internal/IPoTokensService.aidl
@@ -1,0 +1,15 @@
+/*
+ * SPDX-FileCopyrightText: 2023 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.google.android.gms.potokens.internal;
+
+import com.google.android.gms.common.api.internal.IStatusCallback;
+import com.google.android.gms.potokens.internal.ITokenCallbacks;
+
+interface IPoTokensService {
+    void responseStatus(IStatusCallback call, int code) = 1;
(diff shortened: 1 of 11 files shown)


play-services-api/src/main/aidl/com/google/android/gms/potokens/internal/IPoTokensService.aidl
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package com.google.android.gms.potokens.internal;
+
+import com.google.android.gms.common.api.internal.IStatusCallback;
+import com.google.android.gms.potokens.internal.ITokenCallbacks;
+
+interface IPoTokensService {
+    void responseStatus(IStatusCallback call, int code) = 1;
+    // The direction of 'bArr' is not specified. array can be an in, out, or inout parameter.
```
mar-v-in, line 13: No need to state that. For complex types like arrays or objects, direction needs to be specified. For simple native types like ints and Strings, they are redundant.

mar-v-in (changes requested):
Thanks for doing the changes from the previous pull request.

FYI: You can update pull requests without closing them. Just push a new commit on the branch in your fork or even force push (if you don't want to create a new commit). GitHub will handle it and update the existing pull request accordingly. This makes reviewing easier as I will immediately see what changed and can mark each change request as resolved.

play-services-api/src/main/aidl/com/google/android/gms/potokens/internal/IPoTokensService.aidl
DaVinci9196: okay, changed

mar-v-in (approved):

Issue #212, opened 2016-10-07
Add an option to allow / disallow / ask for a remote installation


ale5000-git:
It would be nice to have an option to auto-allow / auto-disallow / ask for a remote installation when I go on the Play Store site and I choose install on an app's page.

mar-v-in:
Remote installations are managed through Play Store (the Play Store app or in case you use BlankStore/FakeStore, it would be the responsibility of those), not Play Services / GmsCore. You could add this issue to the Phonesky repo, but certainly it won't be fixed soon as Phonesky is not even usable right now :)

ale5000-git:
I was hoping it was in play services, well nevermind.

PR #2130, opened 2023-12-22
supplementary number #212 for authorized login services

DaVinci9196:
completion service <com.google.android.gms.auth.api.identity.service.signin.START>


play-services-basement/src/main/java/com/google/android/gms/common/internal/safeparcel/AbstractSafeParcelable.java
```diff
@@ -21,4 +23,13 @@ public static <T extends AbstractSafeParcelable> SafeParcelableCreatorAndWriter<
     public int describeContents() {
         return 0;
     }
+
+    public byte[] toByteArray() {
```
mar-v-in, line 27: Please use `SafeParcelableSerializer.serializeToBytes()` instead.

play-services-auth/src/main/java/com/google/android/gms/auth/api/identity/BeginSignInRequest.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * SPDX-FileCopyrightText: 2023 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.google.android.gms.auth.api.identity;
+
+import org.microg.safeparcel.AutoSafeParcelable;
+
+public class BeginSignInRequest extends AutoSafeParcelable {
```
mar-v-in, line 10: Please make use of the new safe-parcel-processor wherever easily possible.
1. Ensure that `annotationProcessor project(':safe-parcel-processor')` is added to `dependencies` of corresponding `build.gradle`.
2. Make the class extend `AbstractSafeParcelable` instead of `AutoSafeParcelable`
3. Add `@SafeParcelable.Class` annotiation to the class
4. Make sure the `CREATOR` is defined as `public static final SafeParcelableCreatorAndWriter<XXX> CREATOR = findCreator(XXX.class);`
5. Add `writeToParcel` to the class:
  ```java
  @Override
  public void writeToParcel(@NonNull Parcel dest, int flags) {
      CREATOR.writeToParcel(this, dest, flags);
  }
  ```
6. If necessary, make private fields package-visible or provide a getter (referenced via `getterName`) and constructor parameter for them, tagged with `@Param`.
7. If necessary, tag the constructor to be used when deserializing with `@Constructor` (if no constructor is declared, the default constructor will be used).

Using this system severely reduces the use of reflection at runtime, which can significantly improve performance as it allows for better optimizations.

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
```diff
@@ -197,18 +200,37 @@ class AuthSignInActivity : AppCompatActivity() {
...
+            SignInAccount().apply {
                 email = googleSignInAccount.email ?: account?.name
                 this.googleSignInAccount = googleSignInAccount
-                userId = googleSignInAccount.id ?: getSystemService<AccountManager>()?.getUserData(account, "GoogleUserId")
-            })
+                userId = googleSignInAccount.id ?: getSystemService<AccountManager>()?.getUserData(
+                    account,
+                    AuthConstants.GOOGLE_USER_ID
+                )
+            }.let {
```
mar-v-in, line 215: Why did you change this to a `let` expression?

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
```diff
@@ -197,18 +200,37 @@ class AuthSignInActivity : AppCompatActivity() {
...
+                data.putExtra(AuthConstants.SIGN_IN_ACCOUNT, it)
+            }
+            SignInCredential().apply {
+                email = googleSignInAccount.email
+                displayName = googleSignInAccount.displayName
+                familyName = googleSignInAccount.familyName
+                givenName = googleSignInAccount.givenName
+                idToken = googleSignInAccount.idToken
+            }.let {
```
mar-v-in, line 224: Same question here.

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
```diff
@@ -197,18 +200,37 @@ class AuthSignInActivity : AppCompatActivity() {
...
+                bundle.putByteArray(AuthConstants.SIGN_IN_CREDENTIAL, it.toByteArray())
+                bundle.putByteArray(AuthConstants.STATUS, Status.SUCCESS.toByteArray())
```
mar-v-in, line 226: This line does not even depend on the `SignInCredential` object you just created, so it makes little sense to have it in the same `let` block.

play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/assistedsignin/AssistedSignInService.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.microg.gms.BaseService
+import org.microg.gms.common.GmsService
+
+const val TAG = "AssistedSignInService"
+
+class AssistedSignInService : BaseService(TAG, GmsService.IDENTITY_SIGN_IN) {
+
+    override fun handleServiceRequest(callback: IGmsCallbacks, request: GetServiceRequest, service: GmsService) {
+        val connectionInfo = ConnectionInfo()
+        connectionInfo.features = arrayOf(
```
mar-v-in, line 22: This seems to be the feature set already defined in `org.microg.gms.auth.credentials.FEATURES`, so better don't declare it another time, so we don't have to update at two locations.

play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/assistedsignin/AssistedSignInService.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * SPDX-FileCopyrightText: 2023 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.microg.gms.auth.credentials.assistedsignin
```
mar-v-in, line 6: Please use package `org.microg.gms.auth.credentials.identity` and name `IdentitySignInService` to match AIDL, public API and existing naming schemes.

mar-v-in (changes requested):

play-services-basement/src/main/java/com/google/android/gms/common/internal/safeparcel/AbstractSafeParcelable.java
DaVinci9196: applied

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
DaVinci9196: it changed

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
DaVinci9196: it changed

play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/assistedsignin/AssistedSignInService.kt
DaVinci9196: you are right, I found it is the same

play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/assistedsignin/AssistedSignInService.kt
DaVinci9196: applied

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
DaVinci9196: it changed

play-services-auth/src/main/java/com/google/android/gms/auth/api/identity/BeginSignInRequest.java
DaVinci9196: I don’t understand it very well here. AutoSafeParcelable inherits from AbstractSafeParcelable and uses SafeParcelableCreatorAndWriter.

mar-v-in: `AutoSafeParcelable` is using reflection. Reflection isn't good for performance as it can't be optimized. When using the system based on the annotation processor, as explained above, reflection is not used during (de)serialization. I outlined how to use annotation processor above, you can also take a look at other classes that directly inherit from `AbstractSafeParcelable` without using `AutoSafeParcelable`.

DaVinci9196: Oh oh, I changed it

PR #2155, opened 2024-01-26
Phenotype: Fixed crashing when sharing photos 

DaVinci9196:
Fixed the problem of crashing when sharing photos through links in [photos album]


mar-v-in:
This seems like a very specific hack rather than a proper fix. Based on this I would assume we run into issues with other apps/features as well. I'd rather want to see the correct behavior or a functionally equivalent behavior be implemented.

DaVinci9196:
Okay, I'll try to deal with it

DaVinci9196:
I tried multiple times and an exception occurs when the version in the register method is 0. just return Status.CANCELED at this time to avoid exceptions

Issue #2135, opened 2023-12-27
com.google.android.apps.authenticator2 App won't login with Google Account


ExTBH:
**Affected app**
Name: Google Authenticator
Package id: com.google.android.apps.authenticator2

**Describe the bug**
App doesn't login with google account.

**To Reproduce**
1. Open app
2. Click Sign in
3. Select Account
4. Tap Allow
5. Now it wont log in and ask for account again

**Expected behavior**
App logs in and sync.

**System**
Android Version: EMUI 13.0.1
Custom ROM: 

**microG**
microG Core version: 0.3.0.233515-hw
microG Self-Check results: All Ticked

ExTBH:
I've got a log of the error 
https://paste.c-net.org/AcquiredTamper

and i think its the one at `2023-12-28 17:26:21.314`

it gets logged after tapping `Allow` at Step 4

ExTBH:
decompiling the app i found this, think its the issue?

```java
// 
// Decompiled by Procyon v0.6.0
// 

package com.google.android.gms.audit;

import com.google.android.gms.common.api.internal.TaskApiCall$Builder;
import com.google.android.gms.auth.account.data.InternalGoogleAuthServiceClient$$ExternalSyntheticLambda1;
import com.google.android.gms.common.api.internal.TaskApiCall;
import com.google.android.gms.tasks.Task;
import com.google.android.gms.common.api.Api$ApiOptions;
import com.google.android.gms.common.api.GoogleApi$Settings;
import android.content.Context;
import com.google.android.gms.common.api.Api$AbstractClientBuilder;
import com.google.android.gms.common.api.Api;
import com.google.android.gms.common.api.GoogleApi;

public final class InternalAuditClient extends GoogleApi
{
    private static final Api API;
    private static final Api$AbstractClientBuilder clientBuilder;
    
    static {
        API = new Api("Audit.API", clientBuilder = new InternalAuditClient$1());
    }
    
    public InternalAuditClient(final Context context) {
        super(context, InternalAuditClient.API, null, GoogleApi$Settings.DEFAULT_SETTINGS);
    }
    
    public final Task logAuditRecords(final LogAuditRecordsRequest logAuditRecordsRequest) {
        final TaskApiCall$Builder builder = TaskApiCall.builder();
        builder.methodKey = 6901;
        builder.execute = new InternalGoogleAuthServiceClient$$ExternalSyntheticLambda1(logAuditRecordsRequest, 1);
[...]

PR #2156, opened 2024-01-26
Add dummy service for AUDIT(154)

DaVinci9196:
Solve the problem that Google_Authenticator cannot log in using Google (complete the AuditApiService implementation numbered 154 in GmsService)

### DIFF
--- a/play-services-core/src/main/kotlin/com/google/android/gms/audit/internal/AuditApiService.kt
+++ b/play-services-core/src/main/kotlin/com/google/android/gms/audit/internal/AuditApiService.kt
@@ -1,0 +1,34 @@
+/*
+ * SPDX-FileCopyrightText: 2023 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.google.android.gms.audit.internal
+
+import android.util.Log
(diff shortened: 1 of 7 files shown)


mar-v-in (approved):

ale5000-git:
I have tested it now, it fix #2135

play-services-core/src/main/kotlin/com/google/android/gms/audit/internal/AuditApiService.kt
```diff
@@ -0,0 +1,34 @@
...
+        callback.onPostInitComplete(ConnectionResult.SUCCESS, AuditApiServiceImpl().asBinder(), null)
+    }
+
+}
+
+class AuditApiServiceImpl : IAuditService.Stub() {
+
+    override fun logAuditRecords(logAuditRecordsRequest: LogAuditRecordsRequest?, callback: IStatusCallback) {
+        Log.d(TAG, "method 'logAuditRecords' not fully implemented, only return Status.SUCCESS")
+        callback.onResult(Status.SUCCESS)
```
mar-v-in, line 31: Should return SUCCESS_CACHE (-1) if request.writeMode == 2 (cache)

Issue #1960, opened 2023-06-17
Implement `GoogleMap.OnMapLoadedCallback` correctly


fynngodau:
`GoogleMap.OnMapLoadedCallback` is called as soon as the style is loaded in `GoogleMap.setOnMapLoadedCallback`. [Per docs](https://developers.google.com/maps/documentation/android-sdk/reference/com/google/android/libraries/maps/GoogleMap#public-void-setonmaploadedcallback-googlemap.onmaploadedcallback-callback), it should only be called once rendering is entirely finished:

> Sets a callback that's invoked when this map has finished rendering. The callback will only be invoked once.
>
> If this method is called when the map is fully rendered, the callback will be invoked immediately. This event will not fire if the map never loads due to connectivity issues, or if the map is continuously changing and never completes loading due to the user constantly interacting with the map.

This is unfortunately not trivial to implement. Causes https://github.com/signalapp/Signal-Android/issues/12723#issuecomment-1595810834.

PR #2182, opened 2024-02-18
Workaround implementation for `onMapLoadedCallback`

fynngodau:
With this implementation, a delay is added before calling back an `onMapLoadedCallback`.

Issues:

* As it is just a workaround, it does not guarantee that the map is actually loaded. Thus the snapshot Signal takes can still be empty – however, if the map was zoomed in previously, the tiles are already cached and will be rendered correctly within the 500ms delay. (Unlike a _completely_ empty snapshot, not showing even the MapLibre logo, that was the result before.)
* <del>Sometimes there is a crash in Signal that I still need to investigate where it is coming from.</del> (https://github.com/signalapp/Signal-Android/pull/13461)

Fixes https://github.com/microg/GmsCore/issues/1960. Re. https://github.com/signalapp/Signal-Android/issues/12723#issuecomment-1595810834, see also https://github.com/signalapp/Signal-Android/issues/13006.

signal-location-sharing

### DIFF
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
@@ -527,12 +527,7 @@ class GoogleMapImpl(context: Context, var options: GoogleMapOptions) : AbstractG
         if (callback != null) {
             synchronized(mapLock) {
                 if (loaded) {
-                    Log.d(TAG, "Invoking callback instantly, as map is loaded")
-                    try {
-                        callback.onMapLoaded()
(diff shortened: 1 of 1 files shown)


fynngodau:
The crash I mentioned above (`IllegalArgumentException: View[…] not attached to window manager`) was in fact a race condition within Signal, unrelated to maps itself, and can be fixed with https://github.com/signalapp/Signal-Android/commit/b93e6fb456b5ba4c3ed2d4b07ba5260fc5856756 (furthermore, https://github.com/signalapp/Signal-Android/commit/02a424759932b7755915461de31532fef99138be avoids some similar issue for the future).

jonathanklee:
@fynngodau Looks good to me!

jonathanklee (approved):

PR #2185, opened 2024-02-19
Improve dummy for feedback service

DaVinci9196:
Add some feedback service functions. 
The current method is to jump to the help page webview corresponding to the app.


mar-v-in:
I don't think users expect a webview with the help pages to be opened when they click on the feedback button in Google Apps. Here is an example on how it works in the original Play Services:
image
If they would want to open the help pages, they would click on Help (which we already support). I doubt there is any need to implement the feedback flow (or it being worth the effort), so I'm not suggesting to actually implement it, but rather to just display a dialog or toast "Sending feedback is not supported on this device." should be sufficient.

DaVinci9196:
@mar-v-in I added a feedback page, but the content of the report request has not been processed, and the feedback cannot be successful at present.
Screenshot_20240220_105605

mar-v-in:
@DaVinci9196 I appreciate your work, but I doubt there is any need to implement this festure in the same way as Google:
- I doubt a lot of users ever tried pressing that feedback button.
- if they did, it was most likely accidentally or because they're searching for some other feature.
- If a user of a Google App on microG genuinely wants to send feedback, it's most likely because something doesn't work as a result of an issue with microG. It wouldn't make sense to send such feedback to Google.

We may want to add a bug reporting feature directly in microG (that collects bug information directly on the device and sends it to some microG server). As long as we don't have that, it's probably better to not display any feedback form at all and just show an "Feedback currently not possible" dialog, as to not confuse anyone to actually enter a feedback message that couldn't be send anyway.

ale5000-git:
But the feedback "moved" to be a bug reporting to microG.

mar-v-in:
> But the feedback "moved" to be a bug reporting to microG (sent to an e-mail created by you).

Yes exactly. But many details of this dialog would need to be adjusted. E.g. we should allow the user to send the bug report anonymously (not tied to their Google account) or allow them to enter a custom email address. And obviously we wouldn't send anything about the device that the user can't decide to not send.

mar-v-in (approved):

PR #2188, opened 2024-02-19
Add user-install microG User Services

p1gp1g:
Following discussion on #2183  (which may be closed ? Superseded by this one)

@mar-v-in @ale5000-git

I have tested, and org.microg.gms can't be installed on a system with com.google.android.gms (Google or microG) because of _duplicate permission_.

I have used the dimension 'target' for the user flavor, I think that's the one to use.

### DIFF
--- a/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
+++ b/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
@@ -74,6 +74,10 @@ public class MultiConnectionKeeper {
         return signatureIsIn(pm, packageId, signatures);
     }
 
+    private Boolean isSystemGoogleOrMicrogSig(PackageManager pm, String packageId) throws PackageManager.NameNotFoundException {
+        return isSystem(pm, packageId) || isGoogleOrMicrogSig(pm, packageId);
+    }
+
     private Boolean isMicrogSig(PackageManager pm, String packageId) throws PackageManager.NameNotFoundException {
         List<String> signatures = Arrays.asList(MICROG_PRIMARY_KEYS);
         return signatureIsIn(pm, packageId, signatures);
@@ -112,7 +116,7 @@ public class MultiConnectionKeeper {
         // Pref: gms > microG > self
         PackageManager pm = context.getPackageManager();
         try {
-            if (isSystem(pm, GMS_PACKAGE_NAME) || isGoogleOrMicrogSig(pm, GMS_PACKAGE_NAME)) {
+            if (isSystemGoogleOrMicrogSig(pm, GMS_PACKAGE_NAME)) {
                 Log.d(TAG, GMS_PACKAGE_NAME + " found !");
                 return GMS_PACKAGE_NAME;
             } else {
@@ -242,15 +246,24 @@ public class MultiConnectionKeeper {
         private Intent getIntent() {
             Intent intent;
             ResolveInfo resolveInfo;
+            PackageManager pm = context.getPackageManager();
             if (!Objects.equals(targetPackage, context.getPackageName())) {
                 intent = new Intent(actionString).setPackage(targetPackage);
-                if ((resolveInfo = context.getPackageManager().resolveService(intent, 0)) != null) {
-                    if (requireMicrog && !isMicrog(resolveInfo)) {
-                        Log.w(TAG, "GMS service found for " + actionString + " but looks not like microG");
-                    } else {
-                        Log.d(TAG, "GMS service found for " + actionString);
-                        return intent;
+                try {
+                    if ((resolveInfo = context.getPackageManager().resolveService(intent, 0)) != null) {
+                        if (requireMicrog && !isMicrog(resolveInfo)) {
+                            Log.w(TAG, "GMS service found for " + actionString + " but looks not like microG");
+                        } else {
+                            if (isSystemGoogleOrMicrogSig(pm, targetPackage)){
+                                Log.d(TAG, "GMS service found for " + actionString);
+                                return intent;
+                            } else {
+                                Log.w(TAG, "GMS service found for " + actionString + " but is not system, and doesn't have microG or Google signature");
+                            }
+                        }
                     }
+                } catch (PackageManager.NameNotFoundException e) {
+                    Log.d(TAG, targetPackage + " not found");
                 }
             }
             intent = new Intent(actionString).setPackage(context.getPackageName());
--- a/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
+++ b/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
@@ -36,6 +36,7 @@ import java.util.Set;
 import static android.os.Build.VERSION.SDK_INT;
 import static android.os.Build.VERSION_CODES.ICE_CREAM_SANDWICH;
 import static org.microg.gms.common.Constants.GMS_PACKAGE_NAME;
+import static org.microg.gms.common.Constants.MICROG_GMS_PACKAGE_NAME;
 
 public class MultiConnectionKeeper {
     private static final String TAG = "GmsMultiConKeeper";
@@ -138,32 +139,50 @@ public class MultiConnectionKeeper {
             this.requireMicrog = requireMicrog;
         }
 
+        private Intent getIntent() {
+            Intent gmsIntent = new Intent(actionString).setPackage(GMS_PACKAGE_NAME);
+            Intent microGIntent = new Intent(actionString).setPackage(MICROG_GMS_PACKAGE_NAME);
+            Intent selfIntent = new Intent(actionString).setPackage(context.getPackageName());
+            ResolveInfo resolveInfo;
+            // Pref: gms > microG > self
+            // Note that gms and microG can't be both installed on the system because of duplicate
+            // permissions. It is possible to build an app without these permission, but it won't
+            // be a regular build
+            if ((resolveInfo = context.getPackageManager().resolveService(gmsIntent, 0)) != null) {
+                if (requireMicrog && !isMicrog(resolveInfo)) {
+                    Log.w(TAG, "GMS service found for " + actionString + " but looks not like microG");
+                } else {
+                    Log.d(TAG, "GMS service found for " + actionString);
+                    return gmsIntent;
+                }
+            }
+
+            if ((resolveInfo = context.getPackageManager().resolveService(microGIntent, 0)) != null) {
+                if (requireMicrog && !isMicrog(resolveInfo)) {
+                    // If the user has somehow installed another application with the microG package id
+                    Log.w(TAG, "microG service found for " + actionString + " but looks not like microG");
+                } else {
+                    Log.d(TAG, "microG service found for " + actionString);
+                    return microGIntent;
+                }
+            }
+
+            if (context.getPackageManager().resolveService(selfIntent, 0) != null) {
+                Log.d(TAG, "Found service for " + actionString + " in self package, using it instead");
+                return selfIntent;
+            }
+            return null;
+        }
+
         @SuppressLint("InlinedApi")
         public void bind() {
             Log.d(TAG, "Connection(" + actionString + ") : bind()");
-            Intent gmsIntent = new Intent(actionString).setPackage(GMS_PACKAGE_NAME);
-            Intent selfIntent = new Intent(actionString).setPackage(context.getPackageName());
             Intent intent;
-            ResolveInfo resolveInfo;
-            if ((resolveInfo = context.getPackageManager().resolveService(gmsIntent, 0)) == null) {
-                Log.w(TAG, "No GMS service found for " + actionString);
-                if (context.getPackageManager().resolveService(selfIntent, 0) != null) {
-                    Log.d(TAG, "Found service for " + actionString + " in self package, using it instead");
-                    intent = selfIntent;
-                } else {
-                    return;
-                }
-            } else if (requireMicrog && !isMicrog(resolveInfo)) {
-                Log.w(TAG, "GMS service found for " + actionString + " but looks not like microG");
-                if (context.getPackageManager().resolveService(selfIntent, 0) != null) {
-                    Log.d(TAG, "Found service for " + actionString + " in self package, using it instead");
-                    intent = selfIntent;
-                } else {
-                    intent = gmsIntent;
-                }
-            } else {
-                intent = gmsIntent;
+            if ((intent = getIntent()) == null) {
+                Log.w(TAG, "No service found for " + actionString);
+                return;
             }
+
             int flags = Context.BIND_AUTO_CREATE | Context.BIND_DEBUG_UNBIND;
             if (SDK_INT >= ICE_CREAM_SANDWICH) {
                 flags |= Context.BIND_ADJUST_WITH_ACTIVITY;
--- a/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
+++ b/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
@@ -32,14 +32,13 @@ import android.util.Log;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 
 import static android.os.Build.VERSION.SDK_INT;
 import static android.os.Build.VERSION_CODES.ICE_CREAM_SANDWICH;
 import static android.os.Build.VERSION_CODES.TIRAMISU;
 import static org.microg.gms.common.Constants.GMS_PACKAGE_NAME;
-import static org.microg.gms.common.Constants.USER_GMS_PACKAGE_NAME;
+import static org.microg.gms.common.Constants.GMS_MINUS_PACKAGE_NAME;
 
 public class MultiConnectionKeeper {
     private static final String TAG = "GmsMultiConKeeper";
@@ -79,21 +78,21 @@ public class MultiConnectionKeeper {
                 Log.d(TAG, GMS_PACKAGE_NAME + " found !");
                 return GMS_PACKAGE_NAME;
             } else {
-                Log.w(TAG, USER_GMS_PACKAGE_NAME + " found with another name: " + name);
+                Log.w(TAG, GMS_MINUS_PACKAGE_NAME + " found with another name: " + name);
             }
         } catch (PackageManager.NameNotFoundException e) {
             Log.d(TAG, GMS_PACKAGE_NAME + " not found");
         }
         try {
-            String name = getApplicationName(USER_GMS_PACKAGE_NAME);
+            String name = getApplicationName(GMS_MINUS_PACKAGE_NAME);
             if (name.contains("microG")) {
-                Log.d(TAG, USER_GMS_PACKAGE_NAME + " found !");
-                return USER_GMS_PACKAGE_NAME;
+                Log.d(TAG, GMS_MINUS_PACKAGE_NAME + " found !");
+                return GMS_MINUS_PACKAGE_NAME;
             } else {
-                Log.w(TAG, USER_GMS_PACKAGE_NAME + " found with another name: " + name);
+                Log.w(TAG, GMS_MINUS_PACKAGE_NAME + " found with another name: " + name);
             }
         } catch (PackageManager.NameNotFoundException e) {
-            Log.d(TAG, USER_GMS_PACKAGE_NAME + " not found");
+            Log.d(TAG, GMS_MINUS_PACKAGE_NAME + " not found");
         }
         return null;
     }
@@ -105,7 +104,7 @@ public class MultiConnectionKeeper {
         if ((target = prefs.getString(PREF_TARGET, null)) != null) {
             switch (target) {
                 case GMS_PACKAGE_NAME:
-                case USER_GMS_PACKAGE_NAME:
+                case GMS_MINUS_PACKAGE_NAME:
                     return target;
                 case SELF:
                     return null;
--- a/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
+++ b/play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
@@ -38,6 +38,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import static android.os.Build.VERSION.SDK_INT;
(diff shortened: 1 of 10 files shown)


ale5000-git:
@p1gp1g
If I'm not wrong, permissions can be defined more than one time if both apps are signed with the same signature.

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
```diff
@@ -36,6 +36,7 @@
 import static android.os.Build.VERSION.SDK_INT;
 import static android.os.Build.VERSION_CODES.ICE_CREAM_SANDWICH;
 import static org.microg.gms.common.Constants.GMS_PACKAGE_NAME;
+import static org.microg.gms.common.Constants.MICROG_GMS_PACKAGE_NAME;
```
ale5000-git, line 39: Wouldn't be better something like USER_GMS_PACKAGE_NAME?
Both are still microG.

ale5000-git:
Now it fails build:
```
/home/runner/work/GmsCore/GmsCore/play-services-core/src/main/AndroidManifest.xml:137:9-36 Error:

	Attribute application@label at AndroidManifest.xml:137:9-36 requires a placeholder substitution but no value for <appLabel> is provided.
/home/runner/work/GmsCore/GmsCore/play-services-core/src/huawei/AndroidManifest.xml Error:
	Validation failed, exiting
> Task :play-services-core:processHmsHuaweiDebugMainManifest FAILED
```

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
p1gp1g: done

p1gp1g:
I've not tested locally, but it should build now. I'll rebase to *fix* the commit into the previous one later.

---

> @p1gp1g If I'm not wrong, permissions can be defined more than one time if both apps are signed with the same signature.

IIRC signature spoofing works only for system app, so microG (system) should have a different signature. I think we should test this case.

---

A few things should be done on the UI for the user services:
- a dialog on first use, explaining that:
    -  this service is experimental (?)
    -  microG Services use signature spoofing to work with applications using the google libraries, this build does not use this technique and works only with application which use the microG libraries replacement.
    - Help is welcome to word this :)
- Remove the signature spoofing check for the user services
- Maybe remove some other things ?

mar-v-in:
microG Services can be installed as a user / non-system app already. (If that's not possible on GrapheneOS that might be a GrapheneOS specific issue and would be interesting to analyze that) As a matter of fact, I personally run microG as a user app myself (using my homebrowed OS that's just very few modifications on top of AOSP).

Many systems would not allow signature spoofing when installing microG as a user app. While signature spoofing might be restricted to system apps on systems that come with microG, that doesn't mean it needs to be. DivestOS for example supports signature spoofing for microG running as a user app.

Signature spoofing does not have an impact on any (security) feature of the OS that are signature specific (like signature shared permission declarations). Signature spoofing only affects apps that ask the OS for the signing certificate of an app - which essentially is only the play services client library, because use of that OS feature is discouraged by Google for being insecure as a security mechanism (it's possible to run race attacks against it) and unsuitable for anything else.

Some OS include modified or resigned versions of microG Services. Those obviously have a different signature and can't be upgraded from the original microG package. Otherwise, I would nornally expect both variants to be signed using the same signing key, so that it's easy for users to verify.

To prevent users from installing both, we can and probably should just display appropriate messages when the user does so. Reworking the self-check has been on my list for some time already, so this could become part of it.

Given what I wrote I think the branding as "microG User Services" is very much misleading. We should come up with another name before publishing this. Something along the lines of "microG Services without Google Play Services compatibility", which would be more descriptive but not a nice name ;)

p1gp1g:
I'll use _CGAG_ for services with id _com.google.android.gms_ and _OMG_ to talk about services with id _org.microg.gms_.

---

> Many systems would not allow signature spoofing when installing microG as a user app. While signature spoofing might be restricted to system apps on systems that come with microG, that doesn't mean it needs to be. DivestOS for example supports signature spoofing for microG running as a user app.

TIL about DivestOS doing it. Still, you shouldn't be able to install OGM + CGAG because of signature spoofing for the later. Nevertheless, some user without signature spoofing at all may try to install both on their system (which wouldn't work with CGAG).

And even if it were possible to install both, a user installing OMG when CGAG is already installed wouldn't be an issue since CGAG has always the priority.

The main issue is :
1. If a user install CGAG after OMG, then CGAG will take the priority, and some data may be lost (like things needed in shared pref)
2. If a user install CGAG or OMG after using an application embedding a -core (like the Corona tracing app - CCTG). Then, CGAG or OMG will take the priority, and some data may be lost (like things needed in shared pref)

_Note that the 2nd issue is already there._

So. Maybe we should add a dialog to CGAG:
- if it has been installed after the system install, that some application may use it instead of internal mechanism and the user may need to login again or something like this (to make it understandable)

> To prevent users from installing both, we can and probably should just display appropriate messages when the user does so. Reworking the self-check has been on my list for some time already, so this could become part of it.

It isn't important if both is installed, CGAG takes the priority.

> Given what I wrote I think the branding as "microG User Services" is very much misleading. We should come up with another name before publishing this. Something along the lines of "microG Services without Google Play Services compatibility", which would be more descriptive but not a nice name ;)

I know, that's a temporary name :)

microG Minus Services ?

mar-v-in:
> Still, you shouldn't be able to install OGM + CGAG because of signature spoofing for the later.

Signature spoofing has **no**  impact here.

> Signature spoofing does not have an impact on any (security) feature of the OS that are signature specific (like signature shared permission declarations). Signature spoofing only affects apps that ask the OS for the signing certificate of an app

p1gp1g:
Sorry, I missed that point, I've edited the previous comment

p1gp1g:
The last commit introduced a function to remember the service used by the application. That way, if I use the embedded core, then the app won't use an external one after OMG/CGAG install. For instance, CCTG won't lose its data.

Also, I adds a check to be sure microG/Play services is legit to reduce risks of malicious app branded as another. It assures the service
- Has an activity (so it will be listed in the installed app)
- Has a legit name (Note: This is note a space 0x20 between *Google* and *Play* for CGAG)

I think this is important because: if a user doesn't have play services/microG with id *com.google.android.gms* (big *if*), a malicious app installed with id *com.google.android.gms*, branded has another service (a cooking app for instance), may use the FIDO service to get access to some services:
1 When the user wants to do a FIDO authent, a request is send to the FIDO service
2. The malicious app uses the key activation to register to another service (the user think it is a legit login)
3. The malicious app sends the response to a remote server and it displays a screen saying there is an error and to retry
4. Do the user's legit login


PS: I don't think the CI Out Of Memory error is related to this PR, isn't it ? Could it be restarted ?

ale5000-git:
@p1gp1g

Quote:
> If you use the gradle-android-plugin to build your app, then you can use
> BuildConfig.APPLICATION_ID
> to retrieve the package name from any scope, incl. a static one.

Isn't this simpler?

ale5000-git:
For the microG self check it should be the simplest way (it also avoid passing context), for other cases I don't know; I don't have time to try currently.

mar-v-in:
In `AndroidManifest.xml` you can use `${applicationId}` to get the current application id, never tried if that works for other resource files

p1gp1g:
> Yes, for the microG self check it should be the simplest way (it also avoid passing context), for other cases I don't know; I don't have time to try currently.

You mean the check in function `MultiConnectionKeeper.getTargetPackageWithoutPref` ? This is intended to be a library to service check (eg. Browser requests microG)

p1gp1g:
The packageId must be shared with a resValue if it used in resource files (eg. https://stackoverflow.com/questions/57074999/get-application-id-in-xml-dynamically)

p1gp1g:
I have renamed the services *microG Minus Services*, this will be less confusing. Feel free to change the name

I have also added a dialog on first run to inform the user that this flavor is only compatible with microG libraries, I hope it will reduce users misunderstanding.

ArchangeGabriel:
(JFTR, since it was discussed earlier, there is now a major ROM allowing signature spoofing as user app: https://review.lineageos.org/c/LineageOS/android_frameworks_base/+/383573)

p1gp1g:
@ArchangeGabriel 

> This patch enables signature spoofing when the following conditions are
met:
> * Build is debuggable (userdebug/eng)
> * Package name is com.android.vending or com.google.android.gms
> * **Package is signed with microG release keys**
> * **Fake signature is correct**

I'm OK to check the microG signature (or Google Play signature) instead of controlling the application name: https://github.com/microg/GmsCore/pull/2188/files#diff-4fe7d2cf47e0e66eec592e3286e77a70323b66a82f5d43bfe7f903623230525dR77 . I've done that because some other project may want to fork microG/reimplement (I've seen one exists). I can change it if you want

ale5000-git:
@ArchangeGabriel
This PR is about a user build of microG, it has a different internal name so it won't be used by app compiled with Google libraried but only by app compiled wuth microG libraries so signatute spoofing isn't needed at all.

p1gp1g:
But this PR introduces 2 things too:
- the application using the library saves the targetted packageId, so it won't lose data if a service is installed after the first use
- a check [here](https://github.com/microg/GmsCore/pull/2188/files#diff-4fe7d2cf47e0e66eec592e3286e77a70323b66a82f5d43bfe7f903623230525dR77) (posted on the previous post) to ensure another app is not trying to spoof microG. This checks if the app 1. has an activity (to be listed in user app) 2. use a name with _Google Play_ or _microG_ so it won't be branded as something else. This second point could be changed into a signature check.

ale5000-git:
Many ROMs ship a custom microG with a different signature, so a signature check isn't a good thing.

mar-v-in:
I don't think this name check is really necessary or a good idea. I would prefer if forks of microG would not call themselves microG (to not confuse users), but this reinforces them to actually have microG in their name.

I personally don't think apps taking over the package name is a huge risk as there is a ton of other ways that malware could affect your system if you install apps blindly from untrusted source. But anyway, I'd propose to do at most the following:
For the package name `com.google.android.gms`, check if it's a system app or signed using Google or microG certificate and if not, introduce some user interaction, asking the user to confirm they want to use services from that app (displaying it's name and maybe even a signature fingerprint). Same for the `org.microg.gms` package name, except that Google certificate is not accepted for that.

ArchangeGabriel:
> @ArchangeGabriel This PR is about an alternative user build of microG; it has a different internal name so it won't be used by apps compiled with Google libraries but only by apps compiled with microG libraries so signatute spoofing isn't needed at all.

I know, I was just replying to:
>IIRC signature spoofing works only for system app, so microG (system) should have a different signature. I think we should test this case.

from @p1gp1g, even though @mar-v-in already addressed that with:

>Many systems would not allow signature spoofing when installing microG as a user app. While signature spoofing might be restricted to system apps on systems that come with microG, that doesn't mean it needs to be. DivestOS for example supports signature spoofing for microG running as a user app.

I was just wanting to point out that allowing signature spoofing as user app just become suddenly much more common. ;)

p1gp1g:
@mar-v-in 
> I don't think this name check is really necessary or a good idea. I would prefer if forks of microG would not call themselves microG (to not confuse users), but this reinforces them to actually have microG in their name.
> 
> I personally don't think apps taking over the package name is a huge risk as there is a ton of other ways that malware could affect your system if you install apps blindly from untrusted source. But anyway, I'd propose to do at most the following: For the package name `com.google.android.gms`, check if it's a system app or signed using Google or microG certificate and if not, introduce some user interaction, asking the user to confirm they want to use services from that app (displaying its name, app icon and maybe even a signature fingerprint). Same for the `org.microg.gms` package name, except that Google certificate is not accepted for that. The confirmation is obviously to be persisted. This way, most users will have everything work silently, but those that decide to use a custom microG will have to confirm once (per app) to use it.

I'm OK with that. I've set a signature check, tested with Google Play, if someone can test it with microG.

I've started to look into the user confirmation when the signature doesn't match, it introduces 2 main changes:
- Setting gmsPackage will be async
- It needs an activity to open an alert box or something.

I think the user confirmation can fit another PR, but I need to know what's your way to deal with those 2 points :)

ale5000-git:
@p1gp1g
I don't think `GOOGLE_APP_KEY` should be there, that is for other google apps, not Google Play Services.
Also the check should be skipped for system apps (or maybe just for system privileged apps).

p1gp1g:
@ale5000-git , I'll remove GOOGLE_APP_KEY then.

`isGooglePackage` and the other functions used for package signature check are all in the *base-core* module and are needed in the *base* module. I thought you were avoiding importing core modules into non-core modules

ale5000-git:
> isGooglePackage and the other functions used for package signature check are all in the base-core module and are needed in the base module. I thought you were avoiding importing core modules into non-core modules

I meant `PackageUtils.firstSignatureDigest` but we should wait a reply from @mar-v-in

ale5000-git:
Here for example: https://github.com/microg/GmsCore/blob/master/play-services-nearby/core/src/main/kotlin/org/microg/gms/nearby/exposurenotification/ExposureNotificationServiceImpl.kt

It use:
```
import org.microg.gms.common.PackageUtils
PackageUtils.firstSignatureDigest(context, packageName)
```

p1gp1g:
I was thinking about `PackageUtils.firstSignatureDigest` when I said *and the other functions* :+1: 

The thing is ExposureNotificationServiceImpl.kt is a core module (nearby-core). We can also move the PackageUtils to the *base* module

ale5000-git:
Beside that, it say: `Lint found 1 errors`

ale5000-git:
@mar-v-in
What about "microG Limited Services" as name?
In this way people will certainly not use it by default :-)

p1gp1g:
microG Limited Services looks good :+1:

p1gp1g:
I've changed the new app name to *microG Limited Services*. There is already a dialog explaining how it is different from the regular one. I think this should be clear enough for the users

microG lib will now connect to the PlayServices if:
- CGAG is installed as a system app
- CGAG is installed with Google signatures (can be spoofed)
- CGAG is installed with microG signatures
- OMG is installed with microG signatures

I think this is enough for any fork to work. In addition, a user interaction to use CGAG or OMG with other signatures (user installed) is still good to implement

p1gp1g:
@mar-v-in  Is it possible to add it to the next milestone ?

p1gp1g:
@mar-v-in What is missing for this PR to be merged ? I've seen you have changed the milestone but the PR is ready and nothing seems blocking.

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
```diff
@@ -21,32 +21,142 @@
...
 import static android.os.Build.VERSION_CODES.ICE_CREAM_SANDWICH;
 import static org.microg.gms.common.Constants.GMS_PACKAGE_NAME;
+import static org.microg.gms.common.Constants.MICROG_PACKAGE_NAME;
+import static org.microg.gms.common.Constants.GMS_PACKAGE_SIGNATURE_SHA1;
+import static org.microg.gms.common.Constants.GMS_SECONDARY_PACKAGE_SIGNATURE_SHA1;
+import static org.microg.gms.common.Constants.MICROG_PACKAGE_SIGNATURE_SHA1;
 
 public class MultiConnectionKeeper {
     private static final String TAG = "GmsMultiConKeeper";
+    private static final String PREF_MASTER = "GmsMultiConKeeper";
```
mar-v-in, line 53: The preference file name should be properly scoped to ensure there won't be conflicts with the app using the library. Use a name like `org.microg.gms` instead. Also would be preferably to call the constant `PREFERENCES_FILE_NAME` or similar.

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
```diff
@@ -21,32 +21,142 @@
...
+    private static final String PREF_TARGET = "GmsMultiConKeeper_target";
```
mar-v-in, line 54: The preference name is already scoped through the file name (`PREF_MASTER`), so for readability I think it would be preferably to not prefix the preference name with the file name.

play-services-basement/src/main/java/org/microg/gms/common/Constants.java
```diff
@@ -21,9 +21,11 @@
 public class Constants {
     public static final int GMS_VERSION_CODE = (BuildConfig.VERSION_CODE / 1000) * 1000;
     public static final String GMS_PACKAGE_NAME = "com.google.android.gms";
+    public static final String MICROG_PACKAGE_NAME = "org.microg.gms";
```
mar-v-in, line 24: Rename to `USER_MICROG_PACKAGE_NAME` or similar.

play-services-core/src/main/java/org/microg/gms/ui/MainSettingsActivity.java
```diff
@@ -1,29 +1,53 @@
...
+
     private NavController getNavController() {
         return ((NavHostFragment)getSupportFragmentManager().findFragmentById(R.id.navhost)).getNavController();
     }
 
+    private void showDialogIfNeeded() {
+        SharedPreferences prefs = getSharedPreferences(FirstRunMaster, MODE_PRIVATE);
+        if (BuildConfig.APPLICATION_ID == Constants.MICROG_PACKAGE_NAME &&
+                prefs.getBoolean(FirstRunPref, true)) {
+            AlertDialog.Builder builder = new AlertDialog.Builder(this);
```
mar-v-in, line 39: Use `buildAlertDialog()` from `org.microg.gms.ui.buildAlertDialog` (which will use Material design when applicable)

play-services-core/src/main/java/org/microg/gms/ui/MainSettingsActivity.java
```diff
@@ -1,29 +1,53 @@
...
+import org.microg.gms.common.Constants;
 import org.microg.gms.ui.settings.SettingsProvider;
 
 import static org.microg.gms.ui.settings.SettingsProviderKt.getAllSettingsProviders;
 
 public class MainSettingsActivity extends AppCompatActivity {
     private AppBarConfiguration appBarConfiguration;
 
+    private static final String FirstRunMaster = "FirstRun";
+    private static final String FirstRunPref = "FirstRun_pref";
```
mar-v-in, line 29: Constants are all caps

mar-v-in, line 29: As above, don't prefix the preference name with the file name

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
```diff
@@ -21,32 +21,142 @@
...
+    private static final String[] GOOGLE_PRIMARY_KEYS = {GMS_PACKAGE_SIGNATURE_SHA1, GMS_SECONDARY_PACKAGE_SIGNATURE_SHA1};
 
+    private static final String[] MICROG_PRIMARY_KEYS = {MICROG_PACKAGE_SIGNATURE_SHA1};
     private static MultiConnectionKeeper INSTANCE;
 
     private final Context context;
+
+    private final String gmsPackage;
```
mar-v-in, line 62: `targetPackage` would be a better name IMO.

play-services-base/src/main/java/org/microg/gms/common/MultiConnectionKeeper.java
```diff
@@ -21,32 +21,142 @@
...
+            }
+        } catch (PackageManager.NameNotFoundException e) {
+            Log.d(TAG, MICROG_PACKAGE_NAME + " not found");
+        }
+        return null;
+    }
+
+    private String getTargetPackage() {
+        SharedPreferences prefs = context.getSharedPreferences(PREF_MASTER, Context.MODE_PRIVATE);
+        final String SELF = "SELF";
```
mar-v-in, line 138: Instead of special handling for `SELF` / `null`, you can just use `context.getPackageName()` (both in the preference and returned from `getTargetPackage()` and extend `isMicrog()` to return true if the service's package name is the current package name (because the service won't be found inside the same package if it's not microG)

mar-v-in (changes requested):
Sorry for not replying yet. I didn't have a lot of time to do testing yet, so this round of review is only on the code. While it looks as if it should work fine, I want to at least run this once before merging.

p1gp1g:
Gentle ping for the review @mar-v-in

mar-v-in:
Sorry @p1gp1g, I keep moving this to the next milestone. Testing this is a little more involved, so I haven't managed to finish it yet and I don't want to block the release any longer. I'll do my best to finally do this in the next weeks

p1gp1g:
Is it possible to help testing it ? I really wish to use FIDO on my phone :smiling_face_with_tear: 

The only important change is in MultiConnectionKeeper

mar-v-in:
My plan was to hit the merge button right after the 0.3.3 release (which is due to happen in the next days), so I will force myself to test it while on master. 0.3.4 is expected to happen in October, so it won't be too long.

p1gp1g:
Great to hear ! Let me know if you need anything (I'm in the matrix room too)

Vavun:
Hi
Since this pull request is ready enough to get into the main branch, I'd like to ask a naive question about naming.

Why limited?

At the moment, this microg variant provides relatively identical opportunities for applications.
I am sure that in the future functionality can even become extended, since there is no need to mimic Google services for the operating system and applications.

p1gp1g:
It is _limited_ as it only works with applications using the microg libraries.

Also, some features will not work, because they require system rights. For example the capability to add an application in power-saving whitelist: https://github.com/microg/GmsCore/blob/511afe821b20a5545f2ecc3f8db92a3ecb24d87f/play-services-core/src/main/java/org/microg/gms/gcm/McsService.java#L613

For this exact feature, we may check if the OS supports GMSCompatibility (like GraphenOS and DivestOS). They add this service to all application using firebase-messaging: https://github.com/GrapheneOS/platform_frameworks_base/blob/4e0f5bc2fab938de0433b19a0c4eb869a60beccd/core/java/com/android/internal/gmscompat/client/GmsCompatClientService.java#L34. It is possible to bind with foreground priority to this service for some seconds, it will achieve the same thing than whitelisting the app

Arbel-arad:
> Why limited?
> 
> At the moment, this microg variant provides relatively identical opportunities for applications. I am sure that in the future functionality can even become extended, since there is

Hi, have you managed to get this working on GrapheneOS?
i am trying to use FIDO on ironFox but the developer says microG is required, though trying to install the -user apk results in a conflict.

p1gp1g:
> Hi, have you managed to get this working on GrapheneOS? i am trying to use FIDO on ironFox but the developer says microG is required, though trying to install the -user apk results in a conflict.

It doesn't work for GrapheneOS because of conflicting permission, they install GmsCompat by default on the OS. Maybe that can be installed only if play services are installed

mar-v-in:
GrapheneOS has a lot of special handling and os modifications around gms. The respective permissions should not be declared in GmsCompat, but rather in GmsCore itself (or microG). If you install regular microG on GrapheneOS it probably works, because if the package `com.google.android.gms` declares the permission (as the official GmsCore does), GrapheneOS will just ignore it, but all other package names are treated different.

See https://github.com/GrapheneOS/platform_frameworks_base/blob/f3469eb751ca3c8df10b1be4f7925e86436d09f0/services/core/java/com/android/server/pm/ext/GmsCoreHooks.java

PR #2189, opened 2024-02-20
Return proper dummy value in AppSetService.getAppSetIdInfo

DaVinci9196:
open App crash  eg:CallApp、webtoon、Parallel Space


ale5000-git:
@mar-v-in
This fix is very important.
This app: https://play.google.com/store/apps/details?id=com.liuzh.deviceinfo
and maybe also other crash continuosly without this fix.

PR #2193, opened 2024-02-21
Add fr, de, it and es translations for Licensing App feature

jonathanklee:


mar-v-in:
I copied those over to weblate.

PR #2204, opened 2024-02-26
Extend authorized domain scope

DaVinci9196:
need USER_BIRTHDAY_READ  e.g.  Wattpad - Read & Write Stories
need GMAIL_READONLY  e.g. CallApp - Caller ID & Block


mar-v-in:
We currently only inform users that "name, email address and profile picture" are shared with the requesting app. If we authorize more permissions, we should list them so that the user knows. Especially gmail.readonly is a highly sensitive permission (as it allows reading all your emails) and users probably don't want to grant it to untrusted third-party apps.

ale5000-git:
Isn't there a way to know what app it wanting and allow the user to choose with things like "checkbox"?

DaVinci9196:
> Surely is possible, but likely some apps are going to break if we return a token with less permissions than they expected.
> 
> We should first explore what Google does in cases where additional permissions are requested using this API. There is another API which we already support that explicitly allows for requesting various permissions and we currently would show a screen listing all those permissions granted (except for apps on the known Google packages list). That's why we already have the names of many of these permissions at hand and could display then to the user.

I tried to use the above-mentioned apps that require permissions on my piexl mobile phone, and found that GMS did not pop up a window to prompt the user which authorizations were used.
How to use another implemented API to assign authorize？

DaVinci9196:
> Do they actually grant the gmail.readonly permission or is it maybe silently stripped?

I tried to capture the network request, the request contained gmail.readonly, and the auth content was returned successfully.
image

mar-v-in:
For me it shows a second screen after the initial screen asking to confirm the gmail permission:
Screenshot from 2024-02-27 14-40-07
This screen seems to coming from a web view (not natively) so we should be able to show the very same webview. Maybe it's not showing the screen for you if you signed in to CallApp using that account before.

DaVinci9196:
> This screen seems to coming from a web view (not natively) so we should be able to show the very same webview. Maybe it's not showing the screen for you if you signed in to CallApp using that account before.

OK, I'll try it as a new user. Do we currently have a corresponding way to display the page above?

DaVinci9196:
> > This screen seems to coming from a web view (not natively) so we should be able to show the very same webview. Maybe it's not showing the screen for you if you signed in to CallApp using that account before.
> 
> OK, I'll try it as a new user. Do we currently have a corresponding way to display the page above?

Obtain the authorization URL by changing the request parameters, use webview to load the user authorization page, obtain the authorization result and then request the token. Please see if this method works.
image

PR #2205, opened 2024-02-26
Fix account access for google wifi and google fiber

DaVinci9196:


ale5000-git:
@DaVinci9196

The only apps that need to be in the list are the ones that doesn't request `android.permission.GET_ACCOUNTS` or that request it but limit the sdk version, so in short:
- Not requested => Bad
- Requested with: `<uses-permission android:name="android.permission.GET_ACCOUNTS" android:maxSdkVersion="22" />` => Bad
- Requested with: `<uses-permission android:name="android.permission.GET_ACCOUNTS" />` => Good

You can easily drop the apk on https://sisik.eu/apk-tool and see the manifest.

DaVinci9196:
I followed your instructions, why did fiber still prompt me to log in when I logged in to my Google account?

DaVinci9196:
> Google Fiber can access the account if you grant contacts permission.

I have a question. If you agree to the contact permission, you can view the login account. Why does fiber not apply for the contact permission when opening the page? I see that the pixel phone does not allow contact permissions and can also find the login account.

PR #2214, opened 2024-03-01
Fix the problem that jamboard application cannot be opened

DaVinci9196:
Fix the problem that jamboard application cannot be opened


mar-v-in:
Handled as part of a9906e52beb52856fb8920ba7b55a58e0e636f31

PR #2216, opened 2024-03-02
Location: Implement settings request

DaVinci9196:
For some applications that require GPS function, detect and remind the GPS to check in, such as Google Maps


mar-v-in:
The implemented behavior does not match the documented behavior. According to https://developers.google.com/android/reference/com/google/android/gms/location/SettingsClient, determining the status does never directly result in a dialog to request changes. Instead, when the result does not match the request (the `requests` field contains a location request that would require a location source currently not active), return a Status with code `CommonStatusCodes.RESOLUTION_REQUIRED` and a pending intent to the settings dialog as a resolution.

ale5000-git:
In my opinion it is better to simply show a dismissable notification that the user can choose to click or to just ignore.
So if the user deliberately don't want to grant it then he/she won't be disturbed too much.

mar-v-in:
With the intended flow (using a resolution pending intent) it's up to the calling app if they want to show the dialog to the user or not. I think that's a pretty good solution, as only the calling app knows if it's sensible to disturb the user a lot or not.

mar-v-in:
> Google Maps will not pop up a box to request when the GPS is not turned on, so this function was implemented.

That's because the current implementation on master does not return a resolution pending intent if the location request is not met yet. Returning a resolution pending intent should result in the desired behavior in Google Maps without breaking other apps as your implementation would.

DaVinci9196:
> > Google Maps will not pop up a box to request when the GPS is not turned on, so this function was implemented.
> 
> That's because the current implementation on master does not return a resolution pending intent if the location request is not met yet. Returning a resolution pending intent should result in the desired behavior in Google Maps without breaking other apps as your implementation would.
> 
> Check again the documentation on https://developers.google.com/android/reference/com/google/android/gms/location/SettingsClient

I re-improved the requestLocationSettingsDialog function

PR #2219, opened 2024-03-04
Vending: Add PurchaseActivity dummy

DaVinci9196:
Solve the problem of crash caused by some in-app purchases jumping to the LightPurchaseFlowActivity page through action:com.android.vending.billing.PURCHASE.
e.g. Google Play Books

### DIFF
--- a/vending-app/src/main/AndroidManifest.xml
+++ b/vending-app/src/main/AndroidManifest.xml
@@ -127,5 +127,15 @@
             android:name="com.android.vending.licensing.LicenseServiceNotificationRunnable$SignInReceiver"
             android:exported="false" />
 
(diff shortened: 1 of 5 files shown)


mar-v-in:
Can you name this activity just `org.microg.vending.billing.PurchaseActivity` instead of `com.google.android.finsky.billing.lightpurchase.LightPurchaseFlowActivity`? There is no need to go with the original Google class name here, as it is invoked via the `com.android.vending.billing.PURCHASE` no matter the name.

DaVinci9196:
> Can you name this activity just `org.microg.vending.billing.PurchaseActivity` instead of `com.google.android.finsky.billing.lightpurchase.LightPurchaseFlowActivity`? There is no need to go with the original Google class name here, as it is invoked via the `com.android.vending.billing.PURCHASE` action no matter the name.

okey

PR #2220, opened 2024-03-04
Fix exceptions in CellDetailsSource when the ACCESS_FINE_LOCATION permission isn't granted

### LINKED ISSUE #2213: Regularly appearing "microG Services keeps stopping" window
**Describe the bug**
I keep getting "microG Services keeps stopping" window, even as I have the microG app open and fully working.
Today, I just installed microG on a relatively fresh LineageOS in Magisk using [microG Installer Revived][microg-installer-revived]. Between installing Lineage, Magisk, and microG, I only installed a few apps. 

**To Reproduce**
1. Install LineageOS 21
2. Install Magisk
3. Install [microG Installer Revived][microg-installer-revived]
4. Reboot
5. Try using your phone, you will be getting "microG Services keeps stopping" error regularly.

**Expected behavior**
No "microG Services keeps stopping" window appearing.

**System**
Android Version: 14
Custom ROM: LineageOS 21

**Additional context**
Yesterday, I had the same issue on the previously installed LineageOS 20 (Android 13) after updating the [microG Installer Revived][microg-installer-revived] Magisk module. I don't know which version of microG was used previously. The previous OS also had signature spoofing set up, while the current one doesn't yet.

[microg-installer-revived]: https://github.com/nift4/microg_installer_revived

ale5000-git:
Fix exceptions in CellDetailsSource when the ACCESS_FINE_LOCATION permission isn't granted.

Should fix #2213

### DIFF
--- a/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/cell/CellDetailsSource.kt
+++ b/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/cell/CellDetailsSource.kt
@@ -11,7 +11,10 @@ import android.os.Build.VERSION.SDK_INT
 import android.os.WorkSource
 import android.telephony.CellInfo
 import android.telephony.TelephonyManager
+import android.util.Log
 import androidx.core.content.getSystemService
+
+const val TAG = "CellDetailsSource"
 
 class CellDetailsSource(private val context: Context, private val callback: CellDetailsCallback) {
     fun enable() = Unit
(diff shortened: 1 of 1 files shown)


mar-v-in:
There is no field `allCellInfo`, it's a kotlin property that invokes `getAllCellInfo` on the Java side, so it might as well throw a SecurityException (and does so according to the documentation)

ale5000-git:
We should try telephonyManager.networkOperator in case of SecurityException?

mar-v-in:
networkOperator alone won't give enough information for locating (it merely tells us the country)

ale5000-git:
@mar-v-in
I haven't much experience in Kotlin; I hope I get it correctly now.

mar-v-in (approved):

play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/cell/CellDetailsSource.kt
```diff
@@ -11,8 +11,11 @@ import android.os.Build.VERSION.SDK_INT
 import android.os.WorkSource
 import android.telephony.CellInfo
 import android.telephony.TelephonyManager
+import android.util.Log
 import androidx.core.content.getSystemService
 
+const val TAG = "CellDetailsSource"
```
mar-v-in, line 17: If the log tag only applies to the current file, it should be private.

ale5000-git: Done

PR #2223, opened 2024-03-07
Games: Resolved the issue preventing some games from logging in

DaVinci9196:
Resolved the issue preventing the game from logging in.
e.g. Clash of Clans  &  Brawl Stars

### DIFF
--- a/play-services-basement/src/main/java/org/microg/gms/common/GmsService.java
+++ b/play-services-basement/src/main/java/org/microg/gms/common/GmsService.java
@@ -19,7 +19,7 @@ package org.microg.gms.common;
 public enum GmsService {
     UNKNOWN(-2),
     ANY(-1),
-    GAMES(1, "com.google.android.gms.games.internal.connect.service.START", "com.google.android.gms.games.service.START"),
+    GAMES(1, "com.google.android.gms.games.service.START", "com.google.android.gms.games.internal.connect.service.START"),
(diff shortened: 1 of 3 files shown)


play-services-basement/src/main/java/org/microg/gms/common/GmsService.java
```diff
@@ -19,7 +19,7 @@
 public enum GmsService {
     UNKNOWN(-2),
     ANY(-1),
-    GAMES(1, "com.google.android.gms.games.service.START", "com.google.android.gms.games.internal.connect.service.START"),
+    GAMES(1, "com.google.android.gms.games.internal.connect.service.START", "com.google.android.gms.games.service.START"),
```
mar-v-in, line 22: This would likely break the non-connect service. I guess the issue might be in the service provider. Will investigate

DaVinci9196: Yes, if the GmsService class specifies that a serviceid contains multiple actions, ServiceProvider will take the first action by default, but in fact it is another action that needs to be jumped. For example, in the game service, GamesConnectService has a higher priority than GamesService, but in In the call method of ServiceProvider, GamesService will be used

mar-v-in: I removed that part since it's no longer needed with 64307c1c194d21f5794a5d8789ecdd280b301e7d

PR #2228, opened 2024-03-11
Solve the problem that some Google applications cannot be logged in.

DaVinci9196:
Solve the problem that some Google applications cannot be logged in using accounts.
e.g. Youtube Kids & Youtube & Search Lite & Translate & Earth & Fit & Photos & Videos ...


DaVinci9196:
@mar-v-in I found that some non-Google applications require Google login at the beginning, but there is no corresponding GET_ACCOUNTS permission, and the application will not pop up a window for authorization. This will cause the application to be unable to obtain the logged-in account, and it will be necessary to manually enable permissions in the application settings. Can we add corresponding information to this KNOWN_GOOGLE_PACKAGES Map? Or are there other solutions?

DaVinci9196:
Google Tasks is in a similar situation. Can it also be added to KNOWN_GOOGLE_PACKAGES Map?

mar-v-in:
1d9bf3ffc76a51c2919e4dead46cd81e038567cf adds a new `KNOWN_GOOGLE_APP_CERT_HASHES` for this certificate, as normal apps should not get the privileged permissions. I also added Fitbit (which is owned by Google by now and thus can get special privileges in Play Services) and Google Tasks apps to `KNOWN_GOOGLE_PACKAGES` as a reference.

PR #2230, opened 2024-03-11
GCM: add 4 minute ping interval

dep4:
My operator has 5 minute timeout, so 5 minute here is too long.


CoelacanthusHex:
May it be better to allow input interval value manually?

PR #2232, opened 2024-03-12
Refactor preferences and improvement colors

WSTxda:
- Use iconSpaceReserved to remove preferences icon padding
- Update M3 colors and add missing colors attributes
- Use switchbar in "confirm new apps" and "answer license verification requests" options
- Refactor preferences options


ale5000-git:
Currently there is already a crash on KitKat regarding SwitchBarPreference in some places.

I see you are switching SwitchPreferenceCompat to SwitchBarPreference and I wonder if it won't make things worse.

WSTxda:
> Currently there is already a crash on KitKat regarding SwitchBarPreference in some places.
> 
> I see you are switching SwitchPreferenceCompat to SwitchBarPreference and I wonder if it won't make things worse.

Well, it was changed from `SwitchPreferenceCompat` to `org.microg.gms.ui.SwitchBarPreference` If these two are already displayed in other fragments there should be no problem in KitKat

ale5000-git:
I'm investigating these days but it seems to crash in most fragments.

ale5000-git:
I have found a workaround for the current code: #2236
Although I haven't yet tested it with this PR.

mar-v-in:
> Use iconSpaceReserved to remove preferences icon padding

For some of the preferences where you did this, the space is very much intentional. Example is the "None" entry in app lists.

> Update M3 colors and add missing colors attributes

This also changes colors from the intended color scheme (that is derived from Google's original settings color scheme). Please keep the `gm3_` colors as is and reuse them for declaring the missing theme colors. Also, as you declare the colors, you shouldn't override the fixed colors of the day theme in the night theme, as fixed colors are by definition supposed to be the same in day and night.

> Use switchbar in "confirm new apps" and "answer license verification requests" options

Both of these are not supposed to be switch bar. The switch bar should only be used for the topmost primary switch option on a preferences screen. Both of these settings do not form the primary switch on the respective screens, they are equally important to the other settings on the screen (answer license verification requests doesn't have any other settings on the same screen yet, but that's going to change soon).

> Refactor preferences options

Please avoid reordering or reformatting XML where you don't change the respective element anyway. Reordering and reformatting introduces merge conflicts when benefit is very limited.

WSTxda:
@mar-v-in  Thanks for reply, ok so maybe close this and open another with m3 attributes. By the way, can I work with https://m3.material.io/theme-builder ? it also gives colors with `md_` in files.  

`gm3_`  it is generally used by Google in their commercial version of Material3 or SettingsLib

The rest of the answer is perfectly understood.

PR #2233, opened 2024-03-13
Fix googlemaps not displaying gmscore requesting location permission …

DaVinci9196:
Fixed the problem that the location permission of Google Maps does not pop up when entering gmscore for the first time. Fixed the issue where gmscore’s location permission request would only run once

### DIFF
--- a/play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationManager.kt
+++ b/play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationManager.kt
@@ -47,7 +47,12 @@ class LocationManager(private val context: Context, override val lifecycle: Life
     private val postProcessor by lazy { LocationPostProcessor() }
     private val lastLocationCapsule by lazy { LastLocationCapsule(context) }
     val database by lazy { LocationAppsDatabase(context) }
-    private val requestManager by lazy { LocationRequestManager(context, lifecycle, postProcessor, database) { onRequestManagerUpdated() } }
+    private val requestManager by lazy { LocationRequestManager(context, lifecycle, postProcessor, database) {
+        lifecycleScope.launchWhenStarted {
+            ensurePermissions()
+        }
+        onRequestManagerUpdated()
+    } }
     private val gpsLocationListener by lazy { LocationListenerCompat { updateGpsLocation(it) } }
     private val networkLocationListener by lazy { LocationListenerCompat { updateNetworkLocation(it) } }
     private var boundToSystemNetworkLocation: Boolean = false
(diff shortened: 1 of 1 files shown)


play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationManager.kt
```diff
@@ -277,34 +282,24 @@ class LocationManager(private val context: Context, override val lifecycle: Life
...
-                override fun handleMessage(msg: Message) {
-                    if (msg.what == Activity.RESULT_OK) {
-                        lastLocationCapsule.fetchFromSystem()
-                        onRequestManagerUpdated()
-                        val grantResults = msg.data?.getIntArray(EXTRA_GRANT_RESULTS) ?: IntArray(0)
-                        completable.complete(grantResults.size == permissions.size && grantResults.all { it == PackageManager.PERMISSION_GRANTED })
-                    } else {
-                        completable.complete(false)
-                    }
+        val permissionRequest = activePermissionRequestLock.withLock {CompletableDeferred<Boolean>()}
```
mar-v-in, line 285: This can cause the activity to be invoked multiple times, we only want to start it once

play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationManager.kt
```diff
@@ -47,7 +47,12 @@ class LocationManager(private val context: Context, override val lifecycle: Life
     private val postProcessor by lazy { LocationPostProcessor() }
     private val lastLocationCapsule by lazy { LastLocationCapsule(context) }
     val database by lazy { LocationAppsDatabase(context) }
-    private val requestManager by lazy { LocationRequestManager(context, lifecycle, postProcessor, database) { onRequestManagerUpdated() } }
+    private val requestManager by lazy { LocationRequestManager(context, lifecycle, postProcessor, database) {
+        lifecycleScope.launchWhenStarted {
+            ensurePermissions()
```
mar-v-in, line 52: This doesn't seem to be the right spot for this.

mar-v-in (commented):
Google Maps invokes the location settings implemented via #2216. We can use those settings to also request permissions for microG if necessary, allowing for cleaner integration as calling the location settings is expected by the caller.

play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationManager.kt
DaVinci9196: Do you have any good suggestions, since I haven't found a more suitable place.

PR #2234, opened 2024-03-13
feat(SafetyNet): publicly expose Verify Apps API

rushiiMachine:
Seems like SafetyNet wasn't fully implemented


mar-v-in (changes requested):
Please create the Javadocs based on the official documentation, e.g. [this](https://developers.google.com/android/reference/com/google/android/gms/safetynet/SafetyNetApi.VerifyAppsUserResponse). Google publishes them under CC-BY-4.0 license, so we can use them as long as we attribute Google (which we do in the header of applicable file). For parts of the API that are no longer public (as they were removed or deprecated) you can fetch them archives or you might find them in the javadoc jar published in [Google's Maven repository](https://maven.google.com/web/index.html#com.google.android.gms:play-services-safetynet:18.0.1).

mar-v-in (approved):

PR #2248, opened 2024-03-21
Increase memory to fix build

ale5000-git:
@mar-v-in
I have increased only to 2,5GB to avoid increasing it to much.

It works for now but we should discover why we use so much memory, maybe our gradle files are doing something wrong.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 android.useAndroidX=true
-org.gradle.jvmargs=-Xmx2048m -XX:+UseParallelGC --add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED
+org.gradle.jvmargs=-Xmx2560m -XX:+UseParallelGC --add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED


mar-v-in:
The reason is probably too much parallelization. We now have a lot of build targets (debug/release, core/vending/libs, default/huawei). What we could do to improve:
- reduce parallel workers (GitHub gives us 4 cores so probably currently we use 8 workers). Limiting to 4 workers will significantly reduce memory consumption, but also slow down the build a little bit.
- build less targets e.g. only debug
- disable some optimizations in GitHub builds (or debug in general?) The newly introduced optimizations in the vending app are actually pretty heavy
- intentionally introduce some serialization, e.g first assemble then lint. This will also make the nature of the issue easier to see if the output directly states where the failure is.

ale5000-git:
The GitHub action is mainly to test if everything works, even if we reduce the things built on GitHub then when the user run gradlew build at home it will still fails.

mar-v-in:
If you build at home you offen build one specific target, Mike for testing i typically do `./gradlew :play-services-core:installMapboxDefaultDebug`. This results in less build resource usage. But I see your point (I was merely making suggestions on how to improve the situation with GitHub Actions independent of builds at home)

ale5000-git:
Actually the` intentionally introduce some serialization, e.g first assemble then lint. This will also make the nature of the issue easier to see if the output directly states where the failure is.` could be good for the last part **make the nature of the issue easier to see**.

PR #2250, opened 2024-03-21
Disable some optimizations on debug builds

ale5000-git:
Optimizations are mainly for releases, in debug builds they are less needed; this will reduce the pressure.


ale5000-git:
Although it doesn't fix build completely it reduce pressure a lot and in addition to #2249 works perfectly.

PR #2259, opened 2024-03-25
Add microG-UI-tools translations for Brazilian Portuguese

LucasMZReal:


LucasMZReal:
I think it's properly set up, though I'm a bit confused on whether "pt-rBR" is correct. I see other folders using that name, but I never seen it anywhere else, so ehh should be fine?

LucasMZReal:
Decided to not translate or reference licenses the way I did originally since there are other strings which would look out of place, and it gets the job done; I'd rather not mess with legal wording.

LucasMZReal:
Superseded by https://github.com/microg/GmsCore/pull/2331

Decided to put it in another PR so that there's not as many commits (I suppose usually you can solve that with Git but I don't know how that works so better make this easier for you guys), and also reflect the will of adding any missing translations;

PR #2268, opened 2024-03-28
Auth: Improve method <hasCapabilities>

DaVinci9196:
Solve the problem that some apps cannot use some functions after deleting data in settings.
For example, the search function cannot be used on Google App.

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
+++ b/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
@@ -24,6 +24,7 @@ import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.os.Binder;
 import android.os.Bundle;
 import android.os.Parcel;
 import android.os.RemoteException;
@@ -48,6 +49,7 @@ import org.microg.gms.common.PackageUtils;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
```diff
@@ -231,9 +233,17 @@ public Bundle requestGoogleAccountsAccess(String packageName) throws RemoteExcep
...
-    public int hasCapabilities(HasCapabilitiesRequest request) throws RemoteException {
-        Log.w(TAG, "Not implemented: hasCapabilities(" + request.account + ", " + Arrays.toString(request.capabilities) + ")");
-        return 1;
+    public int hasCapabilities(HasCapabilitiesRequest request) {
+        Log.d(TAG, "hasCapabilities(" + request.account + ", " + Arrays.toString(request.capabilities) + ") is Called");
+        int callingUid = Binder.getCallingUid();
+        String[] packagesForUid = context.getPackageManager().getPackagesForUid(callingUid);
+        if (packagesForUid == null || packagesForUid.length == 0) {
+            Log.w(TAG, "hasCapabilities error <no packages>");
+            throw new SecurityException("Caller isn't signed with recognized keys!");
```
mar-v-in, line 242: Exception text doesn't match actual error.

play-services-core/src/main/java/org/microg/gms/auth/CapabilitiesManager.java
```diff
@@ -0,0 +1,94 @@
...
+                    result = 3;
+                }
+            }
+        }
+        return result;
+    }
+
+    private boolean capabilitiesSyncState(Account account) {
+        enabledCapabilities = getUserDataSet(account, "enabled_capabilities");
+        disabledCapabilities = getUserDataSet(account, "disabled_capabilities");
```
mar-v-in, line 63: We never set those, so this makes little sense.

PR #2273, opened 2024-03-29
Games: Improve login auth scope logic

DaVinci9196:
During the test of PUBG, it was found that the Scopes requested by the application were inconsistent with the actual request, resulting in the failure to authorize the login successfully.
And during this process, it was discovered that during the switching between horizontal and vertical screens, there would be multiple login pages, causing authorization failure.

### DIFF
--- a/play-services-base/src/main/java/com/google/android/gms/auth/api/signin/GoogleSignInOptions.java
+++ b/play-services-base/src/main/java/com/google/android/gms/auth/api/signin/GoogleSignInOptions.java
@@ -234,6 +234,17 @@ public class GoogleSignInOptions extends AutoSafeParcelable {
         }
 
         /**
+         * Specifies OAuth 2.0 scopes your application requests. See {@link Scopes} for more information.
+         *
+         * @param scopes More OAuth 2.0 scopes requested by your app.
+         */
+        @NonNull
+        public Builder requestScopes(@NonNull List<Scope> scopes) {
+            this.scopes.addAll(scopes);
+            return this;
+        }
+
+        /**
          * Specifies that offline access is requested. Requesting offline access requires that the server client ID be specified.
          * <p>
          * You don't need to use {@link #requestIdToken(String)} when you use this option. When your server exchanges the code for tokens, an ID token will be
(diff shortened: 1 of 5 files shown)


play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInService.kt
```diff
@@ -80,7 +80,7 @@ class AuthSignInServiceImpl(
                 val account = account ?: options?.account ?: SignInConfigurationService.getDefaultAccount(context, packageName)
                 if (account != null && options?.isForceCodeForRefreshToken != true && options?.includeUnacceptableScope != true) {
                     if (getOAuthManager(context, packageName, options, account).isPermitted || AuthPrefs.isTrustGooglePermitted(context)) {
-                        val googleSignInAccount = performSignIn(context, packageName, options, account)
+                        val googleSignInAccount = performSignIn(context, packageName, options, account, true)
```
mar-v-in, line 83: silent sign in should not set permitted flag by default (because user did not actively allow sign in)

play-services-base/src/main/java/com/google/android/gms/auth/api/signin/GoogleSignInOptions.java
```diff
@@ -233,6 +233,17 @@ public Builder requestScopes(@NonNull Scope scope, @NonNull Scope... scopes) {
...
+         * Specifies OAuth 2.0 scopes your application requests. See {@link Scopes} for more information.
+         *
+         * @param scopes More OAuth 2.0 scopes requested by your app.
+         */
+        @NonNull
+        public Builder requestScopes(@NonNull List<Scope> scopes) {
+            this.scopes.addAll(scopes);
+            return this;
+        }
+
```
mar-v-in, line 246: This is not part of the public API. If it is really needed, and `@Hide`, but better consider using the existing `requestScopes()` method.

play-services-core/src/main/kotlin/org/microg/gms/games/GamesSignInActivity.kt
```diff
@@ -41,19 +48,22 @@ class GamesSignInActivity : AppCompatActivity() {
 
         window.setGravity(popupGravity)
         startActivityForResult(Intent(this, AuthSignInActivity::class.java).apply {
+            Log.d(TAG, "Sign in for $gamePackageName account:${account?.name} scopes:$scopes")
             putExtra("config", SignInConfiguration().apply {
                 packageName = gamePackageName
+                val optionsBuilder =
+                    GoogleSignInOptions.Builder().requestScopes(scopes).requestProfile().requestId().requestEmail()
```
mar-v-in, line 55: Are these really the default scope for Games Sign In? I don't think most Games need or request access to the user's email address or profile.

PR #2275, opened 2024-03-30
DeviceRegistartionFragment: Allow viewing device registartion data

theimpulson:
## Summary

This PR allows users to view the data shared with the google servers during device registration.

## Test

Ensure registration data is visible


theimpulson:
Need to look into formatting it beautifully but the string is manually generated and isn't a valid JSON. Adding GSON dependency and using that with the actual class object may be better.

theimpulson:
@mar-v-in this is ready from my side. Please review when possible.

TheGamerKing561:
@mar-v-in What's the state of this pull request? Seems like a useful feature.

PR #2284, opened 2024-03-31
Update Polish strings.xml for play-service-base

Patriot99:


mar-v-in:
Please update strings via weblate where appropriate: https://hosted.weblate.org/projects/microg/play-services-base-core-strings/pl/

PR #2304, opened 2024-04-09
IdentitySignInService: beginSignIn method adds judgment logic

DaVinci9196:
It was found that some applications call this method abnormally. 
For example, LinkedIn & Netflix will automatically jump to the Google login page without logging in.

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/identity/IdentitySignInService.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/identity/IdentitySignInService.kt
@@ -5,6 +5,7 @@
 
 package org.microg.gms.auth.credentials.identity
 
+import android.accounts.AccountManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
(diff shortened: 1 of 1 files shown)


play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AssistedSignInActivity.kt
```diff
@@ -0,0 +1,182 @@
...
+        val accounts = accountManager.getAccountsByType(AuthConstants.DEFAULT_ACCOUNT_TYPE)
+
+        if (beginSignInRequest != null) {
+            Log.d(TAG, "beginSignInRequest start")
+            if (accounts.isEmpty()) {
+                // In fact, if you log in through the beginSignIn method without logging in to a Google account,
+                // GMS will report an Operation failed error, resulting in no page feedback when the user clicks.
+                // However, some applications want to use this method to log in to a Google account,
+                // so the package name that needs to be logged in is added here to allow the application to jump to the login page.
+                // e.g. Nova
```
mar-v-in, line 99: So if this doesn't work with original GMS, why should it work with microG?

PR #2313, opened 2024-04-12
Location: Add MG location permission dialog

DaVinci9196:
When Map has GPS and network positioning permissions, but MG does not have location information permissions, when you click on the location, there will be a pop-up window prompting that you do not have GPS permissions. Clicking it will jump to the system's positioning permissions page. It should be MG’s location authorization pop-up window. And found that after permission is granted, Map will not refresh the positioning information.


ale5000-git:
I haven't looked to much in details but how does it behave for device that physically doesn't have GPS?

My opinion:
I think that if the device doesn't have GPS it shouldn't request ACCESS_FINE_LOCATION.
Also if the device DOES have GPS but the app using microG doesn't have ACCESS_FINE_LOCATION granted but just ACCESS_COARSE_LOCATION then microG shouldn't request ACCESS_FINE_LOCATION for itself (user may have denied it for privacy purpose).

DaVinci9196:
> I haven't looked to much in details but how does it behave for devices that physically doesn't have GPS?
> 
> My opinion:
> 
> * I think that if the device doesn't have GPS it shouldn't request ACCESS_FINE_LOCATION.
> * Also if the device DOES have GPS but the app using microG doesn't have ACCESS_FINE_LOCATION granted but just ACCESS_COARSE_LOCATION then microG shouldn't request ACCESS_FINE_LOCATION for itself (user may have denied it for privacy purpose).

I made adjustments to meet one of the conditions

play-services-location/core/src/main/kotlin/org/microg/gms/location/settings/LocationSettingsCheckerActivity.kt
```diff
@@ -141,7 +140,22 @@ class LocationSettingsCheckerActivity : Activity(), DialogInterface.OnCancelList
...
         val improvement = improvements.firstOrNull() ?: return finishResult(RESULT_OK)
+        displayList.add(improvement)
         when (improvement) {
+            Improvement.PERMISSIONS -> {
+                val targetIntent = if (intent.hasExtra(EXTRA_SHOW_MG_SETTINGS) && intent.getBooleanExtra(EXTRA_SHOW_MG_SETTINGS, false)) {
+                    Intent().apply {
+                        action = Settings.ACTION_APPLICATION_DETAILS_SETTINGS
+                        data = Uri.fromParts("package", Constants.GMS_PACKAGE_NAME, null)
+                    }
+                }else{
```
mar-v-in, line 151: Decision to show application details settings should happen in `AskPermissionActivity`, it's not specific to the `LocationSettingsCheckerActivity`.

PR #2321, opened 2024-04-17
fix iap get sku details fail eg:FIFA Soccer

DaVinci9196:


CoelacanthusHex:
I'm not familiar to code of microG, but do you think https://github.com/microg/GmsCore/issues/2254 is related to this PR?

mar-v-in:
@CoelacanthusHex This is unrelated to #2254, which is referring to a feature (fetching licenses from server that were not purchased on the device itself), that is not yet implemented in microG at all.

PR #2336, opened 2024-04-23
permission: remove <com.android.vending.CHECK_LICENSE>

DaVinci9196:
image
image

During the test, it was found that after uninstalling and reinstalling the fakestore, games that have been downloaded and need to be updated cannot enter the FakeLicenseService for execution because there is an exception in the <com.android.vending.CHECK_LICENSE> permission.

The general reason is that after the fakestore is uninstalled and reinstalled, the permission statement is removed and reset. It is not practical to turn fakestore into a system application that does not allow uninstallation. Is it possible to remove the permission here?

### DIFF
--- a/vending-app/src/main/AndroidManifest.xml
+++ b/vending-app/src/main/AndroidManifest.xml
@@ -6,10 +6,6 @@
 <manifest xmlns:tools="http://schemas.android.com/tools"
     xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <permission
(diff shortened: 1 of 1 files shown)


ale5000-git:
You can just reinstall the game that have the problem, weaken permission isn't a good solution.

DaVinci9196:
> You can just reinstall the game that have the problem, weaken permission isn't a good solution.
> 
> You never need to uninstall, you can just use `adb install -r ...` or `pm install -r ...`.

PR #2365, opened 2024-05-07
Map: Remove styles that cannot be set

DaVinci9196:
image

It was found that some applications cannot display map content because they are set to use satellite maps, but the SDK does not support them.

### DIFF
--- a/play-services-maps/core/hms/src/main/kotlin/org/microg/gms/maps/hms/GoogleMap.kt
+++ b/play-services-maps/core/hms/src/main/kotlin/org/microg/gms/maps/hms/GoogleMap.kt
@@ -246,11 +246,10 @@ class GoogleMapImpl(private val context: Context, var options: GoogleMapOptions)
 
     fun applyMapType() {
         // TODO: Serve map styles locally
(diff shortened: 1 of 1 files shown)


ale5000-git:
@DaVinci9196
Looking at the screenshot MAP_TYPE_TERRAIN should be used only on Android 7 or later (I don't think we check but I'm not sure).
Also MAP_TYPE_NONE is uspported by the SDK but not by microG.

mar-v-in:
@ale5000-git there is no version of Huawei device with older than Android 7 that supports microG and this is only for when Huawei Maps is used, so it's fine that we assume Android 7 or higher.

ale5000-git:
@mar-v-in
microG could be used also without signature spoofing with apps compiled with microG libraries.
There is, for example, this: Huawei Nova (Android 6.0)

mar-v-in:
Does it support the required version of HMS SDK?

ale5000-git:
I'm not sure but if there is the note about Android version in the screenshot it means that it exist otherwise they wouldn't have added a useless note.

mar-v-in:
The SDK falls back to regular map on old Android when setting Terrain, so this is fine.

PR #2366, opened 2024-05-09
Fix inverted reportAgain condition

t-m-w:
Resolves accidentally-inverted condition for repeatedly reporting network location. Now, network location is reported repeatedly only if it is recent enough, as intended.

See https://github.com/microg/GmsCore/commit/827a8b914cb3ef970ad81897672f4539a423fdcc#r141781679


t-m-w:
Oops, looks already fixed

PR #2369, opened 2024-05-10
Location: Avoid repeated pop-ups

DaVinci9196:
When you start Google Maps for the first time, it will pop up a location information authorization pop-up window through requestLocationSettingsDialog. 
After clicking Allow only during the trial period, some models will cause repeated pop-ups. Some models will jump to the MG authorization page. 
After returning from the page, the ensurePermissions method in LocationManager will jump to the MG authorization page again.


mar-v-in:
This is handled more cleanly in 8baf6948755948c9625a4d10ae2b45050033abfb

PR #2372, opened 2024-05-11
DroidGuard：Fix class repeated loading exception

DaVinci9196:
Fixed the issue where the entry class in the same the.apk is loaded multiple times, resulting in multiple loading of so, and an exception of repeated loading of so is thrown.


mar-v-in:
With ad2ed4da14cf50aea7b5356c1454eb495df4d3f8 we try harder to unload a class after use, but in case it remains loaded it can be retrieved from a weak map (to ensure we won't have a reference forever)

PR #2375, opened 2024-05-14
Auto-purchase free apps upon checking license

fynngodau:
Adds support for auto-purchasing free apps, and adds preference to enable auto-purchasing. Inspired by the implementation in gplayapi. Fixes https://gitlab.e.foundation/e/backlog/-/issues/7949.

To keep the code readable, I did a major refactor and converted the code to be written in Kotlin and use coroutines (thanks to the now-existing `HttpClient` class). As positive effects, the licensing code is now much shorter, easier to understand due to its clearer logical flow and no longer visibly consists of loads of nested callbacks and recursive calls.

cc @jonathanklee


vending-app/src/main/java/com/android/vending/Util.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+private const val TAG = "FakeStoreUtil"
+
+/**
+ * From [StackOverflow](https://stackoverflow.com/a/46688434/), CC BY-SA 4.0 by Sergey Frolov, adapted.
+ */
+fun ByteArray.encodeGzip(): ByteArray {
```
jonathanklee, line 20: Should remain in LicenseRequestHeaders.kt  as it is not used anyway else for now ? 

Util classes usually end up in tote bag classes and become a mess.

vending-app/src/main/java/com/android/vending/Util.kt
```diff
@@ -0,0 +1,45 @@
...
+                return byteOutput.toByteArray()
+            }
+        }
+    } catch (e: IOException) {
+        Log.e(TAG, "Failed to encode bytes as GZIP")
+        return ByteArray(0)
+    }
+}
+
+suspend fun AccountManager.getAuthToken(account: Account, authTokenType: String, notifyAuthFailure: Boolean) =
```
jonathanklee, line 35: Can be moved to LicenseChecker also here ?

jonathanklee (approved):

mar-v-in (approved):

PR #2385, opened 2024-05-27
Fixed the problem of Uber point selection movement and inability to o…

DaVinci9196:
…btain point selection information


play-services-maps/core/hms/src/main/kotlin/org/microg/gms/maps/hms/GoogleMap.kt
```diff
@@ -309,13 +309,13 @@ class GoogleMapImpl(private val context: Context, var options: GoogleMapOptions)
...
-                Log.w(TAG, e)
-            }
-        }
+//        it.setOnCameraIdleListener {
+//            try {
+//                cameraChangeListener?.onCameraChange(map?.cameraPosition?.toGms())
+//            } catch (e: Exception) {
+//                Log.w(TAG, e)
+//            }
+//        }
```
mar-v-in, line 318: Please don't comment those lines. If they shouldn't be here (which i assume is why you commented them), just remove them entirely.

mar-v-in (changes requested):

PR #2388, opened 2024-05-30
Fixed app crash by DynamiteLoader

DaVinci9196:
The MODULE_VERSION specified in the measurement corresponding ModuleDescriptor is too high, which causes the application to be unable to use the MeasurementService to obtain the corresponding parameter content, which will cause the application to crash.
e.g. <PropertyGuru> <iProperty Malaysia>

### DIFF
--- a/play-services-basement/src/main/java/com/google/android/gms/dynamite/DynamiteModule.java
+++ b/play-services-basement/src/main/java/com/google/android/gms/dynamite/DynamiteModule.java
@@ -15,7 +15,7 @@ import java.lang.reflect.Field;
 import java.util.Objects;
 
 public class DynamiteModule {
-    private static final String TAG = "DynamiteModule";
+    private static final String TAG = "GmsDynamiteModule";
 
     public static final int NONE = 0;
     public static final int LOCAL = -1;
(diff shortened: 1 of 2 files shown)


mar-v-in:
Can you please report which app crashes?

This change is definitely wrong. If anything in our implementation breaks because we don't support a functionality required at the version we report, we should fix that instead of changing the loading behavior.

Your change does effectively override the apps selected policy to PREFER_LOCAL if the apps policy result was to use REMOTE, but you do this on the remote side. Your code only works because our current DynamiteLoader.load implementation is incomplete and can't load remote itself - if it could, your code would be an endless loop. As we obviously want to support that at some point, it makes little sense to merge code that is conceptually wrong and going to break eventually.

DaVinci9196:
e.g. <PropertyGuru> <iProperty Malaysia>

DaVinci9196:
e.g. PropertyGuru , iProperty Malaysia

mar-v-in:
Something similar as been part of 0e4135d (it returns a randomly generated ID instead, that has the same structure that could have been returned from this method in Google's implementation as well)

PR #2396, opened 2024-06-06
Updated IT translations (in sync with weblate)

pietrodevo:


mar-v-in:
Can you please add those strings via weblate where appropriate? This makes maintaining translations much easier as there is no merge conflicts.

ale5000-git:
@pietrodevo
I have verified that all the strings are already translated on weblate so there is nothing left to do.
We still thank you for you efforts.

PR #2403, opened 2024-06-12
AD: Changing the initial configuration

DaVinci9196:
During the test, we found that some apps rely on AD_ID to enter the app, but the default option of adTrackingLimitedGlobally is true, which returns EMPTY_AD_ID to the app. This also makes these apps unable to enter the home page normally.
e.g. HKTVmall Lite – Online Shoppin <com.hktv.android.hktvmalllite>

### DIFF
--- a/play-services-ads-identifier/core/src/main/kotlin/org/microg/gms/ads/identifier/AdvertisingIdService.kt
+++ b/play-services-ads-identifier/core/src/main/kotlin/org/microg/gms/ads/identifier/AdvertisingIdService.kt
@@ -29,7 +29,7 @@ class AdvertisingIdService : Service() {
 
 class MemoryAdvertisingIdConfiguration(context: Context) : AdvertisingIdConfiguration(context) {
     override val adTrackingLimitedPerApp: MutableMap<Int, Boolean> = hashMapOf()
-    override var adTrackingLimitedGlobally: Boolean = true
+    override var adTrackingLimitedGlobally: Boolean = false
     override var debugLogging: Boolean = false
     override var adId: String = EMPTY_AD_ID
     override var debugAdId: String = EMPTY_AD_ID


mar-v-in:
Our behavior matches Google's description for when ad tracking id is removed by the user. This is a setting users can do on Google Settings and not handling it correctly is a misbehavior of the respective app.

The app you mentioned crashes for me in the emulator, so I can't verify how it would behave with original GMS and the corresponding setting set. Do you know other apps with the same issue?

mar-v-in:
I checked the documentation of said app and it asks users to enabled ad is in settings when it's disabled, as it fails to work correctly if ad id is disabled. https://hktv.my.site.com/cs/s/article/HKTVmall-Lite-Frequently-Asked-Questions?language=zh_TW
This is a violation of Google Ad Id terms, which clearly states that the app must respect user's configuration https://support.google.com/googleplay/android-developer/answer/9857753#ad-id

DaVinci9196:
> microG behavior matches Google's description for when ad tracking id is removed by the user. This is a setting users can do on Google Settings and not handling it correctly is a misbehavior of the respective app.
> 
> The app you mentioned crashes for me in the emulator, so I can't verify how it would behave with original GMS and the corresponding setting set. Do you know other apps with the same issue?

image
Currently, only this application has been found to have this situation.

DaVinci9196:
> I checked the documentation of said app and it asks users to enable ad id in settings when it's disabled, as it fails to work correctly if ad id is disabled. https://hktv.my.site.com/cs/s/article/HKTVmall-Lite-Frequently-Asked-Questions?language=zh_TW This is a violation of Google Ad Id terms, which clearly states that the app must respect user's configuration https://support.google.com/googleplay/android-developer/answer/9857753#ad-id

In this case, how should we handle it in compliance?

mar-v-in:
You can file a policy violation report at https://support.google.com/googleplay/android-developer/contact/policy_violation_report so Google takes care of it.

DaVinci9196:
> You can file a policy violation report at https://support.google.com/googleplay/android-developer/contact/policy_violation_report so Google takes care of it.

Haha, I'll report it.

DaVinci9196:
@mar-v-in can we explain the advertising service related content on the settings page like GMS does, and provide users with operations to generate, reset, and delete AD_ID?

ale5000-git:
@mar-v-in
I have a suggestion, AppOps are used to negate the permissions but mayke apps think that they are granted so we can leverage its use by doing this:
- If the AppOps is negated for com.google.android.gms.permission.AD_ID then return a random AD_ID that looks like real but instead it change every time (this respect privacy and make bad apps working).
- If it isn't negated then follow the "normal" rules.

What do you think?
It shouldn't be too hard.

mar-v-in:
@ale5000-git 
This is not about the AD_ID permission. If the permission is not granted on Android versions that require it, an exception will be thrown.

This is about Ad ID disabled/deleted. microG behaves as if the user decided to delete his Ad ID via Google settings as described at https://support.google.com/googleplay/android-developer/answer/6048248?hl=en#zippy=%2Cdelete-your-devices-advertising-id . @DaVinci9196's suggestion is to add the Ads settings to microG settings, so users can change microG's behavior and create an Ad ID, so that apps that require it can still work.

I don't think there is any valid purpose for this, no user is interested in being tracked through Ad ID and apps not supporting that setting is in clear violation of the terms of this ID and thus them not working to me is perfectly fine. Apps that require you to enable tracking for ad purposes to use them are also not only in violation of Google's terms, but also in violation of privacy laws in many countries.

If we really want to support this usecase, we could generate a random ad id for each app and reset it regularly (something you can also do in Google's settings). This would prevent cross-app tracking as well as longterm tracking, however I'm not sure if this will work for said app or if it would start to fail if you reset the ad id.

ale5000-git:
There are a lot of functions and I haven't tried any yet but it may be something like this:
```
public int unsafeCheckOpNoThrow ([String](https://developer.android.com/reference/java/lang/String) op, 
                int uid, 
                [String](https://developer.android.com/reference/java/lang/String) packageName)
```

We can check an op for the alling app.

mar-v-in:
I don't know of any way to introduce custom app ops that work reliably across multiple android versions. It's certainly nothing we should be doing. If it's just about storing a custom permission per app that those apps don't know about, we can already do this independent of the app ops system. This would also work properly when unprivileged (unprivileged apps can't access app ops on some android versions).

ale5000-git:
What I mean is to allow this "workaround" per-app and not globally, so it won't be used by default.
App op idea was simpler since it doesn't require coding new UI but if it can be done independently it is better.

DaVinci9196:
> Configuring app ops via adb might be sufficient for you, it's not really suited for the majority of users - adb itself is not meant for users.
> 
> My point remains: Maintaining such a feature (if globally or per app) is extra work and it seems pretty unnecessary if it should never be required to turn ad id on.

It is still necessary to add control options for ads, which will help us improve the overall advertising content later. If possible, please allow us to do so.

Wojtaz0w:
As microG user, I do not want to see ads. This is THE reason to use microG because there are no in-app ads. Please do not add support for them.

PR #2184, opened 2024-02-19
Add regular fido service

p1gp1g:
It adds support for regular FIDO service.

Edit: This PR is not based on the other PR (#2183) any more. Old branch can be found here: p1gp1g:feat/regular_fido_v1

PR #2409, opened 2024-06-18
fido2: Add FIDO2_REGULAR dummy

DaVinci9196:
Prevent application crashes. e.g.  ManulifeMOVE <com.manulife.move>


ale5000-git:
There is already a more complete implementation here: #2184
I haven't tested it, but I have pointed it out in case you didn't notice.

mar-v-in:
Implemented through #2184

PR #2415, opened 2024-06-21
Supplement Google Password Manager functionality

DaVinci9196:
During the testing of the latest version of Google Chrome, clicking the password management tool in the settings did not respond. The CREDENTIAL_MANAGER service method needs to be improved. 
e.g. Google Chrome <126.0.6478.110>

### DIFF
--- a/play-services-core/src/main/kotlin/com/google/android/gms/credential/manager/PasswordManagerActivity.kt
+++ b/play-services-core/src/main/kotlin/com/google/android/gms/credential/manager/PasswordManagerActivity.kt
@@ -1,0 +1,99 @@
+package com.google.android.gms.credential.manager
+
+import android.accounts.AccountManager
+import android.os.Bundle
+import android.util.Log
+import android.view.View
+import android.webkit.WebView
+import android.widget.ProgressBar
+import android.widget.RelativeLayout
+import androidx.appcompat.app.AppCompatActivity
+import org.microg.gms.accountsettings.ui.WebViewHelper
+import org.microg.gms.auth.AuthConstants
+
+const val PASSWORD_MANAGER_CLASS_NAME = "com.google.android.gms.credential.manager.PasswordManagerActivity"
+
+const val EXTRA_KEY_ACCOUNT_NAME = "pwm.DataFieldNames.accountName"
+const val EXTRA_KEY_UTM_SOURCE = "utm_source"
+const val EXTRA_KEY_UTM_MEDIUM = "utm_medium"
+const val EXTRA_KEY_UTM_CAMPAIGN = "utm_campaign"
+const val EXTRA_KEY_UTM_CONTENT = "utm_content"
+
+private const val TAG = "PasswordManagerActivity"
+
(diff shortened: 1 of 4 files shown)


play-services-core/src/main/kotlin/com/google/android/gms/credential/manager/PasswordManagerActivity.kt
```diff
@@ -0,0 +1,99 @@
...
+                append("$EXTRA_KEY_UTM_MEDIUM=$utmMedium")
+            }
+            if (!utmCampaign.isNullOrEmpty()) {
+                if (isEmpty()) append("?") else append("&")
+                append("$EXTRA_KEY_UTM_CAMPAIGN=$utmCampaign")
+            }
+            if (!utmContent.isNullOrEmpty()) {
+                if (isEmpty()) append("?") else append("&")
+                append("$EXTRA_KEY_UTM_CONTENT=$utmContent")
+            }
```
mar-v-in, line 84: If we invoke the URL with different or no UTM vales, will it change the displayed information? If no, we can completely remove them, improving user privacy and reducing code complexity.

play-services-core/src/main/kotlin/org/microg/gms/credential/CredentialManagerService.kt
```diff
@@ -53,13 +61,27 @@ class CredentialManagerService : BaseService(TAG, GmsService.CREDENTIAL_MANAGER)
...
+                    addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                    if (params.caller != null) {
+                        putExtra(EXTRA_KEY_UTM_SOURCE, params.caller.source)
+                        putExtra(EXTRA_KEY_UTM_MEDIUM, params.caller.medium)
+                        putExtra(EXTRA_KEY_UTM_CAMPAIGN, params.caller.campaign)
+                        putExtra(EXTRA_KEY_UTM_CONTENT, params.caller.content)
+                    }
+                }
+                val flags = PendingIntent.FLAG_ONE_SHOT or PendingIntent.FLAG_UPDATE_CURRENT
+                val pendingIntent = PendingIntent.getActivity(context, 0, intent, flags)
```
mar-v-in, line 80: If you use `FLAG_UPDATE_CURRENT` with a static `requestCode` (0), any other app calling this method will override the pending intent. e.g. if app A invokes this with account A and stores the returned pending intent for a brief while, than app B invokes this with account B, then app A invokes its stored pending intent, it will open account B.

play-services-core/src/main/kotlin/com/google/android/gms/credential/manager/PasswordManagerActivity.kt
ale5000-git: In general these values are just used by Google Analytics inside a website but I don't know about this specific site.

DaVinci9196: I tested it and removing these parameters had no effect

PR #2420, opened 2024-06-25
Improve the IAP interface

DaVinci9196:
Improved the IAP interface and fixed some payment issues, such as: Strava: Run, Bike, Hike, SoundCloud: Play Music & Songs

### DIFF
--- a/vending-app/src/main/aidl/com/android/vending/billing/IInAppBillingIsAlternativeBillingOnlyAvailableCallback.aidl
+++ b/vending-app/src/main/aidl/com/android/vending/billing/IInAppBillingIsAlternativeBillingOnlyAvailableCallback.aidl
@@ -6,7 +6,11 @@ package com.android.vending.billing;
 
 import android.os.Bundle;
 
-
 interface IInAppBillingIsAlternativeBillingOnlyAvailableCallback {
+    /**
+     * @param bundle a Bundle with the following keys:
+     *        "RESPONSE_CODE" - Integer
+     *        "DEBUG_MESSAGE" - String
+     */
     void callback(in Bundle bundle);
 }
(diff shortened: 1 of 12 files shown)


vending-app/src/main/java/org/microg/vending/billing/core/HttpClient.kt
```diff
@@ -79,7 +79,7 @@ class HttpClient(context: Context) {
             override fun getHeaders(): Map<String, String> = headers
             override fun getBody(): ByteArray = payload.encode()
             override fun getBodyContentType(): String = "application/x-protobuf"
-        }.setShouldCache(cache).setRetryPolicy(DefaultRetryPolicy(POST_TIMEOUT, 0, 0.0F)))
+        }.setShouldCache(cache).setRetryPolicy(DefaultRetryPolicy(POST_TIMEOUT, 3, 0.0F)))
```
mar-v-in, line 82: This is likely to break things. If the first request goes through but timeouts at the reply, a retry is going to fail. Retries can only succeed if they happen because the timeout happens before sending the request (e.g. at DNS or before TLS termination).

vending-app/src/main/aidl/com/android/vending/billing/IInAppBillingService.aidl
```diff
@@ -367,4 +377,23 @@ interface IInAppBillingService {
 //    void showInAppMessages(int apiVersion, String packageName, in Bundle extraParams, IInAppBillingServiceCallback callback) = 1200;
 
 //    void r(String packageName, Bundle arg2, eql arg3) = 1300;
+
+    void showInAppMessages(int apiVersion, String packageName, in Bundle extraParams, IInAppBillingServiceCallback callback) = 1200;
```
mar-v-in, line 381: `showInAppMessages` is now twice.

vending-app/src/main/aidl/com/android/vending/billing/IInAppBillingIsAlternativeBillingOnlyAvailableCallback.aidl
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.android.vending.billing;
+
+import android.os.Bundle;
+
+// Declare any non-default types here with import statements
+
+interface IInAppBillingIsAlternativeBillingOnlyAvailableCallback {
+    /**
```
mar-v-in, line 13: We don't need these instruction comments in our AIDL files.
Also the first line with the file name is very unnecessary.

mar-v-in (changes requested):

mar-v-in (approved):

Issue #2424, opened 2024-06-30
`AssistedSignInFragment` lacks error handling


fynngodau:
A lack of error handling in `AssistedSignInFragment` can cause apps that try signing in with Google to crash as a consequence of microG crashing.

To reproduce:

* Sign in to Google account
* Turn off network
* Open [booking.com app](https://play.google.com/store/apps/details?id=com.booking)

The app will crash because [`AuthManager.requestAuth`](https://github.com/microg/GmsCore/blob/991da7bdd172add5eb664a4d713ca53513d5d5f3/play-services-core/src/main/java/org/microg/gms/auth/AuthManager.java#L250) throws an `IOException` (like is expected), which is uncaught in the chain:

* [`signin/extensions checkAppAuthStatus`](https://github.com/microg/GmsCore/blob/991da7bdd172add5eb664a4d713ca53513d5d5f3/play-services-core/src/main/kotlin/org/microg/gms/auth/signin/extensions.kt#L97-L101)
* [`AssistedSignInFragment`](https://github.com/microg/GmsCore/blob/991da7bdd172add5eb664a4d713ca53513d5d5f3/play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AssistedSignInFragment.kt#L86)

PR #2446, opened 2024-07-15
Fixed the issue that caused quick login to crash 

DaVinci9196:
Fixed the issue that caused quick login to crash due to UNREGISTERED_ON_API_CONSOLE.
e.g. DeutschlandCard (de.deutschlandcard.app)


fynngodau:
> Fixed the issue that caused quick login to crash due to UNREGISTERED_ON_API_CONSOLE

Does it fix other causes as well, i.e. https://github.com/microg/GmsCore/issues/2424, or only exceptions thrown in some of the cases?

mar-v-in (changes requested):
This only fixes the crash in the case where the error happens:
1. with the first account
2. at first attempt to connect

It will still crash if a connection issue happens later. The better way to fix this is to catch the exceptions when they happen in AssistedSignInFragment instead of doing some useless additional network request early to try to detect issues.

DaVinci9196:
Processed

PR #2448, opened 2024-07-16
Microglg_github

Bodiepoundsign:
Microsoft and play code generator


Bodiepoundsign (commented):
1st time using program software. Pls be gentle

mar-v-in:
We don't use VSCode here and also don't have editor specific configuration files.

PR #2453, opened 2024-07-21
Add Turkish (TR) strings

ysfchn:
This PR adds Turkish (language code is `tr`) strings for all packages from what I could find, except `play-services-nearby` (I had no time to also translate that but plan to do that later.)

* `play-services-ads-identifier`
* `play-services-auth-api-phone`
* `play-services-base`
* `play-services-core`
* `play-services-droidguard`
* `play-services-fido`
* `play-services-location`
* `play-services-oss-licenses`
* `vending-app`

I've kept copyright notices as-is, and I noticed few of strings a wrapped with `"` (quotation) character, I assume it is written to escape `'` (apostrophe), so I added backslash in front of each `'` to avoid wrapping the whole string with `"` instead to make it consistent with other strings. I'm not sure if I have to do something else to make microG display in Turkish but after taking a look at Android documentation, I guess adding a new folder `values-tr` is enough.

Thanks for making microG by the way, I cannot express in words how much it has helped me.

### DIFF
--- a/play-services-ads-identifier/core/src/main/res/values-tr/strings.xml
+++ b/play-services-ads-identifier/core/src/main/res/values-tr/strings.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ SPDX-FileCopyrightText: 2023 microG Project Team
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+
+<resources>
+    <string name="perm_ad_id_label">Reklam kimliği izni</string>
+    <string name="perm_ad_id_description">Bir yayıncı uygulamasının geçerli bir reklam kimliğine doğrudan veya dolaylı olarak erişmesine izin verir.</string>
+    <string name="perm_ad_id_notification_label">Reklam kimliği bildirimi</string>
+    <string name="perm_ad_id_notification_description">Bir uygulamanın, kullanıcının reklam kimliği veya reklam izleme tercihi güncellendiğinde bir bildirim almasını sağlar.</string>
+</resources>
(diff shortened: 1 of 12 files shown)


lucasmz-dev:
You may wanna try and import this to Weblate
https://hosted.weblate.org/projects/microg/#languages
It causes less issues that way with merging and so, though to be fair it seems turkish does not have any files so I'm not sure if there would be a merge conflict there.

Well, anyhow it is nice to have since you can set up notifications for new strings and such so even after this it would be nice

lucasmz-dev:
https://github.com/microg/GmsCore/tree/master/play-services-core/microg-ui-tools/src/main/res

This one is significant and you may wanna do, it isn't on Weblate and you have to open a PR for it atm, so might as well take advantage

ysfchn:
> It causes less issues that way with merging and so, though to be fair it seems turkish does not have any files so I don't think there would be a merge conflict here, but after this is merged it should show up just fine there as well.

I didn't know there was a Weblate project for MicroG until I created this PR, but it seems that Weblate doesn't include a Turkish language yet, so even if I wanted to, it is not possible to contribute to Weblate for Turkish.

> This one is significant and you may wanna do, it isn't on Weblate and you have to open a PR for it atm, so might as well take advantage

Huh, I somehow missed that file, thanks, will push a new commit for it soon.

lucasmz-dev:
> I didn't know there was a Weblate project for MicroG until I created this PR, but it seems that Weblate doesn't include a Turkish language yet, so even if I wanted to, it is not possible to contribute on Weblate for Turkish.

Yeah, it isn't linked anywhere yet except for commits 😵‍💫
If you wanted to add it, you just have to have an account, then add Turkish as one of the languages you can translate to, then you should be able to add that language.

Anyway, good job :P

ysfchn:
I somehow messed up the commit history while pulling the latest commits, everything should be fine now.

lucasmz-dev:
@mar-v-in just ui tools now

PR #2462, opened 2024-07-30
Add Miscellaneous setting and Hide app icon option

### LINKED ISSUE #2263: feat(GmsCore): `Hide App Icon from Drawer/Launcher`
**Is your feature request related to a problem? Please describe.**
The current version of GmsCore lacks an option within the app to hide the app icon from the app drawer/launcher. 

**Describe the solution you'd like**
A toggle within the app to hide the app icon from the app drawer/launcher would be very much appreciated if implemented.

**Additional context**
GmsCore, once signed-in, is an app most simple users won't look back to and it's almost like a system app, which is very much needed for functionality but need not be present in the app drawer among other frequently used apps, as in opened every single time for any app usage. System apps are essential for background functionality but rarely interacted with directly. Hiding the GmsCore icon declutters the app drawer, keeping it focused on frequently used apps and providing a cleaner user experience.

### LINKED ISSUE #2266: Add hide launcher button
Thank you for the time toy spent on the project. I was wondering if you could create a hide from phone launcher button like the previous microg. I don't use the app and use the phone from stock so being able to hide it from the app drawer permanently would be appreciated like the older version of microg. Thanks!

### LINKED ISSUE #2361: Add option to "Hide app icon from launcher"
**Space consuming icon on my Home screen**
I'm a minimalist and seeing the MicroG icon clutter up my drawer space is extremely annoying.

**Option to remove/hide the icon from launcher**
I humbly request you guys to add an option to remove/hide the icon from launcher as was in older versions.

ILoveOpenSourceApplications:
Adds an option to hide icon from launcher.

Resolves #2263, resolves #2266 as well as resolves #2361.

Credits: @rufusin


ale5000-git:
The icon should be enabled by default.

These two should seems to do the opposite:
`CheckIn.HIDE_APP_ICON -> getSettingsBoolean(key, true)`
`android:enabled="false"`

ale5000-git:
Also in my opinion the variable name seems a bit strange:
`private fun updateHideAppIcon(enabled: Boolean)`

Personally I would choose
`private fun updateHideAppIcon(hidden: Boolean)`
or
`private fun updateShowAppIcon(enabled: Boolean)`

ILoveOpenSourceApplications:
> The icon should be enabled by default.

```java
android:defaultValue="false"/>
``` 
Isn't the icon enabled by default because of the above line in preferences_start.xml or do I have to change the value somewhere else as well?

> These two should seems to do the opposite: `CheckIn.HIDE_APP_ICON -> getSettingsBoolean(key, true)` `android:enabled="false"`

I don't understand. Could you elaborate?

ILoveOpenSourceApplications:
> Also in my opinion the variable name seems a bit strange: `private fun updateHideAppIcon(enabled: Boolean)`

May I ask what the variable does here so that I could understand which one from the below two would make sense?
> Personally I would choose `private fun updateHideAppIcon(hide: Boolean)` or `private fun updateShowAppIcon(enabled: Boolean)`

ale5000-git:
> > The icon should be enabled by default.
> 
> ```java
> android:defaultValue="false"/>
> ```

The option in microG settings is disabled so the icon should be enabled by default, but you disable the icon in reality here:
https://github.com/microg/GmsCore/pull/2462/files#diff-d629a5125b2f479f4916b4b963c3ed5bc5b8c8e79951dfb805bc1fa4a7384624R597-R605

play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
```diff
@@ -117,6 +117,7 @@ class SettingsProvider : ContentProvider() {
             CheckIn.SECURITY_TOKEN -> checkInPrefs.getLong(key, 0)
             CheckIn.VERSION_INFO -> checkInPrefs.getString(key, "") ?: ""
             CheckIn.DEVICE_DATA_VERSION_INFO -> checkInPrefs.getString(key, "") ?: ""
+            CheckIn.HIDE_APP_ICON -> getSettingsBoolean(key, true)
```
ale5000-git, line 120: Here you specify that the default value is true.

ILoveOpenSourceApplications:
> This shouldn't be touched => org.microg.gms.ui.MainSettingsActivity This can be disabled (but not by default) => org.microg.gms.ui.SettingsActivity

So the required change is to change `MainSettingsActivity` to `SettingsActivity`?

play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
ILoveOpenSourceApplications: So what's the requested change here?

ale5000-git: Using `false` will make the default not hidden:
`CheckIn.HIDE_APP_ICON -> getSettingsBoolean(key, false)`

ILoveOpenSourceApplications: Done.

ILoveOpenSourceApplications:
>instead with this:
updateHideAppIcon(hide=true)
when hide = true, the icon is hidden so not enabled.

Changed `enabled` to `hide`.

ILoveOpenSourceApplications:
@mar-v-in, please review whenever you have time in your hand.

ILoveOpenSourceApplications:
@mar-v-in & @ale5000-git, a gentle review ping for review.

ILoveOpenSourceApplications:
@ale5000-git, I believe I have addressed all the requested changes in this pull request. Please let me know if there are any additional modifications or suggestions.
@mar-v-in, your input would also be greatly appreciated.
Thank you for your time and consideration.

play-services-core/src/main/AndroidManifest.xml
```diff
@@ -629,7 +629,9 @@
             android:roundIcon="@mipmap/ic_app_settings"
             android:process=":ui"
             android:targetActivity="org.microg.gms.ui.MainSettingsActivity"
-            android:taskAffinity="org.microg.gms.settings">
+            android:taskAffinity="org.microg.gms.settings"
+            android:enabled="false"
```
ale5000-git, line 633: @ILoveOpenSourceApplications
This is wrong.

play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
```diff
@@ -75,6 +76,7 @@
 import static android.view.View.VISIBLE;
 import static android.view.inputmethod.InputMethodManager.SHOW_IMPLICIT;
 import static org.microg.gms.auth.AuthPrefs.isAuthVisible;
+import static org.microg.gms.checkin.CheckinPreferences.hideAppIcon;
```
mar-v-in, line 79: You add new imports without any need. Please remove them again

play-services-core/src/main/res/xml/preferences_start.xml
```diff
@@ -65,9 +65,15 @@
             android:title="@string/service_name_location"/>
     </PreferenceCategory>
     <PreferenceCategory android:layout="@layout/preference_category_no_label" android:key="prefcat_footer">
+        <SwitchPreferenceCompat
+            android:defaultValue="false"
+            android:icon="@drawable/ic_hide_app_icon"
+            android:key="hideAppIcon"
+            android:summary="@string/pref_hide_app_icon_summary"
+            android:title="@string/pref_hide_app_icon_title" />
```
mar-v-in, line 73: Preferences should not be directly modified from the `:ui` process as shared preferences are not safe for multi-process use. That's why we generally use the `SettingsProvider` to store settings.

However, in this case, the settings provider is not needed, as the ground truth of the component enabled state is in the PackageManager. With your code, if someone enables or disables the component without using this setting (e.g. by directly invoking `pm disable` via `adb`), the setting would no longer match the actual live state.

I suggest you make the setting non-persistent (by settings `android:persistent="false"`) and configure it from code by taking the value from `getComponentEnabledSetting` when realizing the setting.

play-services-core/src/main/res/xml/preferences_start.xml
```diff
@@ -65,9 +65,15 @@
             android:title="@string/service_name_location"/>
     </PreferenceCategory>
     <PreferenceCategory android:layout="@layout/preference_category_no_label" android:key="prefcat_footer">
+        <SwitchPreferenceCompat
```
mar-v-in, line 68: I don't think this setting is so important that it deserves a prime location at the top level settings dialog. I suggest to create a new section for these kind of things, although I don't know how to name it or what else would belong there.

play-services-core/src/main/res/xml/preferences_start.xml
```diff
@@ -65,9 +65,15 @@
...
+            android:defaultValue="false"
+            android:icon="@drawable/ic_hide_app_icon"
+            android:key="hideAppIcon"
```
mar-v-in, line 71: This doesn't follow our naming convention for preference keys (`pref_<snake_case_name>`).

play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsContract.kt
```diff
@@ -34,6 +34,7 @@ object SettingsContract {
         const val SECURITY_TOKEN = "securityToken"
         const val VERSION_INFO = "versionInfo"
         const val DEVICE_DATA_VERSION_INFO = "deviceDataVersionInfo"
+        const val HIDE_APP_ICON = "hideAppIcon"
```
mar-v-in, line 37: This does not belong into the Checkin settings contract. In fact it's not needed at all in `SettingsContract` or `SettingsProvider`, see my other comment

mar-v-in (changes requested):
See my comments on the code. Also, please add a way to hide this setting on the Huawei build flavor, as devices with Huawei OS that support microG, the microG settings is automatically hidden from launcher.

play-services-core/src/main/res/xml/preferences_start.xml
ale5000-git: @mar-v-in
Some ideas:
1) Other settings
2) Minor settings
3) Trivial settings

ale5000-git:
@mar-v-in
I think it would also be nice to be able to optionally set the default status in `/system/etc/microg.xml`, so ROM authors and flashable ZIPs can set it without the need to mess with it in other ways.

play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
ILoveOpenSourceApplications: Even though this PR made additions in 
```
play-services-core/src/main/kotlin/org/microg/gms/checkin/CheckinPreferences.kt
``` 
and 

```
play-services-base/core/src/main/kotlin/org/microg/gms/ui/Utils.kt
```

play-services-core/src/main/res/xml/preferences_start.xml
ILoveOpenSourceApplications: I followed a similar name scheme but changed it in https://github.com/microg/GmsCore/pull/2462/commits/3094aff1c6c6142b612a16ba79936e5849100db2. Reverted that commit.

play-services-core/src/main/res/xml/preferences_start.xml
ILoveOpenSourceApplications: What about "Miscellaneous"?

ILoveOpenSourceApplications:
I have made changes according to the requests, not sure if it is what was intended. Do let me know if further changes are needed.

ILoveOpenSourceApplications:
> @mar-v-in I think it would also be nice to be able to optionally set the default status in `/system/etc/microg.xml`, so ROM authors and flashable ZIPs can set it without the need to mess with it in other ways.

Is this within the scope of this PR or just a suggestion for the future of microg?

ale5000-git:
> Is this within the scope of this PR or just a suggestion for the future of microg?

It can also be done later in a different PR.

ale5000-git:
> Also, please add a way to hide this setting on the Huawei build flavor, as devices with Huawei OS that support microG, the microG settings is automatically hidden from launcher.

@ILoveOpenSourceApplications
You misunderstood this, this should happen on Huawei build **flavor**, it is a flavor of microG; it shouldn't check the device manufacturer.

ale5000-git:
Also, although I haven't tried, the changes to AndroidManifest.xml seems totally unnecessary.

ILoveOpenSourceApplications:
> @ILoveOpenSourceApplications You misunderstood this, this should happen on Huawei build **flavor**, it is a flavor of microG; it shouldn't check the device manufacturer.

Can you point me as to where I need to make the changes?

ILoveOpenSourceApplications:
This isn't something I hoped would require these many changes. As it does, I'm not the one capable of doing so and hope someone else in the future may implement such a feature. And in that hope, I'm closing down this PR.

ale5000-git:
@ILoveOpenSourceApplications
I'm busy with other things now but I will look at it when I get more free time

PR #2463, opened 2024-07-31
FIDO: Credential selection, PIN authentication, screenlock authentication fixes

TheMartinizer:
It was near impossible to start working on credential selection without touching code that had already been touched in my previous pull requests, so this pull request contains all the code of pull requests #2194 and #2199, and builds on top of that. Sorry I couldn't find a better way to do this.

If there are several available credentials when trying to sign in to a web page and the allowList is empty, the user is now brought to a page that displays a list of available credentials. The credential chosen from the list is used to log in to the page. If the credentials are taken from the Screen Lock credential store, each credential in the list also has a button to delete the credential (I didn't find another, better place to put this functionality).

Each time a user tries to sign in or register a new key using the ScreenLock mechanism, the key store is cleaned up by deleting invalidated keys. Keys can become invalidated for instance if the user adds new fingerprints to their biometric reader. User info (including display name and icon) is tied to each credential key using an SqLite database in ScreenLockCredentialStore.kt. Each time the database is accessed, it is pruned by removing entries that refer to keys that no longer exist, for instance if they were deleted or invalidated.

Since Android keys require biometric authentication to be unlocked, and we don't know which credential to unlock until the user has selected it, the AuthenticatorActivity no longer receives the credentials themselves, but instead an asynchronous function to get the credential. In case the credential is already available (such as for credentials from an NFC or Bluetooth key), this function just returns the credential directly. In the case of Screen lock credentials, this function prompts the user for biometric authentication and uses this to unlock the key from the keystore, and then returns the credential.

I have tested this with NFC and Screenlock credentials on my own phone. I don't have access to a Bluetooth key, so this is still untested.

At the moment, the list of available credentials from the Screenlock method is shown before the user has supplied any biometric authentication. If you think this is a security risk, I could change it so the user has to provide authentication before being shown the list, but in that case I think the user would still have to give biometric again later to unlock the chosen credential.

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/auth/login/FidoHandler.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/auth/login/FidoHandler.kt
@@ -14,6 +14,7 @@ import com.google.android.gms.fido.fido2.api.common.*
 import kotlinx.coroutines.CancellationException
 import org.json.JSONArray
 import org.json.JSONObject
+import org.microg.gms.fido.core.AuthenticatorResponseWrapper
 import org.microg.gms.fido.core.RequestHandlingException
 import org.microg.gms.fido.core.transport.Transport
 import org.microg.gms.fido.core.transport.TransportHandlerCallback
@@ -68,9 +69,10 @@ class FidoHandler(private val activity: LoginActivity) : TransportHandlerCallbac
         })
     }
 
-    private fun sendSuccessResult(response: AuthenticatorResponse, transport: Transport) {
-        Log.d(TAG, "Finish with success response: $response")
+    private suspend fun sendSuccessResult(responseWrapper: AuthenticatorResponseWrapper, transport: Transport) {
+        val response = responseWrapper.responseChoices.get(0).second.invoke()
         if (response is AuthenticatorAssertionResponse) {
+            Log.d(TAG, "Finish with success response: $response")
             sendResult(JSONObject().apply {
                 val base64Flags = Base64.NO_PADDING + Base64.NO_WRAP + Base64.URL_SAFE
                 put("keyHandle", response.keyHandle?.toBase64(base64Flags))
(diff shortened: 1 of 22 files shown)


ale5000-git:
@TheMartinizer
The other 2 PR are now merged, could you please remove the already merged changes from this PR?

TheMartinizer:
All right, I rebased the credential_selection branch onto the current microG master and pushed it again! Thanks to you guys for looking through the other pull requests, and I'm glad they were useful! :-)

ale5000-git:
Certainly, it is a very good contribution.
Thanks a lot for the help.

dylangerdaly:
Woah, will this totally add support for microG to manage/store/use Passkeys?

ale5000-git:
@mar-v-in
Any plan for this PR?

TheStormN:
@ale5000-git Hey, do you plan to merge this PR?

ale5000-git:
> @ale5000-git Hey, do you plan to merge this PR?

It is @mar-v-in that handle PRs.

HenrikFrtz:
@mar-v-in Any plans to add this to the upcoming milestones? ☺️

ale5000-git:
@TheMartinizer
When you have time can you please rebase onto the current microG master again?

p1gp1g:
It looks good to me, and I'm already using it in my app. But:
- The key selection isn't usable as it is when used in an app that targets SDK 35. @ale5000-git are you ok to [add a workaround](https://github.com/TheMartinizer/GmsCore/pull/2) until edge to edge is implemented in microg ?
- It misses User Handle support, which is required by some websites. I've opened a PR: https://github.com/TheMartinizer/GmsCore/pull/1 ; if this (2463) is merged before, I can open a PR to the upstream repo too

ale5000-git:
@mar-v-in
Any plan for this PR?

ale5000-git:
@mar-v-in
Is this planned for 0.3.10? [Fido2: Support for user handles](https://github.com/TheMartinizer/GmsCore/pull/1)

mar-v-in:
@ale5000-git There's overlap between this one and other PRs, so the idea is to merge them all together in 0.3.10 to prevent them from further driving apart.

PR #2491, opened 2024-08-17
Account Settings: Take product name into MyActivity page

DaVinci9196:
Fixed the issue in YouTube settings where clicking Manage all history would redirect to Google search history.
Google search history and YouTube Manage all history share the same screenId.


WSTxda:
Isn't it better to use a history-related link like: https://myactivity.google.com/product/youtube

It should be noted that the option "manage your data on YouTube" should probably open this same link

mar-v-in:
@WSTxda this is what the PR does: When MyActivity page (screen id 547) is requested and a product is set (e.g. YouTube will set product to `youtub`) replace `search` in `https://myactivity.google.com/product/search` with requested product name (creating e.g. `https://myactivity.google.com/product/youtube`).

Issue #2225, opened 2024-03-08
[CODE] Partial implementation of Work Account and Split Install Service


ale5000-git:
I have adjusted the patches done by others to work with the current code of microG, so they now merge and compile correctly but they aren't ready to be used.

I hope someone pick and complete them.

Work Account (by @fynngodau): https://github.com/microg/GmsCore/compare/master...micro5k:GmsCoreMod:patch-work-account.patch

Split Install Service (by @jonathanklee): https://github.com/microg/GmsCore/compare/master...micro5k:GmsCoreMod:patch-split-install-service.patch

CoelacanthusHex:
I found the functions used to get APK/OBB of Aurora Store is separate kotlin library, so we can reuse it directly.

https://gitlab.com/AuroraOSS/gplayapi

Iey4iej3:
Is downloading performed by MicroG or relayed to third-party APPs such as Aurora Store? It seems conceptually better to relay, but not sure whether it would be difficult to cooperate.

jonathanklee:
I believe the split APKs mentioned [here](https://gitlab.com/AuroraOSS/gplayapi/-/blob/master/lib/src/main/java/com/aurora/gplayapi/helpers/PurchaseHelper.kt?ref_type=heads#L221) are install-time APKs that we need to distinguish from the on-demand APKs: https://developer.android.com/guide/playcore/feature-delivery/on-demand

I have a fork of gplayapi with the proper on-demand APKs requests. I'll push it upstream.

Issue #1629, opened 2021-12-30
Support for split install after installation


Iey4iej3:
Some apps seem to rely on some kind of post-install split installation to function. For example, some functionalities of WeChat does not work. I look at the log, and it contains

> 12-30 17:39:03.153 28367 28538 I chatty  : uid=1010197(com.tencent.mm) [GT]ColdPool#9 expire 2 lines
12-30 17:39:03.163 28367 28481 I chatty  : uid=1010197(com.tencent.mm) SplitInstallSer expire 2 lines
12-30 17:39:03.166  5147  6291 W ActivityManager: Unable to start service Intent { act=com.google.android.play.core.splitinstall.BIND_SPLIT_INSTALL_SERVICE pkg=com.android.vending } U=10: not found
12-30 17:39:03.168     0     0 D cnss    : Setting MHI state: SUSPEND(6)
12-30 17:39:03.168     0     0 D cnss    : Suspending PCI link
12-30 17:39:03.168     0     0 D cnss    : Use PCIe DRV suspend
12-30 17:39:03.170     0     0 D cnss    : Runtime suspend status: 0
12-30 17:39:03.187 29209 29369 I chatty  : uid=1010197(com.tencent.mm) expire 24 lines
12-30 17:39:03.231  1011  1166 I WVCdm   : [app][com.tencent.mm] calling virtual Return<void> wvdrm::hardware::drm::V1_3::widevine::WVDrmFactory::createPlugin(const hidl_array<uint8_t, 16> &, const android::hardware::hidl_string &, android::hardware::drm::V1_3::IDrmFactory::createPlugin_cb)
12-30 17:39:03.231  1011  1166 D WVCdm   : Instantiating CDM.
12-30 17:39:03.265 28349 28583 I chatty  : uid=1010197(com.tencent.mm) [GT]ColdPool#9 expire 2 lines
12-30 17:39:03.266 28349 28513 I chatty  : uid=1010197(com.tencent.mm) SplitInstallSer expire 2 lines
12-30 17:39:03.268  5147  6291 W ActivityManager: Unable to start service Intent { act=com.google.android.play.core.splitinstall.BIND_SPLIT_INSTALL_SERVICE pkg=com.android.vending } U=10: not found
12-30 17:39:03.419  5147  6231 E TaskPersister: File error accessing recents directory (directory doesn't exist?).
12-30 17:39:03.505 29209 29377 I chatty  : uid=1010197(com.tencent.mm) expire 3 lines
[...]

PR #2500, opened 2024-08-19
Implement SplitInstallService

DaVinci9196:
Fix maps language modification and Fix Smart Lens translation language download


CoelacanthusHex:
Will it fix https://github.com/microg/GmsCore/issues/1629 and partially resolve https://github.com/microg/GmsCore/issues/2225?

vending-app/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,8 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="com.google.android.gms.permission.READ_SETTINGS" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES"/>
+    <uses-permission android:name="android.permission.READ_PRIVILEGED_PHONE_STATE"/>
```
ale5000-git, line 23: Is `android.permission.READ_PRIVILEGED_PHONE_STATE` really needed?
I think most microG users won't like this, the main intent of microG is to spoof data specially the IMEI.

mar-v-in:
This PR seems to be mixing completely unrelated features (Phenotype / Experiment Flags and SplitInstallService). I don't think SplitInstallService strictly needs Phenotype related functionality. Please clean this up so the PR only contains the SplitInstallService

vending-app/build.gradle
```diff
@@ -107,6 +107,7 @@ dependencies {
     implementation 'androidx.activity:activity-compose:1.7.2'
     implementation("io.coil-kt:coil-compose:2.4.0")
     implementation("io.coil-kt:coil-svg:2.2.2")
+    implementation 'org.brotli:dec:0.1.2'
```
grote, line 110: What is this used for?

vending-app/src/main/kotlin/com/android/vending/ExperimentAndConfigs.kt
```diff
@@ -0,0 +1,839 @@
...
+            httpURLConnection.readTimeout = 30000
+            httpURLConnection.doOutput = true
+            httpURLConnection.instanceFollowRedirects = false
+            httpURLConnection.setRequestProperty("Accept-Encoding", null)
+            httpURLConnection.setRequestProperty("Content-Type", "application/x-protobuf")
+            httpURLConnection.setRequestProperty("Content-Encoding", "gzip")
+            httpURLConnection.setRequestProperty("Authorization", "Bearer $token")
+            httpURLConnection.setRequestProperty(
+                "User-Agent",
+                "Android-Finsky/41.2.21-31 [0] [PR] 636997666 (api=3,versionCode=84122130,sdk=31,device=redroid_arm64,hardware=redroid,product=redroid_arm64,platformVersionRelease=12,model=redroid12_arm64,buildId=SQ1D.220205.004,isWideScreen=0,supportedAbis=arm64-v8a;armeabi-v7a;armeabi) (redroid_arm64 SQ1D.220205.004); gzip"
```
grote, line 234: `41.2.21-31 [0] [PR] 636997666` is also defined in the manifest now. Does it need to be the same?

Lots of hardware properties seem to be hardcoded in the request. Do those matter at all and potentially change the response?

vending-app/src/main/kotlin/com/android/vending/PhenotypeDatabase.kt
```diff
@@ -0,0 +1,58 @@
...
+        db.execSQL("  CREATE TABLE IF NOT EXISTS LastFetch(\n    \"key\" INTEGER NOT NULL PRIMARY KEY,\n    servertimestamp INTEGER NOT NULL\n  )\n")
+        db.execSQL("  CREATE TABLE IF NOT EXISTS FlagOverrides(\n    packageName TEXT NOT NULL,\n    user TEXT NOT NULL,\n    name TEXT NOT NULL,\n    flagType INTEGER NOT NULL,\n    intVal INTEGER,\n    boolVal INTEGER,\n    floatVal REAL,\n    stringVal TEXT,\n    extensionVal BLOB,\n    committed,\n    PRIMARY KEY(packageName, user, name, committed)\n  );\n")
+        db.execSQL("  CREATE TABLE IF NOT EXISTS LastSyncAfterRequest(\n    packageName TEXT NOT NULL PRIMARY KEY,\n    servingVersion INTEGER NOT NULL DEFAULT 0,\n    androidPackageName TEXT DEFAULT NULL\n  )\n")
+        db.execSQL("  CREATE TABLE IF NOT EXISTS StorageInfos (\n    androidPackageName TEXT UNIQUE NOT NULL,\n    secret BLOB NOT NULL,\n    deviceEncryptedSecret BLOB NOT NULL\n  )\n")
+        db.execSQL("  CREATE TABLE AppWideProperties (\n    androidPackageName TEXT UNIQUE NOT NULL,\n    appWideProperties BLOB NOT NULL\n  );\n")
```
grote, line 41: This doesn't look very maintainable to me. Have you considered using something like Room instead?

DaVinci9196:
> This PR seems to be mixing completely unrelated features (Phenotype / Experiment Flags and SplitInstallService). I don't think SplitInstallService strictly needs Phenotype related functionality. Please clean this up so the PR only contains the SplitInstallService

I submitted a new revision

play-services-nearby/core/src/test/java/com/google/android/finsky/splitinstallservice/ExampleUnitTest.java
mar-v-in, line 1: I don't think we need this...

vending-app/src/main/AndroidManifest.xml
```diff
@@ -93,6 +98,7 @@
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
+                <category android:name="android.intent.category.LAUNCHER"/>
```
mar-v-in, line 101: Please don't add LAUNCHER here for the regular build, this will clutter launchers of users with an unnecessary entry. We already add LAUNCHER category for Huawei devices via the Huawei flavor.

vending-app/src/main/AndroidManifest.xml
```diff
@@ -56,6 +57,10 @@
         android:roundIcon="@mipmap/ic_app"
         android:label="@string/app_name">
 
+        <meta-data
+            android:name="GpVersion"
+            android:value="41.2.21-31 [0] [PR] 636997666"/>
+
```
mar-v-in, line 63: What is this needed for?

vending-app/src/main/java/org/microg/vending/billing/Constants.kt
```diff
@@ -14,4 +14,6 @@ const val VENDING_PACKAGE_NAME = "com.android.vending"
 // TODO: Replace key name
 const val KEY_IAP_SHEET_UI_PARAM = "key_iap_sheet_ui_param"
 const val DEFAULT_ACCOUNT_TYPE = "com.google"
-const val ADD_PAYMENT_METHOD_URL = "https://play.google.com/store/paymentmethods"
\ No newline at end of file
+const val ADD_PAYMENT_METHOD_URL = "https://play.google.com/store/paymentmethods"
+const val FINSKY_REGULAR = "com.google.android.finsky.regular"
+const val FINSKY_STABLE = "com.google.android.finsky.stable"
```
mar-v-in, line 19: Those two are not used.

vending-app/src/main/kotlin/com/google/android/phonesky/header/PhoneskyHeaderValue.kt
```diff
@@ -0,0 +1,275 @@
...
+        return String(Base64.encode(Util.encodeGzip(header), Base64.URL_SAFE or Base64.NO_WRAP or Base64.NO_PADDING))
+    }
+
+    private fun getHeaders(): Map<String, String> {
+        headerMap["X-PS-RH"] = getXHeaders()
+        headerMap["Authorization"] = "Bearer " + AccountManager.get(context).getAuthToken(
+            user, tokenType, null, false, null, null
+        ).result.getString(AccountManager.KEY_AUTHTOKEN)
+        return this.headerMap
+    }
```
mar-v-in, line 216: This seems to mostly overlap with `getLicenseRequestHeaders` in `LicenseRequestHeaders.kt`. Maybe you can refactor this into a common codebase.

mar-v-in (changes requested):

DaVinci9196:
We modified it according to your opinion

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/extensions.kt
```diff
@@ -0,0 +1,347 @@
...
+private val deferredMap = mutableMapOf<Int, CompletableDeferred<Intent>>()
+
+private var lastSplitPackageName: String? = null
+private val splitRecord = arrayListOf<Array<String>>()
+
+private fun Context.splitSaveFile() = File(filesDir, FILE_SAVE_PATH)
+
+suspend fun trySplitInstall(context: Context, httpClient: HttpClient, pkg: String, splits: List<Bundle>) {
+    if (lastSplitPackageName != null && lastSplitPackageName != pkg) {
+        if (mutex.isLocked) mutex.unlock()
```
fynngodau, line 82: When does this case occur?

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/extensions.kt
```diff
@@ -0,0 +1,347 @@
...
+                Log.w(TAG, "downloadSplitPackage urls:${urls.contentToString()}: ", it)
+            }.getOrNull() != null
+        }
+    }.awaitAll()
+    return@coroutineScope results.all { it }
+}
+
+private suspend fun installSplitPackage(context: Context, httpClient: HttpClient, downloadUrl: ArrayList<Array<String>>, packageName: String, language: String?): Intent {
+    Log.d(TAG, "installSplitPackage downloadUrl: ${downloadUrl.firstOrNull()}")
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
fynngodau, line 227: This check is only executed after calling `getDownloadUrls`. I would suggest performing the check earlier in the flow, and annotating relevant methods with `@RequiresApi(Build.VERSION_CODES.LOLLIPOP)`.

DaVinci9196:
@fynngodau Accepted, thanks for the suggestion.
In practice, the packageName provided by the calling method cannot be verified by PackageUtils.getAndCheckCallingPackage to be the same calling

fynngodau:
Thanks for providing us with a new iteration!

I now noticed that `HeaderProvider` and `com.android.vending.extensions` are partially overlapping and duplicate, and the latter is not well-integrated with the classes (`AuthData` and `DeviceEnvInfo`) used by the former. Would it be possible to harmonize and merge the two files? Though I see the difficulty considering that some calls seem to want an `x-ps-rh` header and others do not.

Is there a reason you are using `1` instead of the real Android ID in your queries?

>  In practice, the packageName provided by the calling method cannot be verified by PackageUtils.getAndCheckCallingPackage to be the same calling

I can't see why it wouldn't work the way you tried, but maybe @mar-v-in has an idea.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+                }
+                totalDownloaded += pkgPath.length()
+                pkgPath.delete()
+            }
+            val deferred = CompletableDeferred<Intent>()
+            deferredMap[sessionId] = deferred
+            val intent = Intent(context, InstallResultReceiver::class.java).apply {
+                putExtra(KEY_BYTES_DOWNLOADED, totalDownloaded)
+            }
+            val pendingIntent = PendingIntent.getBroadcast(context, sessionId, intent, 0)
```
jonathanklee, line 150:
```suggestion
            val pendingIntent = PendingIntent.getBroadcast(context, sessionId, intent, PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE)
```

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+                }
+            }.awaitAll()
+            return@coroutineScope results.all { it }
+        }
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    private suspend fun requestDownloadUrls(callingPackage: String, authToken: String, packs: MutableSet<String>): ArraySet<Triple<String, String, Int>> {
+        val versionCode = PackageInfoCompat.getLongVersionCode(context.packageManager.getPackageInfo(callingPackage, 0))
+        val requestUrl =
+            StringBuilder("https://play-fe.googleapis.com/fdfe/delivery?doc=$callingPackage&ot=1&vc=$versionCode&bvc=$versionCode&pf=1&pf=2&pf=3&pf=4&pf=5&pf=7&pf=8&pf=9&pf=10&da=4&bda=4&bf=4&fdcf=1&fdcf=2&ch=")
```
jonathanklee, line 183: I tried to execute the split installations from TikTok, but I wasn't able to download the split APKs until I changed the URL to:
```
https://play-fe.googleapis.com/fdfe/delivery?doc=$callingPackage&ot=1&vc=$versionCode
```
so removing the last part: 

```
&bvc=$versionCode&pf=1&pf=2&pf=3&pf=4&pf=5&pf=7&pf=8&pf=9&pf=10&da=4&bda=4&bf=4&fdcf=1&fdcf=2&ch=
```

What app did you use to test the split install feature ?

jonathanklee: When testing with TikTok, I face the following error;

```
SplitInstallManager: InstallResultReceiver onReceive: install fail -> INSTALL_FAILED_MISSING_SPLIT: Missing split for com.zhiliaoapp.musically
```

jonathanklee: @DaVinci9196 I was able to install a language split apk on PayPal :+1:

DaVinci9196: > I tried to execute the split installations from TikTok, but I wasn't able to download the split APKs (the url triples were always null) until I changed the URL to:
> 
> ```
> https://play-fe.googleapis.com/fdfe/delivery?doc=$callingPackage&ot=1&vc=$versionCode
> ```
> 
> so removing the last part:
> 
> ```
> &bvc=$versionCode&pf=1&pf=2&pf=3&pf=4&pf=5&pf=7&pf=8&pf=9&pf=10&da=4&bda=4&bf=4&fdcf=1&fdcf=2&ch=
> ```
> 
> What app did you use to test the split install feature ?

Tested by switching languages ​​in Google Maps and Bolt

vending-app/src/main/java/org/microg/vending/billing/core/HttpClient.kt
```diff
@@ -10,14 +10,47 @@ import com.squareup.wire.Message
...
+        requestQueue.add(object : Request<String>(Method.GET, uriBuilder.build().toString(), null) {
+            override fun parseNetworkResponse(response: NetworkResponse): Response<String> {
+                if (response.statusCode != 200) throw VolleyError(response)
+                return try {
+                    val parentDir = downloadFile.getParentFile()
+                    if (parentDir != null && !parentDir.exists() && !parentDir.mkdirs()) {
+                        throw IOException("Failed to create directories: ${parentDir.absolutePath}")
+                    }
+                    val fos = FileOutputStream(downloadFile)
+                    fos.write(response.data)
```
fynngodau, line 37: Volley load the entire response into memory. This is a problem for large downloads, as it will consume a lot of RAM (temporarily). The data from the network stream should be written to disk immediately instead.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+        splitInstallRecord[callingPackage]?.let { triples ->
+            val find = triples.find { it.first == triple.first }
+            find?.let { triples.remove(it) }
+            triples.add(triple)
+        } ?: run {
+            val triples = ArraySet<Triple<String, String, Int>>()
+            triples.add(triple)
+            splitInstallRecord[callingPackage] = triples
+        }
+    }
```
fynngodau, line 246: This method would be entirely unnecessary if `splitInstallRecord` had the data type `Map<String, Map<PackageComponent, DownloadStatus>>` instead of `HashMap<String, ArraySet<Triple<String, String, Int>>>`, with these new classes:

```
data class PackageComponent(
    val componentName: String,
    val url: String
)

enum class DownloadStatus {
    PENDING,
    FAILED,
    COMPLETE
}
```

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+
+    private fun notify(context: Context) {
+        val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            notificationManager.createNotificationChannel(
+                NotificationChannel(NOTIFY_CHANNEL_ID, NOTIFY_CHANNEL_NAME, NotificationManager.IMPORTANCE_DEFAULT)
+            )
+        }
+        NotificationCompat.Builder(context, NOTIFY_CHANNEL_ID).setSmallIcon(android.R.drawable.stat_sys_download)
+            .setContentTitle(context.getString(R.string.split_install, context.getString(R.string.app_name))).setPriority(NotificationCompat.PRIORITY_DEFAULT)
```
fynngodau, line 256: The placeholder in `R.string.split_install` should be filled in with the downloading app's name, not with microG's name.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
fynngodau: You can find my suggestions here: https://github.com/e-foundation/android_packages_apps_GmsCore/blob/c3398b989ab1a01aaf0bed7a6900cc0f46225ad6/vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
fynngodau: @jonathanklee With `FLAG_IMMUTABLE`, I am not receiving any response values in the `InstallResultReceiver`, so the receiver can't detect if installation was successful or not. So `FLAG_MUTABLE` seems better.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+            .setDefaults(
+                NotificationCompat.DEFAULT_ALL
+            ).build().also {
+                notificationManager.notify(SPLIT_INSTALL_NOTIFY_ID, it)
+            }
+    }
+
+    private fun Context.splitSaveFile() = File(filesDir, FILE_SAVE_PATH)
```
fynngodau, line 264: Installed packages are deleted after installation, but if installation fails and never retried (for instance: one large package is downloaded successfully, then a small package is downloaded unsuccessfully due to a network error, then the user uninstalls the app because it doesn't work), this can still lead to storage leaks from downloaded-but-never-installed apps. We should use `cacheDir` instead: it will allow the system and users to clear the cache themselves.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -0,0 +1,337 @@
...
+
+    private fun sendCompleteBroad(context: Context, packageName: String, intent: Intent) {
+        Log.d(TAG, "sendCompleteBroadcast: intent:$intent")
+        val extra = Bundle().apply {
+            putInt(KEY_STATUS, 5)
+            putInt(KEY_ERROR_CODE, 0)
+            putInt(KEY_SESSION_ID, 0)
+            putLong(KEY_TOTAL_BYTES_TO_DOWNLOAD, intent.getLongExtra(KEY_BYTES_DOWNLOADED, 0))
+            putString(KEY_LANGUAGES, intent.getStringExtra(KEY_LANGUAGE))
```
fynngodau, line 273: It doesn't seem like this intent would have this field, given that it is not set upon creation of the intent in `installSplitPackage`.

fynngodau:
I opened PR #2553, which contains the code from this PR; I performed some major refactoring to `SplitInstallService` there.

PR #2503, opened 2024-08-21
MovingWifiHelper: Fix _SNCF_WIFI_INTERCITES

valerierx:
The new JSON dosen't include a "fix" field on french intercity trains anymore, but they're still in use on TGVs and OUIGO

### DIFF
--- a/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
+++ b/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
@@ -247,7 +247,7 @@ class MovingWifiHelper(private val context: Context) {
     
     private fun parseSncf(location: Location, data: ByteArray): Location {
         val json = JSONObject(data.decodeToString())
-        if(json.getInt("fix") == -1) throw RuntimeException("GPS not valid")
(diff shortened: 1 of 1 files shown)


ale5000-git:
@valerierx
The build fails.

play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
```diff
@@ -247,7 +247,9 @@ class MovingWifiHelper(private val context: Context) {
     
     private fun parseSncf(location: Location, data: ByteArray): Location {
         val json = JSONObject(data.decodeToString())
-        if(json.getInt("fix") == -1) throw RuntimeException("GPS not valid")
+        json.optInt("fix").takeIf { !it.isNaN() }?.let {
```
mar-v-in, line 250: Int can't be NaN. Also, if this field is just not present in Intercites, you can just change the condition to `json.has("fix") && json.getInt("fix") == -1`

mar-v-in (changes requested):

play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
valerierx: Thanks!

mar-v-in:
Thanks.

PR #2506, opened 2024-08-22
Implement AssetModuleService

DaVinci9196:
The three applications were tested and can be downloaded and installed normally through the service.
e.g. com.wb.goog.mkx , com.gameloft.android.ANMP.GloftDYHM , jp.konami.pesam


bobaoapae:
Hey, can this be released on some canary build? I was running waze on a device thwt don't have play services, and installed gms core but waze was unable to download some files that I think it's because of this missing implementation. I would love to test this.

DaVinci9196:
> Hey, can this be released on some canary build? I was running waze on a device thwt don't have play services, and installed gms core but waze was unable to download some files that I think it's because of this missing implementation. I would love to test this.

If possible, you can first check whether this service is called. At present, we only found that the game will download resource files through this service.

mar-v-in:
Device Sync is not related to Asset Modules, please put it in a separate pull request.

vending-app/src/main/kotlin/com/google/android/finsky/DownloadManager.kt
```diff
@@ -0,0 +1,247 @@
...
+        notificationLayout.setProgressBar(R.id.progress_bar, 100, 0, false)
+        notificationLayout.setImageViewBitmap(R.id.app_icon, bitmap)
+        notificationLayout.setOnClickPendingIntent(R.id.cancel_button, cancelPendingIntent)
+
+        val notifyBuilder =
+            NotificationCompat.Builder(context, CHANNEL_ID).setSmallIcon(R.drawable.ic_app_foreground).setStyle(NotificationCompat.DecoratedCustomViewStyle())
+                .setCustomContentView(notificationLayout).setPriority(NotificationCompat.PRIORITY_LOW).setOngoing(true).setOnlyAlertOnce(true)
+                .setColor(ContextCompat.getColor(context, R.color.notification_color)).setColorized(true)
+        notifyBuilderMap[moduleName] = notifyBuilder
+        notificationLayoutMap[moduleName] = notificationLayout
```
mar-v-in, line 132: Any reason you use a custom notification layout instead of a default with [setContentTitle()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setContentTitle(java.lang.CharSequence)), [setContentText()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setContentText(java.lang.CharSequence)), [setProgress()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setProgress(int,int,boolean)), [setLargeIcon()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setLargeIcon(android.graphics.Bitmap)) and [addAction()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#addAction(int,java.lang.CharSequence,android.app.PendingIntent))? That should cover the same content but not require a custom view with custom styling, thus better compatible with various android versions.

DaVinci9196:
@mar-v-in During our testing, we discovered anomalies in the game's resource download and installation process and have fixed them.

mar-v-in:
Thanks for the work, I merged this despite a few issues already known, which I linked above, so that we get this rolling for testing in 0.3.5 and won't divert further from the main branch.

PR #2521, opened 2024-09-02
Support work account creation through device policy controller

fynngodau:
Implements a work account service API that is used by [Device Policy Controller apps](https://developer.android.com/work/dpc/build-dpc) through the proprietary DPC support library.

Only device / profile owners are allowed to use this API (GMS instead uses an allowlist). The DPC support library will use the API to enable the work account authenticator component, making the option to add a "Managed Google for Work account" visible in the Accounts screen for the work profile. However, the account cannot be created manually. Instead, it can only be created by admin apps / the DPC app which can [create a token for this purpose](https://developers.google.com/android/work/play/emm-api/v1/users/generateAuthenticationToken).

This token is used to sign in to the managed work account. Like on GMS, the account will have a name like `work-25[…]8D@android-for-work.gserviceaccount.com`.

This managed work account should allow access to a scoped version of the Google Play store, but is not used at all as-is. As of now, it is not possible for a user to download any apps to the work profile if the policy controller does not install any app store other than Google Play to the profile.


fynngodau:
The files `AuthRequest.kt` and `AuthResponse.kt` are uselessly duplicated in the current state, because I don't know a good place to put them as I don't want to include `:play-services-core` in the `:play-services-workaccount-core` module. Maybe @mar-v-in has a suggestion?

Scutua (approved):

mar-v-in:
Superseded by #2553 (merged)

PR #2531, opened 2024-09-06
Fix getMapAsync() for LiteGoogleMap

jonathanklee:
Spotify was crashing because of some lateinit property in Spotify snapshot ready callback was not initialized.

Looking at the condition we can observe that GmsCore was calling the snapshot ready callback when the snapshot was null, which does not make sense.

Inverting the logic and calling the callback when the snapshot was not null anymore fixed the issue and Spotify was able to display the snapshot.


fynngodau (approved):
The log messages make it obvious that I mixed this up. Thanks :)

PR #2534, opened 2024-09-09
IAP: Implement old getBuyIntent through forwarding to new one

DaVinci9196:
Fixed the issue that the payment pop-up page could not be displayed after clicking on the payment.
e.g. uk.co.nationalrail.google

### DIFF
--- a/vending-app/src/main/java/org/microg/vending/billing/InAppBillingServiceImpl.kt
+++ b/vending-app/src/main/java/org/microg/vending/billing/InAppBillingServiceImpl.kt
@@ -232,7 +232,8 @@ class InAppBillingServiceImpl(private val context: Context) : IInAppBillingServi
         developerPayload: String?
     ): Bundle {
         if (Log.isLoggable(TAG, Log.DEBUG)) Log.d(TAG, "getBuyIntent(apiVersion=$apiVersion, packageName=$packageName, sku=$sku, type=$type, developerPayload=$developerPayload)")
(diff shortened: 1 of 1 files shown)


vending-app/src/main/java/org/microg/vending/billing/InAppBillingServiceImpl.kt
```diff
@@ -232,7 +232,8 @@ class InAppBillingServiceImpl(private val context: Context) : IInAppBillingServi
         developerPayload: String?
     ): Bundle {
         if (Log.isLoggable(TAG, Log.DEBUG)) Log.d(TAG, "getBuyIntent(apiVersion=$apiVersion, packageName=$packageName, sku=$sku, type=$type, developerPayload=$developerPayload)")
-        return resultBundle(BillingResponseCode.BILLING_UNAVAILABLE, "Not yet implemented")
+        return runCatching { getBuyIntentExtraParams(apiVersion, packageName!!, sku!!, type!!, developerPayload, null) }
```
mar-v-in, line 235: Why is it necessary to catch exceptions here? Shouldn't they be catched in `getBuyIntentExtraParams` if that was necessary?

DaVinci9196: It is not certain whether the passed parameter is empty. Temporarily catch the exception here to avoid abnormal crash.

PR #2536, opened 2024-09-11
Google Help: Improve the link target

DaVinci9196:
The test found that there are two places where you cannot jump to the help page.
1. Gmail meeting menu pop-up window click to learn more Unable to jump.
2. Clicking Learn how to verify your phone number on the Meet add phone number page does not redirect.

### DIFF
--- a/play-services-core-proto/src/main/proto/help.proto
+++ b/play-services-core-proto/src/main/proto/help.proto
@@ -1,0 +1,53 @@
+/*
+ * SPDX-FileCopyrightText: 2023 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+option java_package = "org.microg.gms.googlehelp";
+
+option java_multiple_files = true;
+
+message RequestContent {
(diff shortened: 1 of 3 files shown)


mar-v-in:
I don't think it's relevant that we sign in the account to view help pages, so instead of using an authorized web view, you could safely redirect to the browser (as was done before) after retrieving the URL.

DaVinci9196:
> I don't think it's relevant that we sign in the account to view help pages, so instead of using an authorized web view, you could safely redirect to the browser (as was done before) after retrieving the URL.

Accepted, thanks

ale5000-git:
@DaVinci9196
Look at the errors at the boottom: https://github.com/microg/GmsCore/actions/runs/10822743326?pr=2536

DaVinci9196:
> @DaVinci9196 Look at the errors at the bottom: https://github.com/microg/GmsCore/actions/runs/10822743326?pr=2536
I need to jump to an external browser, how can I do this?

mar-v-in (approved):

PR #2553, opened 2024-09-22
Work app store

fynngodau:
Supersedes and contains #2500. Supersedes and contains #2521.

Allows users to download apps to their work profile using the special vending component "Work app store". This component is disabled, and enabled only when a work profile is added (which in turn can only be done by a device or profile admin).

Yet to do:

* rewrite download code for components (https://github.com/microg/GmsCore/pull/2500#discussion_r1769604559)
    * [x] don't allocate RAM for the entire download
    * [x] download directly to installer session
* add meaningful progress information
    * [x] to store app
    * [x] while downloading in the background through store app
    * [x] while splitinstalling
* [x] determine and verify meaning of policy field in `/getEnterpriseClientPolicy`
* [x] test additional dpc services
* [x] show Work app store only after adding work account
* [x] automatic dependency installation (i.e. installing Google Chrome), see https://gitlab.com/AuroraOSS/AuroraStore/-/issues/346)
* [x] acquire user consent before provisioning work account → new settings screen in microG UI, is shared across accounts (note that custom ROM vendors are expected to grant microG the `INTERACT_ACROSS_USERS` permission for this)
* [ ] automatically install / remove apps on demand of dpc controller → for future investigations

Mostly up-to-date screenshot:

Screenshot_1727017306


ale5000-git:
@fynngodau
Can it help with Shelter? https://f-droid.org/it/packages/net.typeblog.shelter/

fynngodau:
@ale5000-git No. This PR adds support for [Device Policy Controller apps](https://developer.android.com/work/dpc/build-dpc) that want to register a Work google account. See #2521.

ale5000-git:
Shelter use a work profile, and I remember it was failing on my device but I'm not sure if the issue is with microG or with the ROM.

DaVinci9196:
@fynngodau Can the download and notification process of game sub-packages be incorporated into this project? The corresponding functions have been implemented in https://github.com/microg/GmsCore/pull/2506.

fynngodau:
@DaVinci9196 That would be beneficial, though I will not incorporate another PR into this one as it is already quite large.

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
```diff
@@ -97,7 +98,22 @@ class VendingActivity : ComponentActivity() {
...
+                            "ot" to "1",
+                            "doc" to app.packageName,
+                            "vc" to app.versionCode.toString()
+                        )
+                        client.post(
+                            url = URL_PURCHASE,
+                            headers = buildRequestHeaders(auth!!.authToken, auth!!.gsfId.toLong(16)),
+                            params = parameters,
+                            adapter = GoogleApiResponse.ADAPTER
+                        )
```
ale5000-git: @fynngodau
I don't know the code enough but does it check the microG option before purchase?

fynngodau: That wouldn't make sense to me, as this is neither in-app billing nor during an app license check.

ale5000-git: I mean the option: Auto purchase free apps

vending-app/src/main/java/org/microg/vending/billing/core/HttpClient.kt
```diff
@@ -1,140 +1,221 @@
...
+import java.io.FileOutputStream
+import java.io.IOException
+import java.io.OutputStream
 
-private const val POST_TIMEOUT = 8000
+private const val POST_TIMEOUT = 8000L
 
-class HttpClient(context: Context) {
-    private val requestQueue = singleInstanceOf { Volley.newRequestQueue(context.applicationContext) }
+class HttpClient {
```
jonathanklee, line 34: May be renamed to DownloadClient if it makes more sens to you also ?

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
```diff
@@ -0,0 +1,324 @@
...
+                        apps[app] = previousState
+                    }
+
+                }
+            }.start()
+        }
+
+        setContent {
+            VendingUi(account, install, uninstall)
+        }
```
jonathanklee, line 187: There is a lot of logic in the onCreate() callback. Maybe we can extract some part into separate methods ?

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
```diff
@@ -0,0 +1,324 @@
...
+import org.microg.vending.proto.AppMeta
+import org.microg.vending.proto.GetItemsRequest
+import org.microg.vending.proto.RequestApp
+import org.microg.vending.proto.RequestItem
+import org.microg.vending.ui.components.EnterpriseList
+import org.microg.vending.ui.components.NetworkState
+import java.io.IOException
+
+@RequiresApi(android.os.Build.VERSION_CODES.LOLLIPOP)
+class VendingActivity : ComponentActivity() {
```
jonathanklee, line 64: Is there a more specific name we could find for VendingActivity ?

vending-app/src/main/java/org/microg/vending/billing/core/HttpClient.kt
fynngodau: @jonathanklee That doesn't make so much sense to me, as it not only downloads files, but also does other queries using both GET and POST.

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
ale5000-git: @fynngodau
It can also be a new option but it shouldn't automatically purchase apps (even if free) if not configured in microG settings.

mar-v-in:
The PR is getting huge and there are still some issues with split install (can cause repeated downloads resulting in a lot of data usage). To get 0.3.6 out before Christmas, I moved this to 0.3.7. I'll see if we can split this up again and make sure that all the good code can land in 0.3.7 in January.

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
ale5000-git: @fynngodau
If the existing option doesn't apply, I think we should add a new one, since anything that causes contact with a remote server (and thus potentially reduces privacy) should be able to be disabled by the user.

ale5000-git:
Split Install Service is also used outside "work app store", without it some apps may not be able to download the necessary components so I think it should have priority in my opinion.

Why not split it (the rewritten one) in a separate PR so it can be merged in less time?

Uldiniad:
Hi, was wondering if there'd be any use for classes like https://review.calyxos.org/c/CalyxOS/platform_external_microg_GmsCore/+/33556/3/play-services-auth/src/main/java/com/google/android/gms/auth/account/WorkAccountApi.java ?

Iey4iej3:
> Split Install Service is also used outside "work app store", without it some apps may not be able to download the necessary components so I think it should have priority in my opinion.
> 
> Why not split it (the rewritten one) in a separate PR so it can be merged in less time?

Agree with this.

Scutua (approved):

play-services-auth-workaccount/core/src/main/res/values/string.xml
lucasmz-dev, line 1: Shouldn't this file be called `strings.xml` instead of `string.xml`? That's what I usually see on Android apps; might not matter, just kind of weird.

fynngodau:
> Hi, was wondering if there'd be any use for classes like https://review.calyxos.org/c/CalyxOS/platform_external_microg_GmsCore/+/33556/3/play-services-auth/src/main/java/com/google/android/gms/auth/account/WorkAccountApi.java ?

@Uldiniad As far as I can tell, these classes from https://developers.google.com/android/reference/com/google/android/gms/auth/account/WorkAccount are used as an interface for the consumer of the DPC controller library, aren't they? I don't think we need to implement those in microG, as they seem not to be part of the library-to-GMS API.

vending-app/src/main/java/org/microg/vending/ui/VendingActivity.kt
fynngodau: It still doesn't make sense to me to have an option for this. Disabling it would only (randomly) break downloading apps from the Work app store. This "purchase" action also doesn't happen automatically, but only if a user decides to install an app from the store by pressing the corresponding button. It is to be expected that this triggers all the calls necessary to install the app.

vending-app/src/main/java/org/microg/vending/ui/WorkAppsActivity.kt
```diff
@@ -0,0 +1,390 @@
...
+                    "towards misuse of the work account service API by the DPC.")
+        }
+        val account = accounts.first()
+
+        load(account)
+
+        setContent {
+            VendingUi(account,
+                install = { app, isUpdate -> Thread { runBlocking { install(app, isUpdate) } }.start() },
+                uninstall = { app -> Thread { runBlocking { uninstall(app) } }.start() }
```
fynngodau, line 100: This should happen on a service to be independent of the lifecycle of this activity.

mar-v-in (approved):
lgtm

agentorange96:
/e/OS 3.1.1/3.1.2 finally include this fix. I am not sure how to access the "Work App Store" though. I am only sent to the MicroG settings.

I can now set up my work profile, but I don't see how to install work apps.

Am I missing something here? Or is that not implemented yet?

Thank you everyone who made this possible and for any help!

fynngodau:
@agentorange96 Is your work profile set up completely such that you have a managed work account showing up under Accounts in the settings, under the Work section? If not, ensure it is allowed to set up a work account in the microG settings and do the work setup from scratch.

For /e/OS' stock launcher, it is additionally necessary to restart the device after setting up due to https://gitlab.e.foundation/e/backlog/-/issues/8275.

agentorange96:
@fynngodau Thank you for the response!
It is in the accounts section under work. I did enable that setting in MicroG before any of this, including uninstalling Intune Company Portal and removing my work profile.

I am using Nova Launcher, but also I don't seem to have that issue. If I try and launch MicroG Companion it goes to MicroG settings and gives me a toast about how there's no activity for MicroG Companion itself. (I don't remember exact text)

I do have an issue where I get stuck at "Adding your device to Company Portal" and it seems MicroG Companion is trying to launch something but fails and instantly goes back to Intune company Portal to the screen it says that at. I gave this a lot more time this time before leaving this comment, but it's still stuck.

If I close Company Portal and restart it, it's logged in. I can tap my device and register it with my org and then everything seems to be set up fine. But no way to install apps. I do get a banner at the bottom saying to use the company version of the play store, and tapping it attempts to load MicroG Companion but again it redirects to MicroG settings.

Thank you again for your help!

fynngodau:
@agentorange96 Intune is not supposed to get stuck during setup; are you positive that the Google kind of Work account is being added to your accounts screen as well, not just Intune? I.e. I think there should be two accounts, not just one. (Of course you can send a screenshot for me to check.) Because it sounds like it is stuck at creating this account, that would explain this behavior: Intune will never retry this if it fails so you can end up correctly set up but without a work Google account configured. And you need a work Google account for the Work app store to show up in the first place!

In any case: To anaylze what went wrong a log of what is happening would be most helpful. However beware that it can contain personal data, feel free to email me at the address from fynngodau.de/imprint

agentorange96:
@fynngodau It only shows one account and it is labeled "Work account" of the type that shows a blue and white icon depicting an ID on a lanyard. So not Google. My company does not use Google accounts. Would this be a blocker? I think given this you are on the right track. Am I supposed to set up a work profile prior to adding my work account? I was letting Intune's setup do that for me. Also I'm not sure where to find the wanted log, but I can send it if needed. Thanks!

fynngodau:
@agentorange96 Sorry for the delay.

> My company does not use Google accounts. Would this be a blocker?

Despite this being mostly invisible to the user during normal use, there is (from what I have seen) always a pseudonymous Google account associated with your Microsoft account for Intune that microG would sign in to automatically. This account is the one that will later allow you to install apps 

> Am I supposed to set up a work profile prior to adding my work account? I was letting Intune's setup do that for me.

Your behavior is correct – it is not possible / compatible to do this manually / in some other way.

> Also I'm not sure where to find the wanted log, but I can send it if needed.

Please trigger the setup of the work profile again (including waiting the delay), then go to Developer Options → Bug report. The bug report will then be generated and you can send it to my email address or inspect it manually.

agentorange96:
@fynngodau No problem! I'd intended to email you directly, but hadn't gotten around to it. Although I'm finding it harder to remember meetings without my phone reminders, so perhaps it's something I should put some priority on. Thank you for the explanation! I'll get those logs, give them a look over both in case it makes anything obvious to me and to remove anything sensitive and then send them your way if needed. Thank you again!

rotheredspecial:
Is it possible that the basic issue is still unresolved: not being able to create a work profile in InTune so the device can be given access to a company's msoft 365 account?

agentorange96:
> Is it possible that the basic issue is still unresolved: not being able to create a work profile in InTune so the device can be given access to a company's msoft 365 account?

How far are you able to get? I am able to connect fine but I don't have a work version of an app store so it's kinda useless for me. I was supposed to collect logs a long while back but wasn't totally sure how or was worried about what data may be in them, so I kinda put that on the backburner.

rotheredspecial:
After installing the app using the APK that I downloaded from the MS site directly I have no issue. The apps can be obtained from inside the portal app, in the hamburger menu I see "get apps"

agentorange96:
Ah, see for me, "get apps" brings me to the "App Compatability" page of the work profile version of MicroG settings. Glad you're able to get it working for your org though!

agentorange96:
Okay, so I see that if I go to the work account settings page and try to add a Google account, I get an ActivityNotFound exception. So maybe /e/OS is unable to handle adding a Google account. Which ROM are you using?

fynngodau:
After successful setup with Intune there should be a **Google Work** account automatically added to the work profile. Adding an account manually is not necessary to obtain apps. Adding Google Work account manually is not possible. If (and only if) the Google Work account is added, a new app called Work App Store is added (reboot if you don't see it).

If you don't have the Google Work account I suggest you make sure you are on microG version 0.3.14 at least and then remove Intune and re-add. If still not working, please file a new issue.

Setup has been successfully performed with a recent version of /e/OS.

agentorange96:
AHH! The reboot fixed it! Thank you!
Classic solution.

fynngodau:
@agentorange96 On /e/OS' launcher the icon not showing up before reboot is tracked here: https://gitlab.e.foundation/e/backlog/-/work_items/8275

If you can confirm there that you encountered the issue on latest /e/OS, I will reopen the issue.

agentorange96:
I don't _think_ it's that. I'm not using Bliss, and even tapping "get apps" didn't work. So it seems like the activity wasn't even enabled until the reboot. But it definitely is after. I can access it through the Company Portal app or through Nova Launcher directly.

Although if I'm fundamentally misunderstanding the nature of the issue linked, I'd be happy to report it there.

fynngodau:
I don't remember adding anything that checks specifically after reboot, so I don't know why the behavior is as descrined, i.e. with your extended description I am actually surprised that a reboot helped. If not using stock launcher, the issue I linked indeed is of no relevance.

rotheredspecial:
I don't have a Google account added to microG as it keeps chasing issues. MicroG itself crashes, and some other things happened (though I don't recall what). Could you without the Google account and see if that works?

20 Jun 2026 12:45:27 fynngodau ***@***.***>:

PR #2561, opened 2024-09-26
Use the correct name case for beaconDB

flame-0:
From their website:
beaconDB


joelkoen:
Discussed in the BeaconDB matrix room, `BeaconDB` is fine :)

PR #2571, opened 2024-10-08
Fix vending build info initialization parameter is null

DaVinci9196:
Reproduction steps:
1.Install vending without installing mg
2.Find an APP to trigger IAP and trigger build info initialization
3.install mg
4.IAP problem found

The problem is that the build info of vending depends on mg. If mg is not installed, the parameters in the build are null. The build info is initialized only once, so it will not be initialized again after mg is installed, resulting in the parameters in the build being null all the time.

### DIFF
--- a/play-services-base/core/src/main/kotlin/org/microg/gms/profile/ProfileManager.kt
+++ b/play-services-base/core/src/main/kotlin/org/microg/gms/profile/ProfileManager.kt
@@ -325,8 +325,10 @@ object ProfileManager {
                 Log.v(TAG, "<data key=\"$key\" value=\"$value\" />")
             }
         }
-        applyProfileData(profileData)
-        activeProfile = PROFILE_REMOTE
+        if (profileData.isNotEmpty()) {
+            applyProfileData(profileData)
+            activeProfile = PROFILE_REMOTE
+        }
     }
 
     fun getProfileName(context: Context, profile: String): String? = getProfileName { getProfileXml(context, profile) }
@@ -375,6 +377,11 @@ object ProfileManager {
             Log.w(TAG, e)
             return false
         }
+    }
+
+    @JvmStatic
+    fun resetActiveProfile() {
+        activeProfile = null
     }
 
     @JvmStatic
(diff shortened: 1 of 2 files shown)


mar-v-in:
Shouldn't we rather make sure to never call `ProfileManager.ensureInitialized()` before it's needed? It's never needed when the main services app is not installed as we can't check if relevant services are enabled.

Also, the current code doesn't assume the remote profile to be applied if the applied dataset is empty or an exception occurs while fetching it. So it's unclear to me how we can run into the issue you mention.

Your current code also checks a random subset of profile fields to be unset. Why isn't a single one enough and why did you choose those and not others. The check seems a little fuzzy to me.

DaVinci9196:
> Shouldn't we rather make sure to never call `ProfileManager.ensureInitialized()` before it's needed? It's never needed when the main services app is not installed as we can't check if relevant services are enabled.
> 
> Also, the current code doesn't assume the remote profile to be applied if the applied dataset is empty or an exception occurs while fetching it. So it's unclear to me how we can run into the issue you mention.
> 
> Your current code also checks a random subset of profile fields to be unset. Why isn't a single one enough and why did you choose those and not others. The check seems a little fuzzy to me.

You can try this APP
app:SoundCloud
packagename:com.soundcloud.android
1.Install vending without installing mg
2.open SoundCloud and click “upgrade”.   Tips: There is a problem
3.install mg
4.open SoundCloud and click “upgrade”.   Tips: There is a problem
5.Removing the app from the background and re-entering it also has the same problem

ale5000-git:
I think "vending" should display an error and not initialize anything if microg services isn't installed as it isn't a supported configuration.

DaVinci9196:
> I think "vending" should display an error and not initialize anything if microg services isn't installed as it isn't a supported configuration.

If mg is not installed, the third-party APP will prompt. If we prompt again, it will be repeated and affect the user experience.

ale5000-git:
We can (in my opinion):
- Display a not-dismissible notification (it keep track of installed app and disappear only if microg services is installed)
- A message in the logcat (every time)
- Skip initialization if microg services is installed (every time)

PR #2582, opened 2024-10-15
Auth: Fix incomplete silent sign in for games

DaVinci9196:
Fixed the issue that Limbo game cannot be logged in.
e.g. com.playdead.limbo.full
e.g. FRAG <com.ohbibi.fps>


play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInService.kt
```diff
@@ -77,10 +79,14 @@ class AuthSignInServiceImpl(
...
-                if (account != null && options?.isForceCodeForRefreshToken != true && options?.includeUnacceptableScope != true) {
+                val account = account 
+                    ?: options?.account 
+                    ?: SignInConfigurationService.getDefaultAccount(context, packageName) 
+                    ?: AccountManager.get(context).getAccountsByType(AuthConstants.DEFAULT_ACCOUNT_TYPE).firstOrNull()
+                Log.d(TAG, "silentSignIn: account -> ${account?.name}")
+                if (account != null && options?.isForceCodeForRefreshToken != true) {
                     if (getOAuthManager(context, packageName, options, account).isPermitted || AuthPrefs.isTrustGooglePermitted(context)) {
-                        val googleSignInAccount = performSignIn(context, packageName, options, account)
+                        val googleSignInAccount = performSignIn(context, packageName, options, account, true)
```
mar-v-in, line 89: Silent sign-in shouldn't be considered permitted by default (permitted means that the user actively agreed to give the app access to their account)

DaVinci9196: Modified. When there is no permission, a pop-up window will be displayed to prompt the authorization content. It depends on the background pop-up window permission.

play-services-core/src/main/AndroidManifest.xml
```diff
@@ -570,8 +570,8 @@
 
         <activity
             android:name="org.microg.gms.auth.consent.ConsentSignInActivity"
-            android:process=":ui"
```
mar-v-in, line 573: You should keep the `:ui` process, otherwise the ui being closed can break things in the main services.

DaVinci9196: Modified, but only UI thread processing will cause Message exception errors, so bundle is used to pass data

mar-v-in (approved):

PR #2583, opened 2024-10-16
Fix issues with Family Link

DaVinci9196:
Fixed the issue that when creating a child account in Family Link, clicking Continue does not respond.
 com.google.android.apps.kids.familylink


ale5000-git:
@DaVinci9196
I wonder if when you had free time you could look at reimlementing this: `com.google.android.gms/.kids.account.receiver.ProfileOwnerReceiver`

PR #2585, opened 2024-10-16
Fix issues related to AuthSignIn page

DaVinci9196:
Fixed the issue that the account selection page is not fully displayed in landscape mode.
image


mar-v-in:
I think it would be better to have two layout files and have the resource manager pick the appropriate layout file based on viewport size.

DaVinci9196:
> I think it would be better to have two layout files and have the resource manager pick the appropriate layout file based on viewport size.

Changed

play-services-core/src/main/res/layout/signin_picker.xml
mar-v-in, line 1: Please undo the changes in this file (as they're all just formatting, it showing up in the history will only cause confusion)

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
```diff
@@ -53,7 +53,7 @@ class AuthSignInActivity : AppCompatActivity() {
             intent?.extras?.also { it.classLoader = SignInConfiguration::class.java.classLoader }?.getParcelable<SignInConfiguration>("config")
         }.getOrNull()
 
-    private val Int.px: Int get() = (this * resources.displayMetrics.density).toInt()
+    private lateinit var accountManager: AccountManager
```
mar-v-in, line 56: It would be better to not keep this reference but instead just re-invoke `getSystemService<AccountManager>()` in `initView()`.

play-services-core/src/main/kotlin/org/microg/gms/auth/signin/AuthSignInActivity.kt
```diff
@@ -64,8 +64,12 @@ class AuthSignInActivity : AppCompatActivity() {
         val packageName = config?.packageName
         if (packageName == null || (packageName != callingActivity?.packageName && callingActivity?.packageName != this.packageName))
             return finishResult(CommonStatusCodes.DEVELOPER_ERROR, "package name mismatch")
-        val accountManager = getSystemService<AccountManager>() ?: return finishResult(CommonStatusCodes.INTERNAL_ERROR, "No account manager")
+        accountManager = getSystemService<AccountManager>() ?: return finishResult(CommonStatusCodes.INTERNAL_ERROR, "No account manager")
 
+        initView()
+    }
+
+    private fun initView(){
```
mar-v-in, line 72: There's a space missing between `)` and `{`.

play-services-core/src/main/res/layout-land/signin_picker.xml
mar-v-in, line 1: I think it would be better to base the decision on the screen width (using `layout-w600dp`) instead of orientation (`layout-land`) as orientation doesn't tell anything about screen width and would only confuse users if e.g. the screen is almost square (like on the inner screen of some foldables)

mar-v-in (changes requested):

DaVinci9196:
@mar-v-in Modified

mar-v-in (approved):

PR #2620, opened 2024-11-12
Fix the problem of not being able to bind a PayPal account

DaVinci9196:


mar-v-in:
Using `CustomTabsIntent` means moving authentication tokens and cookies into the current browser, meaning you sign in the browser even if the user doesn't want that.

DaVinci9196:
> Using `CustomTabsIntent` means moving authentication tokens and cookies into the current browser, meaning you sign in the browser into a Google account even if the user doesn't want that.



> Using `CustomTabsIntent` means moving authentication tokens and cookies into the current browser, meaning you sign in the browser into a Google account even if the user doesn't want that.

Yes, but during use, I found that when I jump to the web page to add a payment method and view other content, an error page will be displayed.

mar-v-in:
You mean you can't add a PayPal account inside the WebView? There is a pending [TODO in WebViewHelper to redirect some navigations to the browser](https://github.com/microg/GmsCore/blob/master/vending-app/src/main/java/org/microg/vending/billing/ui/WebViewHelper.kt#L49), I guess that's what needs to be implemented correctly. It even shows a message shortly before the PayPal screen hinting that it should open an external browser:
ef53a37_mSEyeYGtSN-W6qCjq7MLiQ

DaVinci9196:
> You mean you can't add a PayPal account inside the WebView? There is a pending [TODO in WebViewHelper to redirect some navigations to the browser](https://github.com/microg/GmsCore/blob/master/vending-app/src/main/java/org/microg/vending/billing/ui/WebViewHelper.kt#L49), I guess that's what needs to be implemented correctly. It even shows a message shortly before the PayPal screen hinting that it should open an external browser: ef53a37_mSEyeYGtSN-W6qCjq7MLiQ

I tried it, and when I jumped to the screenshot page, shouldOverrideUrlLoading did not print the URL of the relevant features, so I could not jump to it in a targeted way. Moreover, I found that when using the original webview, when I jumped to add a payment method and clicked the tab on the web page to view the app details on other pages, the wrong content would appear.

DaVinci9196:
@mar-v-in I tried to click Add PayPal in the webview and then jump to the browser for subsequent operations, but when I clicked Return to the merchant (an error pop-up window appeared with account restriction), I could not return to the page before the jump (the page in the screenshot above).

PR #2621, opened 2024-11-12
Added Device Sync request

DaVinci9196:
Added a device information synchronization request after logging in to the account to solve the problem of missing data in some request interfaces that rely on device information. 
For example, when there is no device synchronization in the /assetModuleDelivery request in https://github.com/microg/GmsCore/pull/2506 , some games lack resource package data returned.


play-services-basement/src/main/java/org/microg/gms/common/Constants.java
```diff
@@ -28,4 +28,5 @@ public class Constants {
     public static final String MICROG_PACKAGE_SIGNATURE_SHA1 = "10321bd893f69af97f7573aafe9de1dc0901f3a1";
     @Deprecated
     public static final int MAX_REFERENCE_VERSION = GMS_VERSION_CODE;
+    public static final String GP_PACKAGE_NAME = "com.android.vending";
```
ale5000-git, line 31: Why not using `VENDING_PACKAGE_NAME` from [here](https://github.com/microg/GmsCore/blob/master/vending-app/src/main/java/org/microg/vending/billing/Constants.kt#L13)?

ale5000-git: I didn't understand your reply.
In my opinion shouldn't there be more than one costant for "com.android.vending" and also the name GP_PACKAGE_NAME isn't intuitive.

DaVinci9196: > `VENDING_PACKAGE_NAME`

OK, got it.

vending-app/src/main/kotlin/com/google/android/finsky/DeviceSyncInfo.kt
```diff
@@ -0,0 +1,612 @@
...
+import android.annotation.SuppressLint
+import android.app.ActivityManager
+import android.app.admin.DevicePolicyManager
+import android.content.Context
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
+import android.content.res.Configuration
+import android.graphics.Point
+import android.opengl.GLES10
+import android.os.Build
```
mar-v-in, line 19: Instead of using real device data, can you please use data from the ProfileManager where possible?

vending-app/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,9 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="com.google.android.gms.permission.READ_SETTINGS" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.READ_PRIVILEGED_PHONE_STATE"
+        tools:ignore="ProtectedPermissions" />
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
mar-v-in, line 24: I don't think we really need these permissions. If you any data you sent during device sync requires these permissions, maybe try to see if you can stub them out instead.

vending-app/src/main/kotlin/com/google/android/finsky/accounts/impl/AccountsChangedReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+                val account = AccountManager.get(context).getAccountsByType(AuthConstants.DEFAULT_ACCOUNT_TYPE).firstOrNull {
+                    it.name == accountName
+                } ?: throw RuntimeException("account is null")
+                val oauthToken = account.let {
+                    AccountManager.get(context).getAuthToken(it, AUTH_TOKEN_SCOPE, false).getString(AccountManager.KEY_AUTHTOKEN)
+                } ?: throw RuntimeException("oauthToken is null")
+                ProfileManager.ensureInitialized(context)
+                val androidId = GServices.getString(context.contentResolver, "android_id", "0")?.toLong() ?: 1
+                HttpClient(context).post(
+                    url = "https://play-fe.googleapis.com/fdfe/sync",
```
mar-v-in, line 50: Please add the URL constant to `GooglePlayApi` class.

mar-v-in (changes requested):

vending-app/src/main/kotlin/com/google/android/finsky/DeviceSyncInfo.kt
DaVinci9196: Modified

vending-app/src/main/kotlin/com/google/android/finsky/accounts/impl/AccountsChangedReceiver.kt
DaVinci9196: Modified

0813950040 (approved):

play-services-base/core/src/main/kotlin/org/microg/gms/profile/Build.kt
```diff
@@ -95,6 +95,23 @@ object Build {
...
+        const val O = 26                    // Android 8.0 (Oreo)
+        const val O_MR1 = 27                // Android 8.1
+        const val P = 28                    // Android 9 (Pie)
+        const val Q = 29                    // Android 10
+        const val R = 30                    // Android 11
+        const val S = 31                    // Android 12
+        const val S_V2 = 32                 // Android 12L
+        const val TIRAMISU = 33             // Android 13
+        const val UPSIDE_DOWN_CAKE = 34     // Android 14
+    }
```
ale5000-git, line 113: Since there are almost all versions we may as well add also these: 19, 20, 35

mar-v-in: I don't think we need to copy them at all ;)

ale5000-git:
@DaVinci9196
This PR may be very useful for many people but others may not like it.
Could you please add an option to enable/disable Device Sync request?

ale5000-git:
@mar-v-in
Maybe we need to add IMEI and IMSI to Device profile to generate random values at first time but consistent between requests.

mar-v-in:
> Maybe we need to add IMEI and IMSI to Device profile code to generate random values at first time but consistent between requests.

I was considering the same, but in most cases IMEI and IMSI are not required anyway (because of wifi-only tablets). But yes, in the long run that would be a good idea.

ale5000-git:
In this PR it send the real IMSI that could be used to identify the SIM.

mar-v-in:
@ale5000-git The current code puts the android id where probably the IMSI belongs. There is no calling of `getSubscriberId`, but also it wouldn't have permission to read it anyway.

ale5000-git:
> I was considering the same, but in most cases IMEI and IMSI are not required anyway (because of wifi-only tablets). But yes, in the long run that would be a good idea.

Devices without SIM slot are rare I think and devices with SIM slot but empty are more common but still have IMEI (device with 4 SIM slots have 4 IMEI even if the slots are unused).
So I think we should at least have slot number in the profile xml and the device profile code should fake a corresponding numbert of IMEI.

DaVinci9196:
Accepted.

vending-app/src/main/kotlin/com/google/android/finsky/DeviceSyncInfo.kt
```diff
@@ -21,9 +21,9 @@ import android.text.TextUtils
 import android.util.Base64
 import android.util.DisplayMetrics
 import android.util.Log
+import android.os.Build
 import android.view.WindowManager
 import org.microg.gms.common.Constants
-import org.microg.gms.profile.Build
 import java.security.MessageDigest
```
ale5000-git: Loo at this: https://github.com/microg/GmsCore/commit/56b468ecbb450844c3e41a2b7d85ff979d98734b

vending-app/src/main/kotlin/com/google/android/finsky/DeviceSyncInfo.kt
```diff
@@ -0,0 +1,584 @@
...
+                    add(featureInfoProto)
+                }
+                builder.featureNames = builder.featureNames.toMutableList().apply {
+                    add(featureInfoProto.name!!)
+                }
+            }
+        }
+        if (SDK_INT >= 21) {
+            builder.supportedAbi(listOf(*Build.SUPPORTED_ABIS))
+        }
```
ale5000-git, line 158: @mar-v-in
Since this may be used to download things (and we don't want to download things incompatible with the device) in case we simulate a device >= 21 from a device < 21 we should generate a SUPPORTED_ABIS list based on the real device instead of the spoofed one.
What do you think?

mar-v-in: If the profile diverts from the actual device too much, this would indeed likely cause issues, but that is to be expected elsewhere as well. I think we can keep it as is and revisit later if and how we want to adjust this device sync feature - as likely we want to extend the device profiles to cover more of the elements that we send here.

mar-v-in (approved):

PR #2622, opened 2024-11-13
Fixed the pop-up window issue on Twitch

DaVinci9196:
image


mar-v-in:
According to all documentation available, `install_referrer` is supposed to return an URL. I also don't see what's wrong in your screenshot. Can you describe the issue in more detail?

DaVinci9196:
> According to all documentation available, `install_referrer` is supposed to return an URL. I also don't see what's wrong in your screenshot. Can you describe the issue in more detail?

After the application is cold started, this pop-up window does not exist on the gms phone. Through the corresponding underlying code, we can see that the install_referrer field does not contain a specific domain name. In addition, it was found at the SDK level that no splicing was performed after obtaining this field.

PR #2634, opened 2024-11-18
Google Maps location information sharing page settings button click n…

DaVinci9196:
Google Maps location information sharing page settings button click no response

app name:google maps
packagename:com.google.android.apps.maps
Steps to reproduce：

1. open google maps
2. click google account avatar
3. click location sharing
4. click "setting icon"
5. no response

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
@@ -122,6 +122,7 @@ private val ACTION_TO_SCREEN_ID = hashMapOf(
     ACTION_ACCOUNT_PREFERENCES_SETTINGS to 1,
     ACTION_SECURITY_SETTINGS to 10006,
     ACTION_PRIVACY_SETTINGS to 10004,
+    ACTION_LOCATION_SHARING to 210,
 )
 
 class MainActivity : AppCompatActivity() {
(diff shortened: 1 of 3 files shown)


play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
```diff
@@ -134,7 +135,7 @@ class MainActivity : AppCompatActivity() {
         Log.d(TAG, "Invoked with ${intent.action} and extras $extras")
         super.onCreate(savedInstanceState)
 
-        val screenId = intent?.getIntExtra(EXTRA_SCREEN_ID, -1).takeIf { it != -1 } ?: ACTION_TO_SCREEN_ID[intent.action] ?: 1
+        val screenId = ACTION_TO_SCREEN_ID[intent.action] ?: intent?.getIntExtra(EXTRA_SCREEN_ID, -1).takeIf { it != -1 } ?: 1
```
mar-v-in, line 138: What is the screen id sent by Google Maps? Giving priority to the action-based screen id will mean that we might not show the appropriate screen in some cases, so maybe there's a better way to handle this.

DaVinci9196: screen id is 1

mar-v-in: I adjusted the code so that only actions that specify a screen id other than 1 have preference over the app provided screen id. As this only applies to 3 actions, we should be reasonable safe now.

PR #2652, opened 2024-11-25
Password: switch the webview to the browser

DaVinci9196:
Fixed the issue on the account and password management settings page where clicking Import Password did not work.
Path: Password management page -> Enter the settings page in the upper right corner -> Click Import password


mar-v-in:
Do not open `weblogin` URLs in the browser directly. This will set cookies to sign in the user in the browser that the user might not want or expect. Best would be if you can figure out why the WebView does not work (you can try to connect a remote chrome debugger to the webview and then see if there are any issues logged), alternatively you need to at least ask the user to confirm they are going to be logged in before redirecting them into the signed in browser.

DaVinci9196:
> Do not open `weblogin` URLs in the browser directly. This will set cookies to sign in the user in the browser that the user might not want or expect. Best would be if you can figure out why the WebView does not work (you can try to connect a remote chrome debugger to the webview and then see if there are any issues logged), alternatively you need to at least ask the user to confirm they are going to be logged in before redirecting them into the signed in browser.

Optimized

PR #2660, opened 2024-12-03
Auth: Optimize quick login user experience

DaVinci9196:
Optimization: 
1. When the Google account is not logged in, a pop-up window will prompt the user to log in. 
2. When the logged-in Google account authentication fails, the user will be prompted with a problem.


DaVinci9196:
Screenshot_20241203_154927

ale5000-git:
@mar-v-in
@DaVinci9196
I wonder if we should put "notifyGamePlayed()" under an option.
Similar to: `Automatically add free apps to library`.

PR #2662, opened 2024-12-04
DynamicLink: Added network connection conversion processing

DaVinci9196:
The test found that in two scenarios, obtaining the DynamicLinksService getDynamicLink method would cause the application to display abnormally.
Scenario 1: Clicking the link shared by SoundCloud on WhatsApp will result in a black screen.
Scenario 2: When opening the link shared by Google Photos, WhatsApp will display that the corresponding sharing link cannot be found.

Improve the createShortDynamicLink method to fix the problem of no response when clicking share in the app.
e.g. com.cyberdavinci.gptkeyboard


play-services-base/core/build.gradle
```diff
@@ -10,6 +10,9 @@ apply plugin: 'signing'
 
 dependencies {
     api project(':play-services-basement-ktx')
+    api project(':play-services-core-proto')
```
mar-v-in, line 13: `play-services-base-core` should not depend on `play-services-core-proto`. proto specifications in `play-services-core-proto` are only meant to be consumed in `play-services-core`. Better keep the `datamixer.proto` and `DynamicLinkUtils` in `play-services-appinvite/core`, where they can be accessed from `play-services-core` if needed.

mar-v-in (changes requested):

play-services-base/core/build.gradle
DaVinci9196: Optimized

mar-v-in (approved):

PR #2664, opened 2024-12-05
DynamiteModule: Added <init> method

DaVinci9196:
Added a new DynamiteModule initialization method to initialize some necessary content. 
For example, this commit can effectively avoid the ANR problem of the new version of X on some mobile devices.

### DIFF
--- a/play-services-ads/core/src/main/java/com/google/android/gms/dynamite/descriptors/com/google/android/gms/ads/dynamite/ModuleDescriptor.java
+++ b/play-services-ads/core/src/main/java/com/google/android/gms/dynamite/descriptors/com/google/android/gms/ads/dynamite/ModuleDescriptor.java
@@ -7,7 +7,6 @@ package com.google.android.gms.dynamite.descriptors.com.google.android.gms.ads.d
 
 import android.content.Context;
 import android.content.ContextWrapper;
-import android.util.Log;
 import android.webkit.WebSettings;
 import androidx.annotation.Keep;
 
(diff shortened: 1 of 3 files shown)


play-services-ads/core/src/main/java/com/google/android/gms/dynamite/descriptors/com/google/android/gms/ads/dynamite/ModuleDescriptor.java
```diff
@@ -16,9 +15,17 @@ public class ModuleDescriptor {
...
+     * which may result in deadlocks when invoked from any other thread. This only happens with microG,
+     * because we don't use the highly privileged SELinux Sandbox that regular Play Services uses
+     * (which allows apps to read the user-agent from Play Services instead of the WebView). To prevent
+     * the issue we pre-emptively initialize the WebView.
+     */
+    public static void init(Context context) {
+        if (context instanceof ContextWrapper) {
+            context = ((ContextWrapper) context).getBaseContext();
+        }
+        WebSettings.getDefaultUserAgent(context);
```
ale5000-git, line 29: @mar-v-in
If I'm not wrong the User-Agent is changed by the profile.
Don't we need to run `ProfileManager.ensureInitialized`?

mar-v-in: This is running inside a dynamite module in the calling app's process. We can't change how the code inside the app accesses the user agent and we also don't have access to the profile manager from here. It also doesn't affect the user agent itself, it just makes sure the WebView is initialized when the client library code inside the app tries to access it later.

PR #2698, opened 2024-12-24
Auth: Fixed the authorization issue

DaVinci9196:
Fixed the issue that the unauthorized account could not log in to the application normally.
e.g. Among Us <com.innersloth.spacemafia> eFootball™ <jp.konami.pesam>


ale5000-git:
@mar-v-in
@DaVinci9196
I wonder if we should put "notifyGamePlayed()" under an option.
Similar to: "Automatically add free apps to library"

PR #2703, opened 2024-12-26
Minor layout/strings fixes and adjustments to the settings app

### LINKED ISSUE #2271: Small UI regression in 0.3.1: preference titles are not multi-line anymore
**Describe the bug**
After the UI design upgrade in 0.3.1, preference titles are now constrained to a single line, unlike previous versions of microG.
This causes some preference titles to be truncated, making it harder to understand what they do, especially on phones with smaller screens or in languages with longer strings than English.

**To Reproduce**
Some truncated titles can be observed in the Google Accounts section, if the screen is not large enough.

**Expected behavior**
Preference titles should be able to span on multiple lines when needed.

**Screenshots (Italian)**
| Before (0.3.0) | After (0.3.1) |
|--------|--------|
| Screenshot_20240327-215503_microG_Services | Screenshot_20240327-234733_microG_Services | 

**System**
Android Version: reproduced on 10 & 11
Custom ROM: LineageOS

Fs00:
This PR includes some fixes for a few minor issues I noticed on Weblate and while using the microG settings app:

- removed some leftover strings related to Mozilla Location Service and the old "custom service URL" dialog (I did this for all languages), so that people don't waste time translating them on Weblate
- replaced the Chinese colon character with the correct (English/latin) one in the `games_leaderboards_score_label` string
- tweaked the app location details screen so that preferences use all the available screen width (using `app:iconSpaceReserved="false"`) and the hardcoded "Last reported location" string can be translated
- fixed the location service selection dialog so that the custom service URL can be set on small screens. Previously, part of the UI would have been cut when the screen wasn't tall enough, as you can see below:

https://github.com/user-attachments/assets/c3c5e819-0a54-4714-bf79-006a16846b76

This is instead how it behaves now, after wrapping the layout in a ScrollView:

https://github.com/user-attachments/assets/178a7ba7-11c5-436b-b0ec-649197d6c676

In the layout, I've removed the `<requestFocus />` tag because it made the ScrollView jump directly to the text field, potentially hiding the initial portion of text that the user is supposed to read.

- Last but not least, I've managed to find a way to allow preference titles to be multi-line by default, effectively restoring how they behaved before v0.3.1. This helps a lot with non-English languages that have long translated strings that cannot be easily shortened to a single line (Italian is one of them).
Fixes #2271


mar-v-in (approved):

ILoveOpenSourceApplications:
> This is instead how it behaves now, after wrapping the layout in a ScrollView:
> 
>  [custom_location_service_after.mp4](https://github.com/user-attachments/assets/3b95b189-1c2e-405b-b85c-0ecd2c20b1d9)

From the attached video clipping, the text `Custom service URL` clips into the text box. That should be fixed as well right?

Fs00:
Ideally, yes. The bug was already present before this PR though, and I didn't have (and currently don't) have enough time to investigate to find the proper way to fix it. I'm open to suggestions 🙌

mar-v-in:
Thanks.

PR #2737, opened 2025-01-08
HmsMap: Uber route drift modification

DaVinci9196:
The judgment conditions before the change can no longer be triggered normally


ale5000-git:
I think "contains" is too much generic, isn't it better startsWith?

DaVinci9196:
> I think "contains" is too much generic, isn't it better startsWith?

Modified, thanks

PR #2740, opened 2025-01-10
PI: Fix Failed to generate key pair and related crashes

DaVinci9196:
java.security.ProviderException: Failed to generate key pair.
                                                                                                    	at android.security.keystore2.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPairHelper(AndroidKeyStoreKeyPairGeneratorSpi.java:620)
                                                                                                    	at android.security.keystore2.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair(AndroidKeyStoreKeyPairGeneratorSpi.java:545)
                                                                                                    	at java.security.KeyPairGenerator$Delegate.generateKeyPair(KeyPairGenerator.java:727)

### DIFF
--- a/vending-app/src/main/kotlin/com/google/android/finsky/IntegrityExtensions.kt
+++ b/vending-app/src/main/kotlin/com/google/android/finsky/IntegrityExtensions.kt
@@ -46,9 +46,11 @@ import java.io.File
 import java.io.FileInputStream
 import java.io.FileOutputStream
 import java.nio.ByteBuffer
+import java.security.KeyPair
 import java.security.KeyPairGenerator
 import java.security.KeyStore
 import java.security.MessageDigest
+import java.security.ProviderException
 import java.security.spec.ECGenParameterSpec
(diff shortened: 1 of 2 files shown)


ale5000-git:
@DaVinci9196
Look at this: https://github.com/microg/GmsCore/pull/2199/files
Apparently some devices fails when attestation challenge is set, I wonder if it make sense does the same as PR #2199 also in Play Integrity (retry without it if it fails with ProviderException).

DaVinci9196:
> @DaVinci9196 Look at this: https://github.com/microg/GmsCore/pull/2199/files Apparently some devices fails when attestation challenge is set, I wonder if it make sense to do the same as in PR #2199 also in Play Integrity (retry without attestation challenge if it fails with ProviderException).

Very good idea, thanks

flamme-demon:
Hi, team,


Without any pressure, do you plan to release a version soon that will correct this behavior?

Issue #2727, opened 2025-01-04
Applications trying to change the WebView data directory suffix crash after Dynamite preinit commit


LucentW:
**Describe the bug**
Seemingly due to commit ad7e79ecdb5efb9a0fe185a066bba7d792bcf8a0, applications which embed ads and try to change the data directory suffix fail with a IllegalStateException due to WebView being already initialised.

Impacted application in my case is AquaMail (I have a lifetime Pro license, but it seems to still try to load the ad module anyways)

**To Reproduce**
1. Open an application that tries to change the directory suffix
2. Observe it crash

**Expected behavior**
Application should be able to initialize the WebView on its own, avoiding the issue at hand.

**Logs**
```
time: 1735961600817
msg: java.lang.IllegalStateException: Can't set data directory suffix: WebView already initialized
stacktrace: java.lang.RuntimeException: Unable to create application org.kman.AquaMail.core.AquaMailApplication: java.lang.IllegalStateException: Can't set data directory suffix: WebView already initialized
	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7592)
	at android.app.ActivityThread.-$$Nest$mhandleBindApplication(Unknown Source:0)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2449)
	at android.os.Handler.dispatchMessage(Handler.java:109)
	at android.os.Looper.loopOnce(Looper.java:232)
	at android.os.Looper.loop(Looper.java:317)
	at android.app.ActivityThread.main(ActivityThread.java:8787)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:594)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:871)
Caused by: java.lang.IllegalStateException: Can't set data directory suffix: WebView already initialized
	at android.webkit.WebViewFactory.setDataDirectorySuffix(WebViewFactory.java:240)
	at android.webkit.WebView.setDataDirectorySuffix(WebView.java:2069)
	at com.microsoft.identity.common.internal.ui.webview.c.a(SourceFile:1)
	at org.kman.AquaMail.core.AquaMailApplication.b(SourceFile:41)
	at org.kman.AquaMail.core.AquaMailApplication.onCreate(SourceFile:13)
	at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1392)
[...]

PR #2746, opened 2025-01-16
Fixed an issue that could cause the app to crash

DaVinci9196:
This fix will probably fix https://github.com/microg/GmsCore/issues/2727. Based on this fix, I did not reproduce the corresponding issue on my test device.
Other known affected applications include: Finance <com.yahoo.mobile.client.android.finance>    Scout Sports <com.bet007.mobile.score.d>

### DIFF
--- a/play-services-ads/core/src/main/java/com/google/android/gms/dynamite/descriptors/com/google/android/gms/ads/dynamite/ModuleDescriptor.java
+++ b/play-services-ads/core/src/main/java/com/google/android/gms/dynamite/descriptors/com/google/android/gms/ads/dynamite/ModuleDescriptor.java
@@ -7,11 +7,16 @@ package com.google.android.gms.dynamite.descriptors.com.google.android.gms.ads.d
 
 import android.content.Context;
 import android.content.ContextWrapper;
+import android.os.Build;
+import android.util.Log;
 import android.webkit.WebSettings;
 import androidx.annotation.Keep;
 
+import java.io.File;
+
 @Keep
 public class ModuleDescriptor {
+    private static final String TAG = "ModuleDescriptor";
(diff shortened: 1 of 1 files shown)


ale5000-git:
@mar-v-in
Is possible to merge this workaround first until there is a better fix to this issue?

mar-v-in:
I believe this fix is going to break a lot of other apps. Notably, setting the web view data directory of apps (where they have not been set before) will mean they might lose access to the actual data directory.

Given the issue was only introduced via #2664 to workaround issues that were not really specific to microG (but also applied if the app was used on any non-Google system) I'm rather tempted to revert #2664 rather than introducing yet another workaround with even higher potential to break things.

ale5000-git:
> I believe this fix is going to break a lot of other apps. Notably, setting the web view data directory of apps (where they have not been set before) will mean they might lose access to the actual data directory.
> 
> Given the issue was only introduced via #2664 to workaround issues that were not really specific to microG (but also applied if the app was used on any non-Google system) I'm rather tempted to revert #2664 rather than introducing yet another workaround with even higher potential to break things.

Quote:
```
     * The ads module might try to access the user agent, requiring initialization on the main thread,
     * which may result in deadlocks when invoked from any other thread. This only happens with microG,
     * because we don't use the highly privileged SELinux Sandbox that regular Play Services uses
     * (which allows apps to read the user-agent from Play Services instead of the WebView). To prevent
     * the issue we pre-emptively initialize the WebView.
```

I don't know this code enough to be sure but isn't possible to maybe "extend" the WebView and make apps use it, so that when an app call methods they are called on our "extended" WebView instead to the real one?
In this way when they try to access the User-Agent when it isn't yet initialized we can just fail without error-out or just return cached data.

PR #2759, opened 2025-01-23
Vending: Add AppUpdateService dummy

DaVinci9196:
Supplement store AppUpdateService to solve the problem that some apps need to return detection results before entering the home page.
e.g. com.hkej.express


rugk:
Is there anything here preventing good from getting merged?

It supposedly fixes https://github.com/microg/GmsCore/issues/2617 so it would be very great for all ePA users and all others in Germany from that insurance which use microG to merge this.

PR #2778, opened 2025-02-07
Maps: Added <GoogleMapZoomInButton> Issue fixed

DaVinci9196:
Fixed the issue that the app cannot use the Maps function normally.
e.g. com.weekendhk.jetsonews


mar-v-in (approved):

ale5000-git:
@DaVinci9196
@mar-v-in

This PR only fix the hw build, but isn't the issue also present in the standard version or I'm wrong?
Look at this: #2794

KyunLFA:
> @DaVinci9196 @mar-v-in
> 
> This PR only fix the `hw` build, but isn't the issue also present in the standard version or I'm wrong? Look at this: #2794

I still face the same issue on the Brazilian ride sharing app 99 (has the exact same UI as that DiDi) even on a recent build I builded from source. Not only is it zoomed out, when the driver accepts the ride the app crashes until the ride finishes. Will try again on latest git and if still facing the issue I will file a dedicated bug report, as the crashing wasn't mentioned in the aforementioned bug report.

PR #2786, opened 2025-02-11
Sms: Improve the problem of incomplete information due to long SMS

DaVinci9196:
When using the following application, I found that the text message I received when registering was not displayed completely through the service, and the text message sent by the application exceeded the length.
e.g. com.boynergrup.hopi

### DIFF
--- a/play-services-auth-api-phone/core/src/main/kotlin/org/microg/gms/auth/phone/SmsRetrieverCore.kt
+++ b/play-services-auth-api-phone/core/src/main/kotlin/org/microg/gms/auth/phone/SmsRetrieverCore.kt
@@ -134,11 +134,11 @@ class SmsRetrieverCore(private val context: Context, override val lifecycle: Lif
         return requests.values.any { it.type == USER_CONSENT }
     }
 
-    private fun sendRetrieverBroadcast(request: SmsRetrieverRequest, message: SmsMessage) {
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

play-services-auth-api-phone/core/src/main/kotlin/org/microg/gms/auth/phone/SmsRetrieverCore.kt
```diff
@@ -246,12 +246,16 @@ class SmsRetrieverCore(private val context: Context, override val lifecycle: Lif
...
+                val messageBodyBuilder = StringBuilder()
+                var senderPhoneNumber: String? = null
                 for (message in messages) {
-                    try {
-                        handleIncomingSmsMessage(message)
-                    } catch (e: Exception) {
-                        Log.w(TAG, "Error handling incoming SMS", e)
-                    }
+                    messageBodyBuilder.append(message.messageBody)
+                    senderPhoneNumber = message.originatingAddress
```
ale5000-git, line 253: @mar-v-in
I don't this code but wouldn't be safer to check that all messages come from the same `originatingAddress`?

mar-v-in: The `SMS_RECEIVED_ACTION` broadcast is sent and handled for a single logical SMS. Long logical SMS for technical reasons are split into multiple PDUs, that result in multiple `SmsMessage` instances in Android. As such there can't be multiple senders of the message objects here.

PR #2787, opened 2025-02-12
MLKit: Add support for ACTION_SCAN_BARCODE activity

DaVinci9196:
Fixed the issue that Google Authenticator cannot use QR code scanning to add functions.


play-services-core/src/main/res/values/strings.xml
```diff
@@ -334,4 +334,9 @@ Please set up a password, PIN, or pattern lock screen."</string>
     <string name="games_delete_snapshot_dialog_ok">OK</string>
     <string name="games_delete_snapshot_error">Deletion failed, please try again later</string>
 
+    <string name="barcode_scanner_brand">Scanned by microG on behalf of %1$s</string>
+    <string name="camera_permission_dialog_button">OK</string>
+    <string name="camera_permission_dialog_message">mircoG services needs to access your device\'s camera to scan a code for %1$s.\n\nTo enable, please grant camera permission to mircoG services in Settings.</string>
```
ale5000-git, line 339: `mircoG` => `microG`

The same error exist also in other languages.

DaVinci9196: I didn't notice this, I've fixed it, thanks.

play-services-core/src/main/AndroidManifest.xml
```diff
@@ -170,6 +170,11 @@
         android:name="android.permission.UPDATE_APP_OPS_STATS"
         tools:ignore="ProtectedPermissions" />
 
+    <uses-permission android:name="android.permission.CAMERA"/>
```
mar-v-in, line 173: This permission should go into the manifest of `play-services-vision/core` where it is used.

play-services-auth-api-phone/core/src/main/kotlin/org/microg/gms/auth/phone/AskPermissionActivity.kt
```diff
@@ -18,7 +18,7 @@ import androidx.core.os.bundleOf
 
 private const val TAG = "AskPermission"
 private const val REQUEST_CODE_PERMISSION = 101
-private val ALLOWED_PERMISSIONS = setOf(Manifest.permission.RECEIVE_SMS, Manifest.permission.READ_CONTACTS)
+private val ALLOWED_PERMISSIONS = setOf(Manifest.permission.RECEIVE_SMS, Manifest.permission.READ_CONTACTS, Manifest.permission.CAMERA)
```
mar-v-in, line 21: If we start using this activity outside of `play-services-auth-api-phone/core`, especially with permissions unrelated to `play-services-auth-api-phone`, it should also be moved elsewhere, e.g. `play-services-base/core`.

play-services-core/src/main/kotlin/org/microg/gms/mlkit/BarcodeScanningActivity.kt
```diff
@@ -0,0 +1,121 @@
...
+            if (activePermissionRequest == null) {
+                val completable = CompletableDeferred<Boolean>()
+                activePermissionRequest = completable
+                completable to activePermissionRequest!!
+            } else {
+                null to activePermissionRequest!!
+            }
+        }
+        if (completable != null) {
+            val intent = Intent(context, AskPermissionActivity::class.java)
```
mar-v-in, line 104: `AskPermissionActivity` is only needed and desirable if you are calling it from within a Service. Here, you already are in the context of an Activity, so you can just call `requestPermissions` and wait for the result on `onRequestPermissionsResult`.

mar-v-in (changes requested):

play-services-core/src/main/AndroidManifest.xml
DaVinci9196: Optimized

play-services-core/src/main/kotlin/org/microg/gms/mlkit/BarcodeScanningActivity.kt
DaVinci9196: Optimized

play-services-auth-api-phone/core/src/main/kotlin/org/microg/gms/auth/phone/AskPermissionActivity.kt
DaVinci9196: Cancel authorization here

ale5000-git:
@DaVinci9196
Look here: https://github.com/microg/GmsCore/pull/2787/files#annotation_32770842822

mar-v-in:
@ale5000-git already working on that. Seems like some broken strings came from Weblate.

mar-v-in (approved):

Issue #2704, opened 2024-12-27
Crash while fetching network location from WIFIonICE moving wifi network


TimoWilken:
**Describe the bug**

When on an ICE train in Germany and connected to the "WIFIonICE" network, I never get a network location from microG that is identical to the train's GPS location (as reported by the [API](https://iceportal.de/api1/rs/status)).

**To Reproduce**

1. Get on an ICE train (or stand next to on in a station?), connect to its WIFIonICE network and log into the network
2. Start a logcat app, e.g. "Logcat Reader", and start recording messages
3. Open any app that requests a location (e.g. a maps app)
5. See error in logcat: "Failed retrieving location for current moving wifi WIFIonICE: java.net.SocketException: Binding socket to network 101 failed: EPERM (Operation not permitted)" (see traceback below)

**Expected behavior**

I expect the maps app to show my location identical with the train's location almost immediately, with an appropriate accuracy (judging by microG code, that would be 100m).

**Screenshots**

N/A

**System**

Android Version: 14 (AP2A.240905.003)
Custom ROM: LineageOS 21-20241211-microG-FP4 (from [lineage.microg.org](https://download.lineage.microg.org/FP4/))

**Additional context**

```
Failed retrieving location for current moving wifi WIFIonICE
java.net.SocketException: Binding socket to network 101 failed: EPERM (Operation not permitted)
	at android.net.Network.bindSocket(Network.java:442)
	at android.net.Network.bindSocket(Network.java:409)
	at android.net.Network$NetworkBoundSocketFactory.createSocket(Network.java:269)
	at com.android.okhttp.internal.io.RealConnection.connect(RealConnection.java:114)
	at com.android.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:186)
	at com.android.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:128)
	at com.android.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
	at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:289)
	at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:232)
	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:465)
	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:411)
[...]

PR #2820, opened 2025-03-16
Location: moving WiFi location needs INTERNET permission

TimoWilken:
Hopefully, this will fix https://github.com/microg/GmsCore/issues/2704.

### DIFF
--- a/play-services-location/core/provider/src/main/AndroidManifest.xml
+++ b/play-services-location/core/provider/src/main/AndroidManifest.xml
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
+    <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.BODY_SENSORS" />
 
     <uses-permission


ale5000-git:
@TimoWilken

Isn't it already fixed by this? https://github.com/microg/GmsCore/commit/fbd84a147baba0718e2db1e95dc5d1ea0b143931

TimoWilken:
Thanks @ale5000-git! I can try to test that the next time I'm on a German train, in a couple weeks' time.

Though I don't have a proxy configured on my phone (only a WireGuard VPN, but that only applies to `10.0.0.*` IPs), and I did have network access through the in-train WiFi. The error in https://github.com/microg/GmsCore/issues/2704 looks to me like it comes from a lower level than TLS/connection errors, but I'm not an Android developer.

ale5000-git:
In the commit description:
```
Bypass ... and disable OCSP/CRL checks
...
OCSP/CRL might be impossible to resolve if in the hotspot is no internet is available (e.g. in planes that have IFE without internet).
```

This is related to the `https` connection, maybe this was the problem.

TimoWilken:
I'm not sure, because in my case, the hotspot *did* have internet available, so GmsCore should've been able to query OCSP and/or fetch CRLs.

To my untrained eye, it doesn't get as far as trying to set up an HTTPS connection, because the `EPERM` happens when setting up the socket.

But anyway, I'll try to test the latest version in two weeks' time.

ale5000-git:
@mar-v-in
I'm not sure about this but maybe we still need the internet permission.

Info: https://developer.android.com/privacy-and-security/local-network-permission?hl=en
Quote:
> Devices on a LAN can be accessed by any app that has the INTERNET permission

Quote:
> If a device's DNS server is on a local network, traffic to / from it (at port 53) doesn't require local network access permission.

So if DNS isn't local we can't access it.

In Android 16+ we probably also need "local network access permission".

mar-v-in:
The INTERNET permission is already granted to microG from other manifests, so this PR does not actually change the resulting `.apk` and will not solve the issue. Regarding the underlying issue, see https://github.com/microg/GmsCore/issues/2704#issuecomment-2748322997

Issue #2769, opened 2025-01-27
Moving WiFi location support for Eurostar trains


TimoWilken:
**Is your feature request related to a problem? Please describe.**

The Eurostar in-train WiFi has a GPS status endpoint, but microg does not currently use it.

**Describe the solution you'd like**

Please support moving WiFi location support for Eurostar trains.

**Describe alternatives you've considered**

GPS fixes through the phone's GPS receiver work fine in ex-Thalys trains, but it would still be nice to get this from the on-board WiFi.

**Additional context**

When connected to the onboard WiFi network called "EurostarTrainsWiFi", a request to `https://www.ombord.info/api/jsonp/position/` returns a response like the following:

```json
({
    "version":"1.11",
    "time":"1737986796",
    "age":"0",
    "latitude":"50.2925",
    "longitude":"2.881392",
    "altitude":"54.6",
    "speed":"83.741",
    "cmg":"27.43",
    "satellites":"22",
    "mode":"3"
});
```

(Note the surrounding `();` and that all values, including numerical ones, are encoded as strings.)

It's possible to specify a `callback=foo` URL parameter to get a response of the form `foo({ ... });`, but even without `callback=`, the `();` stays, so the response isn't "pure" JSON.

The speed is given in meters/sec; altitude is in meters. It seems like `time` + `age` = current unix time. I don't know what `mode` is; I've only ever seen it be 3.

The `cmg` key stands for "course made good", i.e. the train's bearing in degrees, where 0 is north, 90 is east, etc. At some points in my journey the value got stuck and stayed constant for a few minutes before catching itself.

Tested on an ex-Thalys train (now Eurostar) from Paris Nord to Bruxelles-Midi.

TimoWilken:
https://github.com/microg/GmsCore/pull/2821 has been merged; closing.

PR #2821, opened 2025-03-16
Location: Support moving WiFi location for Eurostar trains

TimoWilken:
This is an untested implementation of moving WiFi locations for Eurostar (see https://github.com/microg/GmsCore/issues/2769).

Feedback and/or testers would be welcome.

### DIFF
--- a/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
+++ b/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
@@ -411,7 +411,6 @@ class MovingWifiHelper(private val context: Context) {
                 // TODO: what happens in the Channel Tunnel? Does "satellites" go to zero? Does "mode" change?
                 if (json.has("satellites") && json.getInt("satellites") < 1) throw RuntimeException("Ombord has no GPS fix")
                 location.accuracy = 100f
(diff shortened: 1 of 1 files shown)


ale5000-git:
@TimoWilken
There is an issue: https://github.com/microg/GmsCore/pull/2821/files#annotation_32883288952

TimoWilken:
Thank you @ale5000-git, I've removed the `.toFloat()` call that was causing the error.

play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/wifi/MovingWifiHelper.kt
```diff
@@ -403,6 +404,34 @@ class MovingWifiHelper(private val context: Context) {
             }
         }
 
+        private val SOURCE_EUROSTAR = object : MovingWifiLocationSource("https://www.ombord.info/api/jsonp/position/") {
```
mar-v-in, line 407: Please name it `SOURCE_OMBOARD`. `omboard.info` is owned by Icomera, they likely have this system deployed on other trains as well, so it's not specific to Eurostar.

TimoWilken: Done in https://github.com/microg/GmsCore/pull/2821/commits/0670dc72d386332e4f91d1eea8ddae28805f6faa - thanks!

mar-v-in:
Thanks!

Issue #2745, opened 2025-01-15
google account Invalid Credential from microG 0.3.6


grenadin:
**Describe the bug**
Some Huawei phone found out invalid credentials when install microG 0.3.6 

**To Reproduce**
1. Go to Setting 
2. Click on Users & Accounts
3. Scroll down and select Google
4. Select any option 
    - Sign-in @ Security
    - Personal info & privacy
    - Account preferences
5. Error page show invalid credentials

**Expected behavior**
Should go through google account page

**Screenshots**

**System**
Android Version: 11-12
EMUI 12 - 15

**Additional context**
Error occurred in some Huawei phone that install new from 0.3.6 but if install over from lower it like from 0.3.5 and over it to 0.3.6, this problem not occur.

PR #2836, opened 2025-03-24
Clear cookies before opening web views

DaVinci9196:
Fix this issue.
This issue will occur when a new user logs in to the App through Google and the authorization page pops up.
For example: a new user logs in to dropbox for the first time
https://github.com/microg/GmsCore/issues/2745

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/auth/consent/ConsentSignInActivity.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/auth/consent/ConsentSignInActivity.kt
@@ -8,7 +8,7 @@ package org.microg.gms.auth.consent
 import android.annotation.SuppressLint
 import android.app.Activity
 import android.graphics.Bitmap
-import android.os.Build
+import android.os.Build.VERSION.SDK_INT
 import android.os.Bundle
 import android.os.Message
 import android.os.Messenger
@@ -92,7 +92,7 @@ class ConsentSignInActivity : Activity() {
 
     private fun initCookieManager() {
         val cookieManager = CookieManager.getInstance()
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if (SDK_INT >= 21) {
             cookieManager.removeAllCookies { _ ->
                 setCookiesAndLoadUrl(consentUrl!!, cookieManager)
             }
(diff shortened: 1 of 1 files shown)


play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/WebViewHelper.kt
```diff
@@ -109,7 +111,13 @@ class WebViewHelper(private val activity: AppCompatActivity, private val webView
             }
             Log.d(TAG, "Opening $authUrl")
             webView.post {
-                loadWebViewUrl(authUrl)
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                    CookieManager.getInstance().removeAllCookies {
+                        loadWebViewUrl(authUrl)
+                    }
+                } else {
+                    loadWebViewUrl(authUrl)
```
ale5000-git, line 119: Before Lollipop there is https://developer.android.com/reference/android/webkit/CookieManager#removeAllCookie()
(without the final s).
It seems similar but NOY asynchronous.

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/WebViewHelper.kt
```diff
@@ -109,7 +111,14 @@ class WebViewHelper(private val activity: AppCompatActivity, private val webView
             }
             Log.d(TAG, "Opening $authUrl")
             webView.post {
-                loadWebViewUrl(authUrl)
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
mar-v-in, line 114: Please directly refer to the respective constant value of the API level - in this case `21` instead `Build.VERSION_CODES.LOLLIPOP`.

mar-v-in (changes requested):

play-services-core/src/main/kotlin/org/microg/gms/auth/consent/ConsentSignInActivity.kt
```diff
@@ -163,4 +163,13 @@ class ConsentSignInActivity : Activity() {
             Log.d(TAG, "consent showView: ")
         }
     }
+
+    override fun onStop() {
+        super.onStop()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
mar-v-in, line 169: Same here.

mar-v-in (changes requested):

Issue #2832, opened 2025-03-21
[nl.rabobank.ida] Unable to call Play Integrity API's DecodeIntegrityToken method


mid-kid:
**Affected app**
Name: Datakeeper
Package id: nl.rabobank.ida

**Describe the bug**
This app doesn't manage to pass the Play Integrity checks, due to what I think is a microg bug. With the same phone setup, but with play services instead of microg, the checks pass.

**To Reproduce**
1. Open the app, configure a password
2. Click on 'Data' on the bottom bar, and click on the '+' to add data.
3. Select "Email address" from the list, and enter any valid email address.
4. Observe error 1004, and the logcat showing the play integrity results.

The logcat logs the following relevant message, coming from the Firebase App Check library that it's using:
```
E/APP CHECK PLUGIN(7171): Get token operation failed with message: Error returned from API. code: 400 body: Unable to call Play Integrity API's DecodeIntegrityToken method; ensure that the Play Integrity API is enabled and that the Cloud project you registered in the Play Integrity console is the same as the Firebase project containing this app.
```

**Expected behavior**
The app reaches the next screen, asking for an email confirmation code.

**Screenshots**
App blocks screenshots.

**System**
Android Version: 15
Custom ROM: crDroid 11.2
Phone: beryllium (Pocophone F1)

**microG**
microG Core version: 0.3.6.244735
microG Self-Check results: All ticked

ale5000-git:
Could you please re-try with the version 0.3.9 of microG to see if something is improved?

mid-kid:
@ale5000-git sorry for the late reply. Right now I get:
```
08-18 00:48:48.761  8604  8604 E APP CHECK PLUGIN: Get token operation failed with message: Error returned from API. code: 403 body: App attestation failed.
```
Unfortunately, since I last tested this, I can't get my device to pass DEVICE_INTEGRITY (only BASIC_INTEGRITY)... So I have no idea if the above bug is fixed now, nor how to get my device to pass so I can test it again.

mid-kid:
I don't think it'll ever be possible to make my device pass integrity again, as it's not well supported anymore.

PR #2860, opened 2025-04-16
Integrity: Add cloudProjectNumber to IntegrityParams

bluemods:
Field 6 in IntegrityParams is the cloud project number. 

Adding this to the request may fix issues like https://github.com/microg/GmsCore/issues/2832 where FirebaseAppCheck is used, which in turn passes a cloud project number to the request.


mar-v-in:
I don't think this generally solves issues with Firebase AppCheck, but it also shouldn't hurt to send the cloud project number, as this is what Google's implementation also does.

PR #2863, opened 2025-04-18
Password:Open with Chrome browser first

DaVinci9196:
Streamline the process.


mar-v-in:
Why is it important to start this in Chrome? The user should always be free to choose any browser they want to use, even if Chrome is installed.

Issue #2731, opened 2025-01-05
microG does not behave gracefully with invalid sessions of Google accounts


lucasmz-dev:
**Describe the bug**
When an user has a Google account logged in, if the session turns invalid (gets logged out remotely for example), many things break in microG and the user is not notified, only leaving the status of certain apps breaking as somewhat of a clue.

**To Reproduce**
1. Log in using a Google account
2. e.g. Google Maps works fine
3. Log out the device out of the Google account remotely via the web online account page (account.google.com)
4. Google Maps does not work properly anymore
5. No notification or other indication that the logged in account isn't a valid session anymore

**Expected behavior**
microG should automatically detect when a Google account gets logged out and remove the account and potentially warn the user.

**Screenshots**

**System**
Android Version: 14
Custom ROM: CalyxOS 5.14.2
microG version: 0.3.6

**Additional context**
I notice this as a issue every now and then in the Calyx channels.

fynngodau:
GMS' behavior is to show a notification, re-prompting the user to enter their password. I intended for this functionality to be added when I wrote this code:

https://github.com/microg/GmsCore/blob/27bacdde26e91015c7e1fad8e214826a3f8df10b/play-services-core/src/main/kotlin/org/microg/gms/accountaction/ErrorResolver.kt#L132-L133

…but then I didn't get around to implementing it. What is needed in that place is some kind of UI flow that first tells users to reauthenticate using a notification (in the background case) or directly open that page (in the foreground case). Also the background / foreground decision might still be optimized, as so far I only was able to test it for the enterprise account cryptauth flow, which won't occur for all different types of call.

noaho:
I just had this with Maps, was very confusing because I thought it was due to the latest signature checking changes, but it wasn't (my phone is Jan 2025 L4M build), but actually it was just because I updated my Google password. Maybe this could be a pinned topic?

dimaryaz:
I think I was seeing a similar issue - though in my case, it wasn't just Google Maps, but the whole OS somehow became unstable: #2405

ale5000-git:
Now that this is fixed and a new version of microG is out I'm removing this from pinned issues.

PR #2865, opened 2025-04-22
Auth: Handle invalid account

DaVinci9196:
https://github.com/microg/GmsCore/issues/2731#issuecomment-2571748442
Based on @fynngodau’s solution, use notifications to inform users whether their account has expired. Users can choose to log in again or log out.
Choosing to re-login will be handled based on the newly added reauth interface.

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -601,7 +601,8 @@
         <activity
             android:name="org.microg.gms.accountaction.AccountActionActivity"
             android:exported="true"
-            android:theme="@style/Theme.LoginBlue"/>
+            android:excludeFromRecents="true"
+            android:theme="@style/Theme.App.Translucent"/>
 
         <!-- Games -->
 
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -601,8 +601,7 @@
         <activity
             android:name="org.microg.gms.accountaction.AccountActionActivity"
             android:exported="true"
-            android:excludeFromRecents="true"
-            android:theme="@style/Theme.App.Translucent"/>
+            android:theme="@style/Theme.LoginBlue"/>
 
         <!-- Games -->
 
--- a/play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
@@ -1,9 +1,11 @@
 package org.microg.gms.accountaction
 
 import android.accounts.Account
+import android.accounts.AccountManager
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
+import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.annotation.RequiresApi
@@ -27,6 +29,8 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.google.android.gms.R
+import org.microg.gms.auth.AuthConstants
+import org.microg.gms.auth.login.LoginActivity
 
 internal const val INTENT_KEY_USER_ACTION = "userAction"
 internal const val INTENT_KEY_ACCOUNT_NAME = "accountName"
@@ -39,19 +43,42 @@ class AccountActionActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        Log.d(TAG, "onCreate: ")
 
-        if (savedInstanceState == null) {
-            val requirements = intent.getSerializableExtra(INTENT_KEY_USER_ACTION) as Array<Requirement>
-            taskMap.addAll(requirements.map { it to false })
-        }
+        val actionExtra = intent.getSerializableExtra(INTENT_KEY_USER_ACTION)
+        val accountName = intent.getStringExtra(INTENT_KEY_ACCOUNT_NAME)
 
-        val accountName = intent.getStringExtra(INTENT_KEY_ACCOUNT_NAME) ?: "<?>"
+        Log.d(TAG, "actionExtra: $actionExtra account: $accountName")
 
-        setContent {
-            Content(accountName, taskMap.toMap()) {
-                finish()
+        if (actionExtra is UserSatisfyRequirements) {
+            if (savedInstanceState == null) {
+                val requirements = actionExtra.actions.toTypedArray()
+                taskMap.addAll(requirements.map { it to false })
             }
+            setContent {
+                Content(accountName ?: "<?>", taskMap.toMap()) {
+                    finish()
+                }
+            }
+            return
         }
+
+        if (actionExtra is Reauthenticate) {
+            val account = AccountManager.get(this).accounts.find { it.name == accountName && it.type == AuthConstants.DEFAULT_ACCOUNT_TYPE } ?: return finish()
+            val reAuth = actionExtra.reAuth
+            if (reAuth) {
+                val intent = Intent(this, LoginActivity::class.java).apply {
+                    putExtra(LoginActivity.EXTRA_RE_AUTH_ACCOUNT, account)
+                }
+                startActivity(intent)
+            } else {
+                AccountManager.get(this).removeAccount(account, null, null)
+            }
+            cancelAccountNotificationChannel(account)
+            finish()
+            return
+        }
+
     }
 
     override fun onResume() {
@@ -63,9 +90,9 @@ class AccountActionActivity : ComponentActivity() {
     }
 
     companion object {
-        fun createIntent(context: Context, account: Account, action: UserSatisfyRequirements) =
+        fun createIntent(context: Context, account: Account, action: Resolution) =
             Intent(context, AccountActionActivity::class.java).apply {
-                putExtra(INTENT_KEY_USER_ACTION, action.actions.toTypedArray())
+                putExtra(INTENT_KEY_USER_ACTION, action)
                 putExtra(INTENT_KEY_ACCOUNT_NAME, account.name)
             }
     }
--- a/play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
@@ -1,11 +1,9 @@
 package org.microg.gms.accountaction
 
 import android.accounts.Account
-import android.accounts.AccountManager
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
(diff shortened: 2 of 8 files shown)


play-services-core/src/main/kotlin/org/microg/gms/accountaction/ErrorResolver.kt
```diff
@@ -128,8 +128,11 @@ fun <T> Resolution.initiateFromBackgroundBlocking(context: Context, account: Acc
             }
             return null
         }
-        Reauthenticate -> {
+        is Reauthenticate -> {
             Log.w(TAG, "Your account credentials have expired! Please remove the account, then sign in again.")
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
ale5000-git, line 133: @DaVinci9196
@mar-v-in
[This](https://github.com/microg/GmsCore/pull/2865/files#diff-b9b9347e3a0cd101b758bf9a6e34d55e490fde3821086bb48e2e29978d3a8408L5) is another remnant before the decision of:
- For fake data (org.microg.gms.profile.Build) import all
- For real data (android.os.Build)  import just the specific fields

lucasmz-dev:
Nice! Great to see this fixed 😭😭😭

play-services-core/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,11 @@ This can take a couple of minutes."</string>
     <string name="auth_action_notification_title">Account action required</string>
     <string name="auth_action_notification_content">Your Google account needs additional setup.</string>
 
+    <string name="auth_action_reauth_notification_title">Account Issue Detected</string>
```
lucasmz-dev, line 46:
```suggestion
    <string name="auth_action_reauth_notification_title">Account issue detected</string>
```

play-services-core/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,11 @@ This can take a couple of minutes."</string>
...
+    <string name="auth_action_reauth_notification_content">There is an issue with your account (%1$s). Please log in again.</string>
+    <string name="auth_action_reauth_notification_re_login">Log In Again</string>
```
lucasmz-dev, line 48:
```suggestion
    <string name="auth_action_reauth_notification_re_login">Log in again</string>
```

play-services-core/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,11 @@ This can take a couple of minutes."</string>
...
+    <string name="auth_action_reauth_notification_remove_account">Remove Account</string>
```
lucasmz-dev, line 49:
```suggestion
    <string name="auth_action_reauth_notification_remove_account">Remove account</string>
```

Basically try not to use title case, it can also end up being a bit annoying when translating because it's not always obvious what's actually meant to be like that and what isn't: https://m3.material.io/foundations/content-design/style-guide/ux-writing-best-practices (check "Use sentence case"

play-services-core/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,11 @@ This can take a couple of minutes."</string>
     <string name="auth_action_notification_title">Account action required</string>
     <string name="auth_action_notification_content">Your Google account needs additional setup.</string>
 
+    <string name="auth_action_reauth_notification_title">Account issue detected</string>
+    <string name="auth_action_reauth_notification_content">There is an issue with your account (%1$s). Please perform relevant actions.</string>
+    <string name="auth_action_reauth_notification_re_login">Log in again</string>
+    <string name="auth_action_reauth_notification_remove_account">Remove account</string>
```
fynngodau, line 49: From my personal perspective on UX, the notification content should be very short (on my screen the last few words are cropped off). To achieve this, I would not include the email address again as it is already shown in the notification subtext. Additionally, rather than "issue detected", I would use something more actionable (the other type of notification uses "Account action required", which could fit here as well). "Relevant actions" is also not very specific – are there cases where the flow will do anything other than ask for user's credentials?

If I remember correctly, the notification Google Play services uses talks about "entering your password again". I propose the following texts:

```suggestion
    <string name="auth_action_reauth_notification_title">Signed out of account</string>
    <string name="auth_action_reauth_notification_content">Please sign in again to keep using your account.</string>
    <string name="auth_action_reauth_notification_re_login">Sign in again</string>
    <string name="auth_action_reauth_notification_remove_account">Remove account</string>
```

play-services-core/src/main/kotlin/org/microg/gms/accountaction/Resolution.kt
```diff
@@ -30,7 +30,9 @@ enum class Requirement {
  * Represents a situation in which the user's authentication has become
  * invalid, and they need to enter their credentials again.
  */
-data object Reauthenticate : Resolution()
+data class Reauthenticate(var reAuth: Boolean = true) : Resolution(), Serializable {
+    fun remove() = apply { reAuth = false }
+}
```
fynngodau, line 35: The `reAuth` boolean seems to have the purpose of telling the activity that handles the intent that is sent after the user clicked on one of the two notification options about which option they chose. Please revert the change here and make this a part of the intent proper.

play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
```diff
@@ -39,19 +43,42 @@ class AccountActionActivity : ComponentActivity() {
...
+        if (actionExtra is Reauthenticate) {
+            val account = AccountManager.get(this).accounts.find { it.name == accountName && it.type == AuthConstants.DEFAULT_ACCOUNT_TYPE } ?: return finish()
+            val reAuth = actionExtra.reAuth
+            if (reAuth) {
+                val intent = Intent(this, LoginActivity::class.java).apply {
+                    putExtra(LoginActivity.EXTRA_RE_AUTH_ACCOUNT, account)
+                }
+                startActivity(intent)
+            } else {
+                AccountManager.get(this).removeAccount(account, null, null)
```
fynngodau, line 75: This doesn't really have anything to do with the rest of the activity – please extract it elsewhere. Then you can also simply revert all of the changes to this activity to fix the regression with `UserSatisfyRequirements` (see first comment).

play-services-core/src/main/AndroidManifest.xml
```diff
@@ -601,7 +601,8 @@
         <activity
             android:name="org.microg.gms.accountaction.AccountActionActivity"
             android:exported="true"
-            android:theme="@style/Theme.LoginBlue"/>
+            android:excludeFromRecents="true"
+            android:theme="@style/Theme.App.Translucent"/>
```
fynngodau, line 605: Does this not cause issues with normal `UserSatisfyRequirements` usage of the activity?

fynngodau (changes requested):
Many thanks for implementing this feature. However, as a regression, the `UserSatisfyRequirements` resolution feature no longer works with these changes.

play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
```diff
@@ -63,9 +90,9 @@ class AccountActionActivity : ComponentActivity() {
     }
 
     companion object {
-        fun createIntent(context: Context, account: Account, action: UserSatisfyRequirements) =
+        fun createIntent(context: Context, account: Account, action: Resolution) =
             Intent(context, AccountActionActivity::class.java).apply {
-                putExtra(INTENT_KEY_USER_ACTION, action.actions.toTypedArray())
+                putExtra(INTENT_KEY_USER_ACTION, action)
                 putExtra(INTENT_KEY_ACCOUNT_NAME, account.name)
```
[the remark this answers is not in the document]
DaVinci9196: Thanks for the reminder, the changes to UserSatisfyRequirements have been removed

play-services-core/src/main/kotlin/org/microg/gms/accountaction/ErrorResolver.kt
```diff
@@ -150,15 +153,18 @@ fun <T> Resolution.initiateFromForegroundBlocking(context: Context, account: Acc
...
                 }
             }
             return null
         }
-        Reauthenticate -> {
+        is Reauthenticate -> {
             Log.w(TAG, "Your account credentials have expired! Please remove the account, then sign in again.")
+            if (SDK_INT >= 21) {
+                context.sendAccountReAuthNotification(account, this)
+            }
```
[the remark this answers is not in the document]
DaVinci9196: It has been modified. However, when verifying an invalid account, for example, when jumping to the web page for managing accounts, the error that occurs is not BAD_AUTHENTICATION but server exception <500>. I have added a SERVER_ERROR category. Do you think it will affect the UserSatisfyRequirements process?

play-services-core/src/main/kotlin/org/microg/gms/accountaction/Resolution.kt
DaVinci9196: Revoked

play-services-core/src/main/kotlin/org/microg/gms/accountaction/AccountActionActivity.kt
DaVinci9196: Revoked

play-services-core/src/main/AndroidManifest.xml
DaVinci9196: Revoked

play-services-core/src/main/res/values/strings.xml
DaVinci9196: It is necessary to remind users which account is invalid, as different apps will use different accounts. However, the reminder text has been simplified.

play-services-core/src/main/kotlin/org/microg/gms/accountaction/ErrorResolver.kt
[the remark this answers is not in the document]
fynngodau: Thanks for testing this again. Indeed this is an improvement that I don't think has any negative side-effects, I'm really happy with it :+1:

fynngodau:
One thing I noticed is that reauth is not possible if Android ID has changed (i.e. when microG data has been deleted, but accounts were kept). This should be relatively easy to fix, but I don't think it needs to be solved in this PR. :slightly_smiling_face:

fynngodau:
When testing this PR, I removed some specific devices from the Security tab in myaccount. A few days later, unrelated to this PR, I was not able to sign in (or reauth) from microG until I changed Android ID; perhaps the device was locked out in a way microG is unable to recover from? The server response was only a generic error.

fynngodau (approved):
Thanks for the improvements. Reauth and enterprise setup both work at the same time now. :+1:

I'm not entirely happy with the UI choices made for the notification, but that's for @mar-v-in to decide.

play-services-core/src/main/res/values/strings.xml
mar-v-in: I think we eventually may want to review the notification content here, but I think for now it is good to merge.

PR #2866, opened 2025-04-23
Auth: Support Google 2-step verification

DaVinci9196:
Regarding the implementation of Google's two-step verification function. 
Users can receive information by turning on <Allow to receive two-step verification information> under their Google account. 
It should be noted that this function relies on Google Cloud Message Push, so please make sure it is turned on and connected normally.


000exploit:
If I thinking correctly — is this a feature for getting a code when I'm signing in on another device?

Sense101:
This is great to see. Awesome work!

play-services-core/src/main/res/values/strings.xml
```diff
@@ -166,6 +166,8 @@ Please set up a password, PIN, or pattern lock screen."</string>
     <string name="pref_auth_include_android_id_summary">When disabled, authentication requests won\'t be linked to the device registration, which may allow unauthorized devices to sign in, but may have unforeseen consequences.</string>
     <string name="pref_auth_strip_device_name_title">Strip device name for authentication</string>
     <string name="pref_auth_strip_device_name_summary">When enabled, authentication requests won\'t include the device\'s name, which may allow unauthorized devices to sign in, but may have unforeseen consequences.</string>
+    <string name="pref_auth_two_step_verification_title">Receive Google Two-Step Verification</string>
```
lucasmz-dev, line 169:
```suggestion
    <string name="pref_auth_two_step_verification_title">Receive two-step verification prompts</string>

File: play-services-core/src/main/res/values/strings.xml [hunk 2]
```diff
@@ -166,6 +166,8 @@ Please set up a password, PIN, or pattern lock screen."</string>
...
+    <string name="pref_auth_two_step_verification_summary">When enabled, the device can receive Two-Step verification notifications from Google (Cloud Messaging must be enabled and keep connected).</string>
```
lucasmz-dev, line 170:
```suggestion
   <string name="pref_auth_two_step_verification_summary">When enabled, the device can receive two-step verification prompts from Google (Cloud Messaging is required).</string>
```

play-services-core/src/main/java/org/microg/gms/gcm/McsService.java
```diff
@@ -367,9 +369,7 @@ private void handleSendMessage(Intent intent) {
                 ttl = maxTtl;
             }
         } catch (NumberFormatException e) {
-            // TODO: error TtlUnsupported
-            Log.w(TAG, e);
```
ale5000-git, line 371: Why remove the warning?

DaVinci9196: The value is assigned after the exception. Do you need this exception log?

ale5000-git: I think it may be useful to keep the warning even if a new value is assigned later.
Because if the "NumberFormatException" happens rarely it make sense to just workaround it by using max TTL but it if happens often it may make more sense to debug it and support these "strange" TTL values.

DaVinci9196: Got it, I'll join this

mar-v-in (approved):

Idcrafter:
Hey i noticed a small issue with this, microG never asks for Cloud messaging registration if you have turned on that you can decide to allow new apps to register. for now can you just shortly turn it off so that microG can register for cloud messaging and then turn it back on again.
would it be possible to allow microG to register while ignoring the setting or make it ask for permission?

lucasmz-dev:
@Idcrafter why not open an issue

PR #2879, opened 2025-04-28
OcWeb: Improve OC corresponding JS method

DaVinci9196:
Supplement the JS method corresponding to OcIdWebview.
Also modify the following:
1. Fixed the problem that the link corresponding to <ID:400> request reports 400.
2. Fixed the problem that the link corresponding to <ID:10052> does not match the actual display.
3. MainActivity adds a toolbar, the title changes with the js method, and a button to close the page.
4. Modify user avatar to support uploading from local gallery.
5. Modify user avatar to support taking photos and uploading.


play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
```diff
@@ -94,7 +113,7 @@ private val SCREEN_ID_TO_URL = hashMapOf(
     10007 to "https://myaccount.google.com/payments-and-subscriptions",
     10015 to "https://support.google.com/accounts",
     10050 to "https://myaccount.google.com/profile",
-    10052 to "https://myaccount.google.com/embedded/family/create",
+    10052 to "https://myaccount.google.com/u/1/family/details",
```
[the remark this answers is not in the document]
DaVinci9196: Modified, thanks

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcAndroidIdBridge.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.microg.gms.checkin.LastCheckinInfo
+
+class OcAndroidIdBridge(val context: Context) {
+
+    companion object {
+        const val NAME = "ocAndroidId"
+        private const val TAG = "JS_$NAME"
+    }
+
+    @OptIn(ExperimentalStdlibApi::class)
```
mar-v-in, line 20: Please don't use experimental API.

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcAndroidIdBridge.kt
```diff
@@ -0,0 +1,28 @@
...
+    @JavascriptInterface
+    fun getAndroidId(): String? {
+        Log.d(TAG, "getAndroidId: ")
+        val androidId = LastCheckinInfo.read(context).androidId
+        return if (androidId != 0L) androidId.toHexString() else null
```
mar-v-in, line 25:
```suggestion
        return if (androidId != 0L) androidId.toString(16) else null
```

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcAppBarBridge.kt
```diff
@@ -0,0 +1,142 @@
...
+
+class OcAppBarBridge(val toolBar: Toolbar, val webView: WebView) {
+
+    companion object {
+        const val NAME = "ocAppBar"
+        private const val TAG = "JS_$NAME"
+    }
+
+    @JavascriptInterface
+    @Synchronized
```
mar-v-in, line 26: Why all the `@Synchronized`? JavaScript is already single-thread anyway.

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcClientInfoBridge.kt
```diff
@@ -0,0 +1,51 @@
...
+    @JavascriptInterface
+    fun getGmsCoreVersion(): Int {
+        Log.d(TAG, "getGmsCoreVersion: ")
+        return Constants.GMS_VERSION_CODE
+    }
+
+    @JavascriptInterface
+    fun getOsVersion(): String? {
+        Log.d(TAG, "getOsVersion: ")
+        return RELEASE
```
mar-v-in, line 42: I think returning from the `org.microg.gms.profile.Build` would be more appropriate here.

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcFilePickerBridge.kt
```diff
@@ -0,0 +1,236 @@
...
+    private fun launchChooser(mimeType: String) {
+        if (mimeType.startsWith("image/") || mimeType == "image/*" || mimeType == "*/*") {
+            when {
+                ContextCompat.checkSelfPermission(activity, Manifest.permission.CAMERA) == PackageManager.PERMISSION_GRANTED -> {
+                    launchChooserInternal(mimeType)
+                }
+
+                else -> {
+                    pendingMimeType = mimeType
+                    cameraPermissionLauncher.launch(Manifest.permission.CAMERA)
```
mar-v-in, line 131: Don't request the CAMERA permission before the user selected to upload a photo from camera, there is no reason to request and grant the permission if the user only wants to upload a file from storage.

mar-v-in (changes requested):

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcAndroidIdBridge.kt
DaVinci9196: Changed

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcFilePickerBridge.kt
DaVinci9196: The idea here is that after the user chooses to take a photo, if there is no camera permission, it will actively apply for permission to allow the user to use the photo function.

mar-v-in: It would be fine to ask for camera permission *after* the user selected to use the camera. So user flow would be:
1. User selects to set new avater
2. User is presented options "Take picture" and "Choose from Gallery"
3. User selects "Take picture"
4. User is asked for Camera permission
5. Camera is opened

DaVinci9196: Understand. Of course, it has been modified now to only let users choose to take pictures or albums when they have camera permissions. If they do not have permissions, they will only jump to the album. This can avoid users from participating in too many permission granting processes.

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcClientInfoBridge.kt
```diff
@@ -0,0 +1,50 @@
...
+    @JavascriptInterface
+    fun getOsVersion(): String? {
+        Log.d(TAG, "getOsVersion: ")
+        return org.microg.gms.profile.Build.VERSION.RELEASE
+    }
+
+    @JavascriptInterface
+    fun getSdkVersion(): Int {
+        Log.d(TAG, "getSdkVersion: ")
+        return SDK_INT
```
mar-v-in: This should better use `org.microg.gms.profile.Build.SDK_INT`. Best is to `import org.microg.gms.profile.Build` and then use `Build.SDK_INT` here and `Build.VERSION.RELEASE` above.

mar-v-in (changes requested):

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/bridge/OcClientInfoBridge.kt
DaVinci9196: changed

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
```diff
@@ -62,7 +82,7 @@ private val SCREEN_ID_TO_URL = hashMapOf(
     310 to "https://myaccount.google.com/reservations",
     312 to "https://myaccount.google.com/accessibility",
     313 to "https://myaccount.google.com/inputtools",
-    400 to "https://myaccount.google.com/security-checkup/",
+    400 to "https://myaccount.google.com/security-checkup/6",
```
ale5000-git: Instead of hardcoding 6 isn't better to use this? `https://myaccount.google.com/smartlink/security-checkup`
Or also `https://myaccount.google.com/security-checkup` without final slash.

play-services-core/src/main/java/org/microg/gms/settings/GmsFileProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+import android.os.ParcelFileDescriptor
+import android.util.Log
+import androidx.core.content.FileProvider
+import java.io.FileNotFoundException
+
+private const val TAG = "GmsFileProvider"
+
+class GmsFileProvider : FileProvider() {
+    private val emptyProjection = arrayOfNulls<String>(0)
+    private var isInitializationFailed = false
```
ale5000-git: Isn't better `hasInitializationFailed` or just `initializationFailed`?

mar-v-in:
I tried uploading a user avatar with this merged and it did not work for me. The button is not responsive and there is no log message.

DaVinci9196:
> I tried uploading a user avatar with this merged and it did not work for me. The button is not responsive and there is no log message.

I suspect this is a problem with the webview. I tried to implement a method to get local images from the loaded js. I couldn't select the file through the device webview, but it was normal through chrome or computer chrome.

document.getElementById("uploadBtn").addEventListener("click", () => {
      const input = document.createElement("input");
      input.type = "file";
      input.accept = "image/*";  
      input.multiple = false;
      input.style.display = "none";

      input.addEventListener("change", (event) => {
        const file = event.target.files[0];
        if (file) {
          previewImage(file);
          uploadFile(file);
        }
      });

      document.body.appendChild(input);
      input.click();
      document.body.removeChild(input);
});

mar-v-in:
> I suspect this is a problem with the webview. I tried to implement a method to get local images from the loaded js. I couldn't select the file through the device webview, but it was normal through chrome or computer chrome.

I'm a little surprised because there is specific code for user avatar uploading in this PR, but I can't find how to trigger it.

The only bridge functions that I see being called are:
- `ocConsistency.verifyActualAccountId`
- `ocUi.isOpenScreenEnabled`
- `ucUi.isOpenHelpEnabled`
- `ocAppBar.hide`
- `ocPlayProtect.isPlayProtectEnabled` (when opening security checkup)
- `ocUdc.getSupportedDeviceSettings`
- `ocAppBar.*`

Notably, there is no call to any `ocFilePicker` functions. Can you explain how to reach and test them?

DaVinci9196:
@mar-v-in  There are two places to change your avatar.
1. image
2. image
The first replacement can execute ocFilePicker normally, but the second one cannot execute normally.

play-services-core/src/main/java/org/microg/gms/settings/GmsFileProvider.kt
DaVinci9196: changed

play-services-core/src/main/kotlin/org/microg/gms/accountsettings/ui/MainActivity.kt
DaVinci9196: thanks, changed

Issue #2868, opened 2025-04-25
I can't give the new permission to search device accounts


girshcatt:
**Describe the bug**
I cant activate the search device accounts permission checkbox

**To Reproduce**
1. Go to 'Permissions section in microg'
2. Click on 'Search device accounts checkbox'
4. See nothing changed

**Expected behavior**
Its expected for microg to redirect me on giving permission window or something

**Screenshots**

![Image](https://github.com/user-attachments/assets/9f7033d9-20d4-4dc3-ae58-80fd74c567a1)

**System**
Android Version: 14
Custom ROM: CrDroid 10.9

dimka40185:
I have the same problem on Huawei Pura 70

![Image](https://github.com/user-attachments/assets/cc76e173-b50c-4e57-8045-7c44064dcd71)

ale5000-git:
Hi,
it is opportune that you provide an english translation of the thing not ticked in the image.
It isn't like everything know your language and an image can't be translated.

ale5000-git:
Also about the new permissions, most are privileged permission so they can't be granted unless you place the updated files in the system priv-app folder (with the corresponding new xml).

This apply only to normal phones, for Huawei there is no solution other then wait a ROM update (hopefully Huawei will fix it).

dimka40185:
> Hi, it is opportune that you provide an english translation of the thing not ticked in the image. It isn't like everything know your language and an image can't be translated.

![Image](https://github.com/user-attachments/assets/1a209e64-cec1-4f34-aaaa-72a50c2ed082)

ale5000-git:
For the GSF issue you just need to install Services Framework Proxy from here: https://microg.org/download.html

For the issue with permission to "find accounts on the device" try to grant manually the contacts permission to microG Services in the Android settings.

girshcatt:
also i noticed that i can give location & notification permissions to companion, how are these permissions used in it?

ale5000-git:
> Btw do I need only to have MicroG 0.3.7 as system app, or now i need also to put Companion there?

If possible also Companion should be a privileged system app (with the corresponding xml file) to have everything working.
Note that all microG works also as user app, but there may be some limitations.

> also i noticed that i can give location & notification permissions to companion, how are these permissions used in it?

There have been quite a few changes in the new version and I don't remember everything in detail but others may know it better.

girshcatt:
@HanabishiRecca how can i fix this?

HanabishiRecca:
[...]

PR #2882, opened 2025-04-29
play-services-auth-workaccount: Remove maxSdkVersion from perms

chirayudesai:
* This is causing them not to be granted at all on newer devices, _potentially_ leading to issues.
* They're already declared in play-services-core as well.

Bug: #2868

### DIFF
--- a/play-services-auth-workaccount/core/src/main/AndroidManifest.xml
+++ b/play-services-auth-workaccount/core/src/main/AndroidManifest.xml
@@ -6,15 +6,9 @@
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <uses-permission
-        android:name="android.permission.AUTHENTICATE_ACCOUNTS"
-        android:maxSdkVersion="22" />
-    <uses-permission
-        android:name="android.permission.GET_ACCOUNTS"
-        android:maxSdkVersion="22" />
-    <uses-permission
-        android:name="android.permission.MANAGE_ACCOUNTS"
-        android:maxSdkVersion="22" />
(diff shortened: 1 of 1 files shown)


chirayudesai:
See https://github.com/microg/GmsCore/issues/2868#issuecomment-2839030920 for more.

Could also remove them from `-auth-workaccount`, if that's preferred.

We've got at least two reports so far of people having issues with Gmail and Calendar sync after 0.3.7 - and this felt related. Need to investigate further, but in the interim, this was an easy thing to do.

ale5000-git:
@chirayudesai
We need to keep `android.permission.GET_ACCOUNTS` on all Android versions to be sure to have everything working but not `android.permission.AUTHENTICATE_ACCOUNTS` and `android.permission.MANAGE_ACCOUNTS` that are deprecated in new Android versions (they no longer do anything).

See:
https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-6.0.1_r81/core/res/AndroidManifest.xml#804
https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-6.0.1_r81/core/res/AndroidManifest.xml#809

So I suggest to remove `maxSdkVersion` only from `android.permission.GET_ACCOUNTS` and update self-check accordingly.

ale5000-git:
@mar-v-in
What do you think of this?
https://github.com/micro5k/GmsCoreMod/commit/d44a9d85c5a0a1f423cea62b30c265f28d5336a5

mar-v-in:
@ale5000-git I'm going with this PR for now, which should be fine for now. We can revisit correct permissions on various sdk versions at a later point.

PR #2885, opened 2025-05-02
Fido: fix support for discoverable keys

p1gp1g:
When a FIDO request for resident key is sent, allowList is null or empty. This commit restore resident keys


ale5000-git:
Isn't this already being done as part of #2463?

p1gp1g:
Indeed, #2463 contains the same fix. I didn't check that. If you plan to merge 2463 for the next release, then this PR can be closed

mar-v-in:
Merged via #3031

PR #2799, opened 2025-02-28
split install service

DaVinci9196:
Based on this commit https://github.com/microg/GmsCore/pull/2553, Split install Service is extracted and some issues in splite Install are fixed

PR #2898, opened 2025-05-08
Fix split install issue

DaVinci9196:
Extract the optimization and resubmit it to https://github.com/microg/GmsCore/pull/2799
- Download notifications for split packages should be dismissible.
- After a split package is downloaded, clicking the notification should trigger installation (to handle cases where the install prompt can't appear due to a locked screen).
- Prevent multiple notifications for the same split package download.
- Remove notifications when the network is disconnected or the download fails.
- Implement resumable downloads to save data usage.
- Ensure device synchronization to avoid missing split package information.
- Support account elevation to prevent failure in accessing split package data.

### DIFF
--- a/vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
+++ b/vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
@@ -5,10 +5,20 @@
 
 package org.microg.vending.delivery
 
+import android.accounts.AccountManager
+import android.content.Context
 import android.util.Log
 import com.android.vending.buildRequestHeaders
+import com.google.android.finsky.BulkGrant
+import com.google.android.finsky.BulkRequest
+import com.google.android.finsky.BulkRequestWrapper
+import com.google.android.finsky.BulkResponseWrapper
+import com.google.android.finsky.DeviceSyncInfo
+import com.google.android.finsky.SyncResponse
 import com.google.android.finsky.splitinstallservice.PackageComponent
+import com.google.android.finsky.syncDeviceInfo
 import org.microg.vending.billing.core.AuthData
+import org.microg.vending.billing.core.GooglePlayApi
 import org.microg.vending.billing.core.GooglePlayApi.Companion.URL_DELIVERY
 import org.microg.vending.billing.core.HttpClient
 import org.microg.vending.billing.proto.GoogleApiResponse
@@ -22,11 +32,12 @@ private const val TAG = "GmsVendingDelivery"
  * only those will be contained in the result.
  */
 suspend fun HttpClient.requestDownloadUrls(
-    packageName: String,
-    versionCode: Long,
-    auth: AuthData,
-    requestSplitPackages: List<String>? = null,
-    deliveryToken: String? = null,
+        context: Context,
+        packageName: String,
+        versionCode: Long,
+        auth: AuthData,
+        requestSplitPackages: List<String>? = null,
+        deliveryToken: String? = null,
 ): List<PackageComponent> {
 
     val requestUrl = StringBuilder("$URL_DELIVERY?doc=$packageName&ot=1&vc=$versionCode")
@@ -48,15 +59,34 @@ suspend fun HttpClient.requestDownloadUrls(
     }
     Log.d(TAG, "requestDownloadUrls languages: $languages")
 
+    val androidId = auth.gsfId.toLong(16)
     val headers = buildRequestHeaders(
         auth = auth.authToken,
-        // TODO: understand behavior. Using proper Android ID doesn't work when downloading split APKs
-        androidId = if (requestSplitPackages != null) 1 else auth.gsfId.toLong(16),
+        androidId = androidId,
         languages
     ).minus(
         // TODO: understand behavior. According to tests, these headers break split install queries but may be needed for normal ones
         (if (requestSplitPackages != null) listOf("X-DFE-Encoded-Targets", "X-DFE-Phenotype", "X-DFE-Device-Id", "X-DFE-Client-Id") else emptyList()).toSet()
     )
+
+    kotlin.runCatching {
+        //Synchronize account device information to prevent failure to obtain sub-package download information
+        syncDeviceInfo(context, AccountManager.get(context).accounts.firstOrNull { it.name == auth.email }!!, auth.authToken, androidId)
+    }
+    kotlin.runCatching {
+        //Authorize the account to prevent the inability to obtain split information
+        post(
+                url = GooglePlayApi.URL_BULK,
+                headers = headers,
+                payload = BulkRequestWrapper.build {
+                    request(BulkRequest.build {
+                        packageName(packageName)
+                        grant(BulkGrant.build { grantLevel = 1 })
+                    })
+                },
+                adapter = BulkResponseWrapper.ADAPTER
+        )
+    }
 
     val response = get(
         url = requestUrl.toString(),
@@ -76,9 +106,9 @@ suspend fun HttpClient.requestDownloadUrls(
         if (requestSplitPackages != null) {
             // Only download requested, if specific components were requested
             requestSplitPackages.firstOrNull { requestComponent ->
-                requestComponent.contains(it.splitPackageName!!)
+                (it.splitPackageName?.contains(requestComponent) ?: false || requestComponent.contains(it.splitPackageName!!))
             }?.let { requestComponent ->
-                PackageComponent(packageName, requestComponent, it.downloadUrl!!, it.size!!.toLong())
+                PackageComponent(packageName, it.splitPackageName!!, it.downloadUrl!!, it.size!!.toLong())
             }
         } else {
             // Download all offered components (server chooses)
--- a/vending-app/src/main/java/org/microg/vending/ui/InstallProgressNotification.kt
+++ b/vending-app/src/main/java/org/microg/vending/ui/InstallProgressNotification.kt
@@ -41,15 +41,25 @@ internal fun Context.notifySplitInstallProgress(packageName: String, sessionId: 
     when (progress) {
         is Downloading -> getDownloadNotificationBuilder().apply {
             setContentTitle(getString(R.string.installer_notification_progress_splitinstall_downloading, label))
-            setProgress(progress.bytesDownloaded.toInt(), progress.bytesTotal.toInt(), false)
+            setProgress(100, ((progress.bytesDownloaded.toFloat() / progress.bytesTotal) * 100).toInt().coerceIn(0, 100), false)
         }
-        CommitingSession -> getDownloadNotificationBuilder().apply {
-            setContentTitle(getString(R.string.installer_notification_progress_splitinstall_commiting, label))
+        is CommitingSession -> getDownloadNotificationBuilder().apply {
+            if (progress.deleteIntent != null) {
+                setDeleteIntent(progress.deleteIntent)
+            }
+            if (progress.installIntent != null) {
+                setContentTitle(getString(R.string.installer_notification_progress_splitinstall_click_install, label))
+                addAction(R.drawable.ic_download, getString(R.string.vending_overview_row_action_install), progress.installIntent)
+                setContentIntent(progress.installIntent)
+                setAutoCancel(true)
+            } else {
+                setContentTitle(getString(R.string.installer_notification_progress_splitinstall_commiting, label))
+            }
             setProgress(0, 1, true)
         }
         else -> null.also { notificationManager.cancel(sessionId) }
     }?.apply {
-        setOngoing(true)
+        setOngoing(false)
 
         notificationManager.notify(sessionId, this.build())
     }
@@ -84,7 +94,7 @@ internal fun Context.notifyInstallProgress(
                 setOngoing(true)
                 return this.build().also { notificationManager.notify(sessionId, it) }
             }
-            CommitingSession -> {
+            is CommitingSession -> {
                 setContentTitle(
                     getString(
                         if (isDependency) R.string.installer_notification_progress_splitinstall_commiting
--- a/vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
+++ b/vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
@@ -1,11 +1,7 @@
-/*
- * SPDX-FileCopyrightText: 2025 e foundation
- * SPDX-License-Identifier: Apache-2.0
- */
-
 package com.android.vending.installer
 
 import android.annotation.SuppressLint
+import android.app.KeyguardManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
@@ -13,9 +9,12 @@ import android.content.pm.PackageInfo
 import android.content.pm.PackageInstaller
 import android.content.pm.PackageInstaller.SessionParams
 import android.os.Build
+import android.os.PowerManager
 import android.util.Log
 import androidx.annotation.RequiresApi
+import androidx.core.content.pm.PackageInfoCompat
 import com.google.android.finsky.splitinstallservice.PackageComponent
+import com.google.android.finsky.splitinstallservice.SplitInstallService
 import kotlinx.coroutines.CompletableDeferred
 import org.microg.vending.billing.core.HttpClient
 import org.microg.vending.enterprise.CommitingSession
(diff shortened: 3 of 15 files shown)


vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -1,21 +1,20 @@
-/*
- * SPDX-FileCopyrightText: 2025 e foundation
- * SPDX-License-Identifier: Apache-2.0
- */
-
```
fynngodau: Please don't remove the license information header.

vending-app/src/main/java/org/microg/vending/ui/InstallProgressNotification.kt
```diff
@@ -41,15 +41,25 @@ internal fun Context.notifySplitInstallProgress(packageName: String, sessionId:
...
+                setDeleteIntent(progress.deleteIntent)
+            }
+            if (progress.installIntent != null) {
+                setContentTitle(getString(R.string.installer_notification_progress_splitinstall_click_install, label))
+                addAction(R.drawable.ic_download, getString(R.string.vending_overview_row_action_install), progress.installIntent)
+                setContentIntent(progress.installIntent)
+                setAutoCancel(true)
+            } else {
+                setContentTitle(getString(R.string.installer_notification_progress_splitinstall_commiting, label))
+            }
```
fynngodau: While the behavior here implemented is generally okay (for situations where microG is not privileged to install apps), this also changes the notification when the package is already installed. So this new type of notification should:

* be displayed if and only if user interaction is necessary
* not show a progress indicator

For this it needs to be a separate progress state, I think.

vending-app/src/main/java/org/microg/vending/ui/InstallProgressNotification.kt
```diff
@@ -41,15 +41,25 @@ internal fun Context.notifySplitInstallProgress(packageName: String, sessionId:
...
             setProgress(0, 1, true)
         }
         else -> null.also { notificationManager.cancel(sessionId) }
     }?.apply {
-        setOngoing(true)
+        setOngoing(false)
```
fynngodau: Generally this seems like an improper solution, as we don't want notifications of ongoing downloads to be dismissable. What is the problem you wanted to resolve here? Why is it not solved by explicitly removing the download notification when a download fails? (Or is there perhaps an issue with that code that should be fixed instead?)

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
```diff
@@ -76,9 +106,9 @@ suspend fun HttpClient.requestDownloadUrls(
         if (requestSplitPackages != null) {
             // Only download requested, if specific components were requested
             requestSplitPackages.firstOrNull { requestComponent ->
-                requestComponent.contains(it.splitPackageName!!)
+                (it.splitPackageName?.contains(requestComponent) ?: false || requestComponent.contains(it.splitPackageName!!))
```
fynngodau: Can add a comment why this is a good behavior? It seems like a rather lax matching – can we be sure that we download the correct package?

(Also, I think the following is more idiomatic and easier to understand:)

```suggestion
                (it.splitPackageName?.contains(requestComponent) == true || requestComponent.contains(it.splitPackageName!!))
```

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
```diff
@@ -48,16 +59,35 @@ suspend fun HttpClient.requestDownloadUrls(
...
+                payload = BulkRequestWrapper.build {
+                    request(BulkRequest.build {
+                        packageName(packageName)
+                        grant(BulkGrant.build { grantLevel = 1 })
+                    })
+                },
+                adapter = BulkResponseWrapper.ADAPTER
+        )
+    }
+
```
fynngodau: It seems inappropriate to do this here, as this method is also used not only by `SplitInstallManager`, but also by `InstallService`, which itself already makes sure that device info is synced before a download can be started. So please move up one layer.

I would also be interested in knowing what a bulk entitlement request does. Is it similar to purchasing a free app? Will it add the app that requested the package to the Play Store Library view?

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -39,71 +43,77 @@ private const val TAG = "SplitInstallManager"
 class SplitInstallManager(val context: Context) {
 
     private var httpClient: HttpClient = HttpClient()
+    private val mutex = Mutex()
 
     suspend fun splitInstallFlow(callingPackage: String, splits: List<Bundle>): Boolean {
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) return false
+        var packagesToDownload: List<String> = listOf()
+        var components:List<PackageComponent>? = null
+        mutex.withLock {
```
fynngodau, line 53: Just to make sure I get this right: this `mutex` ensures that we only have one split install download at the same time, correct? Sounds sensible. :slightly_smiling_face:

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
```diff
@@ -48,16 +59,35 @@ suspend fun HttpClient.requestDownloadUrls(
     }
     Log.d(TAG, "requestDownloadUrls languages: $languages")
 
+    val androidId = auth.gsfId.toLong(16)
     val headers = buildRequestHeaders(
         auth = auth.authToken,
-        // TODO: understand behavior. Using proper Android ID doesn't work when downloading split APKs
```
fynngodau, line 53: I'm guessing the reason it works now is related to the new behavior that syncs device data before download. Nice! :slightly_smiling_face:

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -25,126 +24,156 @@ import org.microg.vending.enterprise.InstallError
...
     val packageInstaller = packageManager.packageInstaller
-    val installed = packageManager.getInstalledPackages(0).any {
-        it.applicationInfo.packageName == packageName
-    }
     // Contrary to docs, MODE_INHERIT_EXISTING cannot be used if package is not yet installed.
     val params = SessionParams(
-        if (!installed || isUpdate) SessionParams.MODE_FULL_INSTALL
-        else SessionParams.MODE_INHERIT_EXISTING
+            if (isUpdate) SessionParams.MODE_FULL_INSTALL
+            else SessionParams.MODE_INHERIT_EXISTING
```
fynngodau: Here I had left the comment that "[c]ontrary to docs, `MODE_INHERIT_EXISTING` cannot be used if package is not yet installed." Are you positive that my observation is false, and that this change won't break installing packages from work app store? (Unfortunately I wasn't able to test this today due to certain issues.) It doesn't seem like it helps with split installs, as they should be unaffected, or does it?

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -25,126 +24,156 @@ import org.microg.vending.enterprise.InstallError
...
-        Log.d(TAG, "installPackages session commit")
-        return deferred.await()
+            Log.d(TAG, "installPackages session commit")
+            return deferred.await()
+        } else {
+            Log.d(TAG, "installPackagesInternal: The screen is locked and waiting for the user to click the notification to install")
+            // don't abandon if `finally` step is reached after this point
+            session = null
+            throw Exception("The device screen is off, waiting for installation")
+        }
```
fynngodau: This behavior breaks background installations that would otherwise work without user interaction. Shouldn't this rather be handled in the case `PackageInstaller.STATUS_PENDING_USER_ACTION` of `SessionResultReceiver`?

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -25,126 +24,156 @@ import org.microg.vending.enterprise.InstallError
...
+            isUpdate = isUpdate,
+            emitProgress = emitProgress,
+    ) {tempFiles, notifyId, fileName, to ->
         val component = components.find { it.componentName == fileName }!!
-        Log.v(TAG, "installing $fileName for $packageName from network")
-        // Emit progress for the first time as soon as possible, before any network interaction
-        emitProgress(session, Downloading(
-            bytesDownloaded = downloadProgress.values.sum(),
-            bytesTotal = components.sumOf { it.size }
-        ))
```
fynngodau, line 69: This behavior is expected by callers, which depend on it to promote themselves to a foreground service. If the notification is only emitted after network interaction, this can take too long, so the thread might be killed in the meantime.

DaVinci9196:
> Thanks for these improvements. I think resumable downloads, and only downloading one split install package at a time, really make sense. However, I would prefer a different implementation approach: the way you are doing it is that the packages are first downloaded to a temporary directory, and then copied to the install session from this directory. (A previous iteration of the code also had this behavior, but we changed it to avoid occupying 2× the space of the installed package while installing it.)
> 
> I think it should be possible to instead resume downloads into the pending session. To do this:
> 
> * We need a registry of session IDs, so we know which download corresponds to which session.
> * Sessions must not be discarded when download fails.
> * We don't need temporary files anymore.
> 
> Do you see any reason that speaks against implementing it like this?

We temporarily store the downloaded files locally to avoid repeated downloads in case of issues, which could lead to unnecessary data usage for users.
Regarding your suggestion of resuming directly into the install session — isn’t there a limitation on how many active sessions the system allows?
That’s part of the reason we initially chose to manage downloads outside the session system.

vending-app/src/main/java/org/microg/vending/ui/InstallProgressNotification.kt
DaVinci9196: > While the behavior here implemented is generally okay (for situations where microG is not privileged to install apps), this also changes the notification when the package is already installed. So this new type of notification should:
> 
> * be displayed if and only if user interaction is necessary
> * not show a progress indicator
> 
> For this it needs to be a separate progress state, I think.

Okay, does this approach sound good to you?
I added an internal data object: InstallPending : InstallProgress.

The reason for this is that when the screen is off, the installation prompt won't appear after the split package is downloaded.
As a result, the user has to trigger the split package service again in order to complete the installation.

So I’d like to show a notification in this case, allowing the user to manually trigger the installation by tapping on it.

fynngodau:
> Regarding your suggestion of resuming directly into the install session — isn’t there a limitation on how many active sessions the system allows?

I don't know if there is such a limit. What do you expect the limit to be? What happens if it is exceeded? Do you expect to open many sessions, i.e. many partial and failed downloads? This would also mean a large storage consumption in either implementation, so it's not really desirable in either implementation.

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
DaVinci9196: Okay, I’ll update it like that.

DaVinci9196: > Can add a comment why this is a good behavior? It seems like a rather lax matching – can we be sure that we download the correct package?
> 
> (Also, I think the following is more idiomatic and easier to understand:)

The reason I added this additional matching logic is because I encountered a case like this during development.
Unfortunately, it was quite a long time ago, so I can’t recall exactly which app it was.
But I remember its split package information was a bit different from the usual format.
That’s why I decided to include this fallback logic.

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
DaVinci9196: I'll move the syncDeviceInfo call to a different location.
As for the bulk entitlement request, we don’t know much about it either at this point.
What we do know is that without calling it, some accounts fail to retrieve split package information.
I also captured network traffic on other devices and noticed they invoke this API before making the actual request.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
DaVinci9196: > Just to make sure I get this right: this `mutex` ensures that we only have one split install download at the same time, correct? Sounds sensible. 🙂

yes

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
DaVinci9196: > This behavior breaks background installations that would otherwise work without user interaction. Shouldn't this rather be handled in the case `PackageInstaller.STATUS_PENDING_USER_ACTION` of `SessionResultReceiver`?

ok

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
DaVinci9196: > This behavior is expected by callers, which depend on it to promote themselves to a foreground service. If the notification is only emitted after network interaction, this can take too long, so the thread might be killed in the meantime.

ok

DaVinci9196:
> > Regarding your suggestion of resuming directly into the install session — isn’t there a limitation on how many active sessions the system allows?
> 
> I don't know if there is such a limit. What do you expect the limit to be? What happens if it is exceeded? Do you expect to open many sessions, i.e. many partial and failed downloads? This would also mean a large storage consumption in either implementation, so it's not really desirable in either implementation.

This is just an extreme case because Marvin initially raised concerns about data usage with me. That’s why I chose this approach.
Regarding your suggestion of tracking through sessions, wouldn’t the sessions be lost if the app is killed?

fynngodau:
> Regarding your suggestion of tracking through sessions, wouldn’t the sessions be lost if the app is killed?

The session will even be persisted across multiple boots, please refer to: https://developer.android.com/reference/android/content/pm/PackageInstaller#createSession(android.content.pm.PackageInstaller.SessionParams)

DaVinci9196:
> > Regarding your suggestion of tracking through sessions, wouldn’t the sessions be lost if the app is killed?
> 
> The session will even be persisted across multiple boots, please refer to: https://developer.android.com/reference/android/content/pm/PackageInstaller#createSession(android.content.pm.PackageInstaller.SessionParams)

So do we need to store the sessions locally to facilitate reuse later?
If we use this method, how can we track the split package download progress, and how can we implement resumable downloads?

DaVinci9196:
> > Regarding your suggestion of tracking through sessions, wouldn’t the sessions be lost if the app is killed?
> 
> The session will even be persisted across multiple boots, please refer to: https://developer.android.com/reference/android/content/pm/PackageInstaller#createSession(android.content.pm.PackageInstaller.SessionParams)

I think your solution still cannot meet the requirements for resumable downloads in special cases.

fynngodau:
Yes, as I mentioned, we could use a registry of session IDs, so we know which download corresponds to which session. Since we are session owner, we might also be able to use: https://developer.android.com/reference/kotlin/android/content/pm/PackageInstaller?hl=en#getStagedSessions() and compare the package name of the session.

I suppose we might need to store how many bytes we wrote to the session, if it's not possible to ask the session that, though it might be feasible to use: https://developer.android.com/reference/kotlin/android/content/pm/PackageInstaller.Session?hl=en#openRead(kotlin.String)

What special cases do you have in mind?

DaVinci9196:
> Yes, as I mentioned, we could use a registry of session IDs, so we know which download corresponds to which session. Since we are session owner, we might also be able to use: https://developer.android.com/reference/kotlin/android/content/pm/PackageInstaller?hl=en#getStagedSessions() and compare the package name of the session.
> 
> I suppose we might need to store how many bytes we wrote to the session, if it's not possible to ask the session that, though it might be feasible to use: https://developer.android.com/reference/kotlin/android/content/pm/PackageInstaller.Session?hl=en#openRead(kotlin.String)
> 
> What special cases do you have in mind?

I tried using your method, but I kept getting the following error when calling `openSession`: `java.lang.SecurityException: Caller has no access to session 2068134291`. @fynngodau

fynngodau:
@DaVinci9196 Can you provide me with the code where this issue occurs? That will make it much easier for me to find out why that could be the case.

DaVinci9196:
> @DaVinci9196 Can you provide me with the code where this issue occurs? That will make it much easier for me to find out why that could be the case.

[Install.txt](https://github.com/user-attachments/files/20786968/Install.txt)
I modified Install.kt to Install.txt and uploaded it。
The core code is as follows. An error occurs when calling openSession
```
val cacheSessionId = SessionStorage.getInstance(this).getSessionId(key)
        sessionId = if (cacheSessionId == -1) {
            packageInstaller.createSession(params)
        } else {
            cacheSessionId
        }
        for (info in packageInstaller.allSessions) {
            Log.d(TAG, "id=${info.sessionId}, createdBy=${info.installerPackageName}")
        }

        SessionStorage.getInstance(this).saveSessionId(key, sessionId)
        session = packageInstaller.openSession(sessionId)
```

DaVinci9196:
> > @DaVinci9196 Can you provide me with the code where this issue occurs? That will make it much easier for me to find out why that could be the case.
> 
> [Install.txt](https://github.com/user-attachments/files/20786968/Install.txt) I modified Install.kt to Install.txt and uploaded it。 The core code is as follows. An error occurs when calling openSession
> 
> ```
> val cacheSessionId = SessionStorage.getInstance(this).getSessionId(key)
>         sessionId = if (cacheSessionId == -1) {
>             packageInstaller.createSession(params)
>         } else {
>             cacheSessionId
>         }
>         for (info in packageInstaller.allSessions) {
>             Log.d(TAG, "id=${info.sessionId}, createdBy=${info.installerPackageName}")
>         }
> 
>         SessionStorage.getInstance(this).saveSessionId(key, sessionId)
>         session = packageInstaller.openSession(sessionId)
> ```

@fynngodau Can you take a look? This is quite urgent.

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
DaVinci9196: > Can add a comment why this is a good behavior? It seems like a rather lax matching – can we be sure that we download the correct package?
> 
> (Also, I think the following is more idiomatic and easier to understand:)

When I use PayPal, I use this judgment

fynngodau:
@DaVinci9196 I could not reproduce your problem, but I also do not have access to the `SessionStorage` class that you implemented. Hence I instead tried the following code (replacing lines 134–135 of `Install.kt`):

```
        sessionId = packageInstaller.createSession(params)
        session = packageInstaller.openSession(sessionId)
        packageInstaller.updateSessionAppLabel(sessionId, packageName)
        session.close()
        Log.d(TAG, "session closed")
        Log.d(TAG, packageInstaller.mySessions.joinToString(",") { sessionInfo -> sessionInfo.sessionId.toString() + " of package " + sessionInfo.appLabel })
        session = packageInstaller.openSession(sessionId)
        Log.d(TAG, "session reopened")
```

This worked without problems:

* `mySessions` returned the currently pending session
* I could access its metadata (this could be a good alternative to your `SessionStorage` implementation, as I would expect `mySessions` to always be in sync with whether the system has discarded the session due to timeout (after a day or so per docs))
* I was able to reopen the session (but I did not try with closing and reopening the app)

Does the above code already produce a problem for you?

I noticed that you are never closing the session. I would suggest adding this to the `finally` block in your version of the code, maybe it helps.

DaVinci9196:
> @DaVinci9196 I could not reproduce your problem, but I also do not have access to the `SessionStorage` class that you implemented. Hence I instead tried the following code (replacing lines 134–135 of `Install.kt`):
> 
> ```
>         sessionId = packageInstaller.createSession(params)
>         session = packageInstaller.openSession(sessionId)
>         packageInstaller.updateSessionAppLabel(sessionId, packageName)
>         session.close()
>         Log.d(TAG, "session closed")
>         Log.d(TAG, packageInstaller.mySessions.joinToString(",") { sessionInfo -> sessionInfo.sessionId.toString() + " of package " + sessionInfo.appLabel })
>         session = packageInstaller.openSession(sessionId)
>         Log.d(TAG, "session reopened")
> ```
> 
> This worked without problems:
> 
> * `mySessions` returned the currently pending session
> * I could access its metadata (this could be a good alternative to your `SessionStorage` implementation, as I would expect `mySessions` to always be in sync with whether the system has discarded the session due to timeout (after a day or so per docs))
> * I was able to reopen the session (but I did not try with closing and reopening the app)
> 
> Does the above code already produce a problem for you?
> 
> I noticed that you are never closing the session. I would suggest adding this to the `finally` block in your version of the code, maybe it helps.

This code cannot run on my phone and will report an error @fynngodau

DaVinci9196:
> and
@fynngodau I see the following code in the Delivery.kt class. What will happen if these codes are not included?
```
pf=2&pf=3&pf=4&pf=5&pf=7&pf=8&pf=9&pf=10&da=4&bda=4&bf=4&fdcf=1&fdcf=2
```
I have found that &da=4 will cause problems when PayPal switches to voice. Can this be removed?  @fynngodau

mar-v-in:
@DaVinci9196 I tried with just the code fragment from https://github.com/microg/GmsCore/pull/2898#issuecomment-3003730353 in a minimal demo app on multiple devices, including P60 Pro with EMUI 14 and Mate X5 with Harmony OS 4 and it worked as expected and reopened the session on all of them. What device did you use for this?

Looking at the code, I believe the issue might be a concurrency issue. Note that `installPackagesInternal` is a suspendable function and does not have any mutexes or anything preventing it from being invoked more than once. This means that a second call to `installPackagesInternal` will fail if it happens for the same package as one that is already ongoing.

DaVinci9196:
> @DaVinci9196 I tried with just the code fragment from [#2898 (comment)](https://github.com/microg/GmsCore/pull/2898#issuecomment-3003730353) in a minimal demo app on multiple devices, including P60 Pro with EMUI 14 and Mate X5 with Harmony OS 4 and it worked as expected and reopened the session on all of them. What device did you use for this?
> 
> Looking at the code, I believe the issue might be a concurrency issue. Note that `installPackagesInternal` is a suspendable function and does not have any mutexes or anything preventing it from being invoked more than once. This means that a second call to `installPackagesInternal` will fail if it happens for the same package as one that is already ongoing.

I have modified it according to this comment [#2898 (comment)](https://github.com/microg/GmsCore/pull/2898#issuecomment-3003730353), you can test it with the latest code

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
```diff
@@ -48,16 +59,31 @@ suspend fun HttpClient.requestDownloadUrls(
...
-        androidId = if (requestSplitPackages != null) 1 else auth.gsfId.toLong(16),
+        androidId = androidId,
         languages
     ).minus(
         // TODO: understand behavior. According to tests, these headers break split install queries but may be needed for normal ones
         (if (requestSplitPackages != null) listOf("X-DFE-Encoded-Targets", "X-DFE-Phenotype", "X-DFE-Device-Id", "X-DFE-Client-Id") else emptyList()).toSet()
     )
 
+    kotlin.runCatching {
```
mar-v-in, line 72: This is only needed when requesting split packages, no? In this case it should also only be called `if (requestSplitPackages != null)`

vending-app/src/main/kotlin/com/android/vending/installer/SessionResultReceiver.kt
```diff
@@ -1,25 +1,25 @@
-/*
- * SPDX-FileCopyrightText: 2025 e foundation
- * SPDX-License-Identifier: Apache-2.0
- */
-
```
mar-v-in: Should not remove copyright header here either

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -2,7 +2,6 @@
  * SPDX-FileCopyrightText: 2025 e foundation
  * SPDX-License-Identifier: Apache-2.0
  */
-
```
mar-v-in: No reason to remove this empty line.

vending-app/src/main/kotlin/com/google/android/finsky/splitinstallservice/SplitInstallManager.kt
```diff
@@ -1,26 +1,31 @@
 /**
- * SPDX-FileCopyrightText: 2024 microG Project Team
+ * SPDX-FileCopyrightText: 2025 microG Project Team
```
mar-v-in: No need to update copyright header, we keep it at the year first added.

mar-v-in (changes requested):
Some minor nits to be fixed.

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
DaVinci9196: The original code is like this, I am not good at making changes

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +33,193 @@ import java.io.OutputStream
...
-        Log.e(TAG, "Error installing packages", e)
-        emitProgress(sessionId, InstallError(e.message ?: "UNKNOWN"))
+    } catch (e: Exception) {
+        Log.w(TAG, "Error installing packages", e)
+        emitProgress(notifyId, InstallError("ERROR"))
         throw e
     } finally {
         // discard downloaded data
         session?.let {
             Log.d(TAG, "Discarding session after error")
```
fynngodau: This log line is no longer accurate.

DaVinci9196: Modified

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +33,193 @@ import java.io.OutputStream
...
 }
+
+private fun Context.computeUniqueKey(packageName: String, componentNames: List<String>) : String {
+    val versionCode = PackageInfoCompat.getLongVersionCode (
+            packageManager.getPackageInfo(packageName, 0)
+    )
+    val key = componentNames.joinToString(separator = "_", prefix = "${packageName}_${versionCode}")
+    Log.d(TAG, "computeUniqueKey: $key")
+    return key
+}
```
fynngodau: This function doesn't work for packages that are not yet installed, instead throwing `PackageManager$NameNotFoundException`. Please handle this case – from my tests it works to simply not consider the version code for the key in that case.

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +33,193 @@ import java.io.OutputStream
...
+                } catch (e: Exception) {
+                    Log.w(TAG, "Error writing component notifyId $notifyId")
+                    emitProgress(notifyId, InstallError("Download Error"))
+                    throw e
+                }
             }
         }
         val deferred = CompletableDeferred<Unit>()
-
+        Log.e(TAG, "installPackagesInternal pendingSessions size: ${SessionResultReceiver.pendingSessions.size}")
```
fynngodau: Please make this a lower log level than "error".

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +33,193 @@ import java.io.OutputStream
...
+            Log.d(TAG, "id=${info.sessionId}, createdBy=${info.appLabel}")
+        }
+        Log.d(TAG, "installPackagesInternal sessionId: $sessionId")
         session = packageInstaller.openSession(sessionId)
+
         for (component in componentNames) {
+            val currentSize: Long = try {
+                val inputStream = session.openRead(component)
+                val available = withContext(Dispatchers.IO) {
+                    inputStream.available()
```
fynngodau: Resuming downloads does not work in my tests. It could be caused by the use of `InputStream.available()` here, which is documented to _not_ do what we want:

> Note that while some implementations of `InputStream` will return the total number of bytes in the stream, many will not. It is never correct to use the return value of this method to allocate a buffer intended to hold all data in this stream.

So this function need not actually return the size of the input stream; for this it is necessary to manually read it (using `readNBytes(byte[], int, int)` seems like the best way).

fynngodau (changes requested):
Thanks for the improvements to this PR so far. However, it does not fully work yet from my tests.

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
DaVinci9196: readNBytes has requirements for SDK version
Is it okay to write the following?
```
val currentSize: Long = try {
                val inputStream = session.openRead(component)
                val totalSize = withContext(Dispatchers.IO) {
                    val buffer = ByteArray(4096)
                    var total = 0L
                    var bytesRead: Int
                    while (inputStream.read(buffer).also { bytesRead = it } != -1) {
                        total += bytesRead
                    }
                    inputStream.close()
                    total
                }
                totalSize
            } catch (e: IOException) {
                Log.w(TAG, "installPackagesInternal", e)
                0L
            }
```

fynngodau: @DaVinci9196 Yes, I believe that this code is accurate.

DaVinci9196: Modified，I'm always online, if you have any other suggestions, you can always tell me

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +34,204 @@ import java.io.OutputStream
...
+                    total
+                }
+                totalSize
+            } catch (e: IOException) {
+                Log.w(TAG, "installPackagesInternal", e)
+                0L
+            }
+
+            Log.d(TAG, "installPackagesInternal component: $component currentSize:$currentSize")
             session.openWrite(component, 0, -1).use { outputStream ->
```
fynngodau: Resuming downloads cannot possibly work if you do not provide the correct offset here.

fynngodau:
> I have found that &da=4 will cause problems when PayPal switches to voice. Can this be removed?

My answer is: nobody really knows what these options mean. I see no regressions caused by removing it so I consider it ok.

vending-app/src/main/java/org/microg/vending/delivery/Delivery.kt
fynngodau: I agree that this request should be moved up one layer. I previously remarked the same [up here](https://github.com/microg/GmsCore/pull/2898#discussion_r2147483929).

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
DaVinci9196: Modified

quimodotcom:
0.3.9 here we come!

mar-v-in:
LGTM, Thanks @DaVinci9196 and @fynngodau for the work!

vending-app/src/main/kotlin/com/android/vending/installer/Install.kt
```diff
@@ -30,127 +34,204 @@ import java.io.OutputStream
...
+            val currentSize: Long = try {
+                val inputStream = session.openRead(component)
+                val totalSize = withContext(Dispatchers.IO) {
+                    val buffer = ByteArray(4096)
+                    var total = 0L
+                    var bytesRead: Int
+                    while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                        total += bytesRead
+                    }
+                    inputStream.close()
```
ale5000-git, line 158: @mar-v-in
Wouldn't make sense to put `inputStream.close()` inside a `finally` block to be sure it is always closed even in case of unknown errors.

PR #2900, opened 2025-05-08
Measurement: Make stub app instance id 16 byte hex string

DaVinci9196:
Generate app instance id。Solution to APP not working
app : AI Chatbot -Nova
packagename:com.scaleup.chatai
Problem phenomenon: Open the app and enter the homepage. After entering text, an error message appears


play-services-core/src/main/kotlin/com/google/android/gms/measurement/internal/AppMeasurementDynamiteService.kt
```diff
@@ -132,7 +132,7 @@ class AppMeasurementDynamiteService : IAppMeasurementDynamiteService.Stub() {
         Log.d(TAG, "Not yet implemented: getAppInstanceId")
         // Generate a random ID -> equivalent to ephemeral app instance id
         // Correct behavior would be to generate appropriate AppMetadata and call IMeasurementService.getAppInstanceId
-        val ephemeralAppInstanceId = Random.nextBytes(32).toHexString("")
```
mar-v-in, line 135: Wouldn't it be effectively the same to just change 32 for 16 here?

DaVinci9196: He needs a 32-bit string.

DaVinci9196: Here you need to return a string of length 32

mar-v-in: Yes, I understood so far. The bug here was that we generate a hex string of 32 bytes which has a length of 64 chars, instead of a hex string of 16 bytes, that would use 32 chars. But we don't need the new function to generate 16 bytes hex string, the current could already do that.

mar-v-in (approved):

PR #2908, opened 2025-05-13
DynamicLink: Added file link processing

DaVinci9196:
Fixed the issue where opening a file using an application from a local file manager would not respond. For example: using Polaris Office to open PDF/DOCX files

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/firebase/dynamiclinks/DynamicLinksService.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/firebase/dynamiclinks/DynamicLinksService.kt
@@ -52,7 +52,7 @@ class DynamicLinksServiceImpl(private val context: Context, private val callingP
         Log.d(TAG, "getDynamicLink: callingPackageName: $callingPackageName link: $link")
         if (link != null) {
             val linkUri = Uri.parse(link)
-            if ("content".equals(linkUri.scheme, ignoreCase = true)) {
+            if ("content" == linkUri.scheme) {
                 Log.d(TAG, "getDynamicLink: $link -> null")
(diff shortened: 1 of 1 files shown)


play-services-core/src/main/kotlin/org/microg/gms/firebase/dynamiclinks/DynamicLinksService.kt
```diff
@@ -50,42 +51,40 @@ class DynamicLinksServiceImpl(private val context: Context, private val callingP
...
-                            response.metadata?.info?.url, response.data_?.intentData, (response.data_?.app?.minAppVersion ?: 0).toInt(), System.currentTimeMillis(), null, null
-                        )
+                        ?: throw RuntimeException("requestLinkResponse failed")
+                    val data = DynamicLinkData(link, response.data_?.intentData, (response.data_?.app?.minAppVersion ?: 0).toInt(), System.currentTimeMillis(), null, null)
+                    Log.d(TAG, "getDynamicLink: $link -> $data")
+                    callback.onStatusDynamicLinkData(Status.SUCCESS, data)
+                    return@launchWhenCreated
+                }
+                val deepLink = linkUri.getQueryParameter("link")
+                if (!TextUtils.isEmpty(deepLink)) {
```
mar-v-in: Please use Kotlins `String.isNullOrEmpty()` instead.

mar-v-in (changes requested):

play-services-core/src/main/kotlin/org/microg/gms/firebase/dynamiclinks/DynamicLinksService.kt
DaVinci9196: changed

mar-v-in (approved):

PR #2909, opened 2025-05-14
Vending: Add deep link handler

DaVinci9196:
Fixed the issue that some apps prompt "not available" when redirecting to the page with the specified path.
For example: Youtube clicks on the app content that needs to be installed.


mar-v-in (commented):
Couldn't you just use an [`activity-alias`](https://developer.android.com/guide/topics/manifest/activity-alias-element) instead?

DaVinci9196:
> Couldn't you just use an [`activity-alias`](https://developer.android.com/guide/topics/manifest/activity-alias-element) instead?

Thanks for the reminder.

mar-v-in (approved):

PR #2927, opened 2025-06-01
Fix the names of the components in the self-check

ale5000-git:
@mar-v-in

Fix the names of the components in the self-check.
How it was before it only cause confusion.

See also: https://github.com/microg/GsfProxy/pull/25

### DIFF
--- a/play-services-core/src/main/res/values/strings.xml
+++ b/play-services-core/src/main/res/values/strings.xml
@@ -128,9 +128,9 @@ Please set up a password, PIN, or pattern lock screen."</string>
     <string name="self_check_name_system_spoofs">System spoofs signature: </string>
     <string name="self_check_resolution_system_spoofs">Please check the documentation on which steps might be required.</string>
 
-    <string name="self_check_pkg_gms">Play Services (GmsCore)</string>
-    <string name="self_check_pkg_vending">Play Store (Phonesky)</string>
-    <string name="self_check_pkg_gsf">Services Framework (GSF)</string>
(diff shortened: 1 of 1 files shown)


mar-v-in:
The companion app can be replaced with regular play store (even if that is not officially supported), so companion is not strictly needed to use microG (and the check will be checked even if it's not installed, but regular Play Store).

My idea with the self-check is to completely replace it with:
- Dynamic permission request at the time they are actually needed (and thereby stopping to ask users to prematurely grant permissions without an obvious reason)
  - We already have that for most permissions, this really mostly needs some cleanup to ensure it is really everywhere and works as expected
- Warnings, both on the main screen of microG settings and in form of notification, in case of misconfiguration

ale5000-git:
> The companion app can be replaced with regular play store (even if that is not officially supported)

I know but this is only for advanced users (that already know that); normal users that maybe have just installed microG Services just feel that they have to install Play Store (and not microG Companion) because the self check say this.

We can maybe change it to `microG Companion / Play Store`.
What do you think?

---

> My idea with the self-check is to completely replace it with:
> 
> * Dynamic permission request at the time they are actually needed (and thereby stopping to ask users to prematurely grant permissions without an obvious reason)
>   
>   * We already have that for most permissions, this really mostly needs some cleanup to ensure it is really everywhere and works as expected
> * Warnings, both on the main screen of microG settings and in form of notification, in case of misconfiguration

These are good but I still think there should also be a self check to see the global situation all at once (maybe inside an hidden menu that can be opened by tapping the microG version multiple times).

PR #2954, opened 2025-06-24
Fix the issue where third-party apps encounter "Invalid argument for …

DaVinci9196:
Fix the issue where third-party apps encounter "Invalid argument for the given fid" when registering with GCM.
**Reproduction Steps:**

1. In microG, allow a third-party app (e.g., WhatsApp, Messenger) to register with GCM.
2. Go to the app info page and clear microG’s app data. At this point, the previously registered GCM entries are lost, and the Android ID is reset.
3. When the third-party app triggers GCM registration again, an error occurs:
   **“Invalid argument for the given fid”**, causing GCM registration to fail.

 Common triggers for re-registration include:

   * Updating the third-party app
   * Reopening the third-party app

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/gcm/PushRegisterService.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/gcm/PushRegisterService.kt
@@ -81,7 +81,7 @@ suspend fun completeRegisterRequest(context: Context, database: GcmDatabase, req
     PushRegisterManager.completeRegisterRequest(context, database, requestId, request) {
         val errorMsg = it.getString(EXTRA_ERROR)
         Log.w(TAG, "completeRegisterRequest error: $errorMsg")
-        if (errorMsg == ERROR_INVALID_FID && !request.delete) {
+        if (errorMsg == PushRegisterManager.attachRequestId(ERROR_INVALID_FID, requestId) && !request.delete) {
             Log.d(TAG, "completeRegisterRequest register error, You need to call delete first before you can re-register")
             request.delete = true
             request.response


mar-v-in:
I worry that this could accidentally unregister apps from push notifications, potentially doing more harm than use in some cases. Could you tie this unregister+reregister to this specific error message instead?

play-services-core/src/main/kotlin/org/microg/gms/gcm/PushRegisterService.kt
```diff
@@ -78,7 +78,21 @@ private suspend fun ensureAppRegistrationAllowed(context: Context, database: Gcm
 }
 
 suspend fun completeRegisterRequest(context: Context, database: GcmDatabase, request: RegisterRequest, requestId: String? = null): Bundle = suspendCoroutine { continuation ->
-    PushRegisterManager.completeRegisterRequest(context, database, requestId, request) { continuation.resume(it) }
+    PushRegisterManager.completeRegisterRequest(context, database, requestId, request) {
+        val errorMsg = it.getString(EXTRA_ERROR)
+        Log.w(TAG, "completeRegisterRequest error: $errorMsg")
+        if (errorMsg == ERROR_INVALID_FID && !request.delete) {
```
mar-v-in: If `requestId` is non-null, the value in the error field will be prefixed with `|ID|$requestId|`, so you should compare errorMsg against `PushRegisterManager.attachRequestId(ERROR_INVALID_FID, requestId)`

DaVinci9196: Updated

PR #2958, opened 2025-06-30
Phenotype: About using Gemini in Gmail

DaVinci9196:
Added Phenotype configuration to use Gemini related functions in Gmail menu bar.
The requestGoogleAccountsAccess method needs to be adapted.

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
+++ b/play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
@@ -49,6 +49,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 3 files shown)


play-services-core/src/main/java/org/microg/gms/auth/AuthManagerServiceImpl.java
```diff
@@ -216,10 +217,15 @@ public Bundle removeAccount(Account account) {
...
+        Log.d(TAG, "requestGoogleAccountsAccess: " + packageName);
         PackageUtils.assertGooglePackagePermission(context, GooglePackagePermission.ACCOUNT);
         if (SDK_INT >= 26) {
+            Map<Account, Integer> visibilityForPackage = get(context).getAccountsAndVisibilityForPackage(packageName, AuthConstants.DEFAULT_ACCOUNT_TYPE);
             for (Account account : get(context).getAccountsByType(AuthConstants.DEFAULT_ACCOUNT_TYPE)) {
-                AccountManager.get(context).setAccountVisibility(account, packageName, VISIBILITY_VISIBLE);
+                Integer visibility = visibilityForPackage.get(account);
+                if (visibility != null && visibility != VISIBILITY_VISIBLE) {
+                    get(context).setAccountVisibility(account, packageName, VISIBILITY_VISIBLE);
+                }
```
mar-v-in, line 227: What is the purpose of this change? I don't see why it makes a difference.

Old code is rougly:
```
setAccountVisibility(account, packageName, VISIBILITY_VISIBLE);
```

New code is rougly:
```
if (getAccountVisibility(account, packageName) != VISIBILITY_VISIBLE)
  setAccountVisibility(account, packageName, VISIBILITY_VISIBLE);
```

The result should effectively be the same. In both cases it is set to VISIBILITY_VISIBLE afterwards, no matter what it was set before.

play-services-core/src/main/kotlin/org/microg/gms/phenotype/extensions.kt
```diff
@@ -0,0 +1,19 @@
...
+    "iw", "he", "tr", "cs", "id", "in", "sv", "da", "no",
+    "nb", "pl", "vi", "th", "fi", "uk", "ar", "el", "ru",
+    "hu", "ro", "ca"
+)
+
+fun decodeLanguageList(): ByteArray {
+    return supportedLanguages.flatMap { lang ->
+        listOf(0x0A.toByte(), 0x02.toByte()) + lang.toByteArray(Charsets.UTF_8).toList()
+    }.toByteArray()
+}
```
mar-v-in: This is an encode function, not a decode function. Specifically, it encodes a protobuf message that has a single repeated field with id 1:
```proto
message StringList {
  repeated string code = 1;
}
```

DaVinci9196: Because repeatedly granting permissions to an app that already has access can cause its locally stored configuration data to be passively cleared.
Tested with the Gmail app.

mar-v-in (approved):

PR #2984, opened 2025-07-28
Wear: Avoiding application crashes

DaVinci9196:
Added an empty page to the protocol to avoid crashes when the application calls the page.
e.g. Montblanc Summit<com.montblanc.summit.companion.android>

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -441,6 +441,16 @@
             </intent-filter>
         </service>
 
+        <activity
+            android:name="com.google.android.gms.wearable.consent.TermsOfServiceActivity"
+            android:process=":ui"
+            android:configChanges="screenSize|orientation|keyboardHidden">
+            <intent-filter>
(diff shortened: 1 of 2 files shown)


ale5000-git:
@DaVinci9196
Isn't possible to simulate the TOS being accepted instead of RESULT_CANCELED?

mar-v-in:
That would likely require actually implementing it. This is just to make sure apps don't crash.

ale5000-git:
@mar-v-in
Doesn't it work to return ok without really displaying anything?

mar-v-in:
It will certainly not mean the Wear implementation in microG is finished. And we have to review the consequences it may have to return ok without actual user consent (it might be better to not give consent as to not cause any data to be sent to Google or elsewhere).

ale5000-git:
> It will certainly not mean the Wear implementation in microG is finished. And we have to review the consequences it may have to return ok without actual user consent (it might be better to not give consent as to not cause any data to be sent to Google or elsewhere).

It may be a setting of microG "Auto-approve TOS" like the one to auto-buy free apps.

PR #3018, opened 2025-08-16
Fix the issue where location updates fail to callback when `maxUpdate…

DaVinci9196:
Fix the issue where location updates fail to callback when `maxUpdateAgeMillis` is set to 0.
eg:
app:Navionics® Boating
packagename:it.navionics.singleAppMarineLakesHD

https://github.com/user-attachments/assets/772b3ddb-a691-4926-9986-233b03af3375


mar-v-in:
According to the [documentation](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest#getMaxUpdateAgeMillis()), a `maxUpdateAgeMillis` "value of 0 indicates that no initial historical location will be delivered, only freshly derived locations". This means the proposed change will break the behavior documented and instead give 0 the same meaning as `Long.MAX_VALUE`.

The underlying issue was that in `LocationRequestManager.processNewLocation`, `maxUpdateDelayMillis` was wrongly passed for `maxUpdateAgeMillis`, which has a completely different meaning.
https://github.com/microg/GmsCore/blob/b3b3bc6a5fc9efaf97b25e74a0e2d42f9572a583/play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LocationRequestManager.kt#L151

I fixed this in 4b77c9785f3bf7daba1d5fe78a31d836a394473c.

PR #3022, opened 2025-08-18
Family Sharing: Add settings

DaVinci9196:
Entry: Modification of family sharing settings for YouTube or YouTube Music membership services

### DIFF
--- a/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsContract.kt
+++ b/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsContract.kt
@@ -277,6 +277,8 @@ object SettingsContract {
         const val BILLING = "vending_billing"
         const val ASSET_DELIVERY = "vending_asset_delivery"
         const val ASSET_DEVICE_SYNC = "vending_device_sync"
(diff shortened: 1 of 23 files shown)


play-services-core/src/main/kotlin/org/microg/gms/googlehelp/ui/GoogleHelpRedirectActivity.kt
```diff
@@ -46,7 +50,7 @@ class GoogleHelpRedirectActivity : AppCompatActivity() {
             finish()
             return
         }
-        val callingPackage = callingPackage ?: callingActivity?.packageName ?: return finish()
+        val callingPackage = intent.getStringExtra(KEY_PACKAGE_NAME) ?: callingPackage ?: callingActivity?.packageName ?: return finish()
```
mar-v-in, line 53: This means every app can claim to be a different app by adding an extra parameter to the start intent. This is bad.

mar-v-in: In the end, the calling activity has little impact on what we do here, so it's actually acceptable (although not a perfect solution).

PR #3031, opened 2025-08-23
Fido: Add credential selection for discoverable keys

DaVinci9196:
- Add missing parameters.

- Fix the failure of third-party application verification key. Contains: https://github.com/microg/GmsCore/pull/2885

- Fixed the issue that the created key cannot be used, and the latest key is created based on the existing code.

- Implement multiple passkey switching logins. Similar logic to https://github.com/microg/GmsCore/pull/2463  & Contains: https://github.com/microg/GmsCore/pull/3000

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,14 @@ jobs:
       GRADLE_MICROG_VERSION_WITHOUT_GIT: 1
 
     steps:
+      - name: "Free disk space"
+        run: |
+          # Deleting unneeded software packages
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          sudo rm -rf /opt/hostedtoolcache/go
+
+          # Log available space
+          df -h
       - name: "Checkout sources"
         uses: actions/checkout@v5
         with:


play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
```diff
@@ -213,21 +217,23 @@ fun getApplicationName(context: Context, options: RequestOptions, callingPackage
 }
 
 fun getApkKeyHashFacetId(context: Context, packageName: String): String {
-    val digest = context.packageManager.getFirstSignatureDigest(packageName, "SHA1")
+    // Default: SHA-256
+    val digest = context.packageManager.getFirstSignatureDigest(packageName, "SHA-256")
         ?: throw RequestHandlingException(NOT_ALLOWED_ERR, "Unknown package $packageName")
     return "android:apk-key-hash:${digest.toBase64(HASH_BASE64_FLAGS)}"
```
mar-v-in: This seems wrong. It should be either `android:apk-key-hash:<sha-1>` or `android:apk-key-hash-sha256:<sha-256>`, but not a mix.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
```diff
@@ -213,21 +217,23 @@ fun getApplicationName(context: Context, options: RequestOptions, callingPackage
...
+
     return when (facetId) {
-        "android:apk-key-hash:${firstSignature.digest("SHA1").toBase64(HASH_BASE64_FLAGS)}" -> {
-            "android:apk-key-hash-sha256:${firstSignature.digest("SHA-256").toBase64(HASH_BASE64_FLAGS)}"
-        }
-        "android:apk-key-hash-sha256:${firstSignature.digest("SHA-256").toBase64(HASH_BASE64_FLAGS)}" -> {
-            "android:apk-key-hash:${firstSignature.digest("SHA1").toBase64(HASH_BASE64_FLAGS)}"
-        }
+        "android:apk-key-hash:$sha1BASE64" -> "android:apk-key-hash-sha256:$sha256BASE64"
+        "android:apk-key-hash:$sha256BASE64" -> "android:apk-key-hash-sha256:$sha256BASE64"
```
mar-v-in: And you need this here, because you wrongly created it above.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
Nuyoah-Yc: I also noticed this, but in the fido specification, the pr server recognizes SHA-256

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
DaVinci9196: Removed sha-1 related.

PR #3036, opened 2025-08-26
Blockstore: Add local-only implementation

DaVinci9196:
The service some method is simply implemented, which is only for local storage on the same device and cannot be used across devices. 
Verification function: When logging in to WhatsApp & Uber , it will automatically prompt the account that has been logged in.


lucasmz-dev:
> The service some method is simply implemented, which is only for local storage on the same device and cannot be used across devices.

So this wouldn't work with Seedvault? It uses device-to-device transfers for backups.
Didn't even know this API was a thing

grote:
According to https://developers.google.com/android/reference/com/google/android/gms/auth/blockstore/Blockstore and https://developer.android.com/identity/block-store this does seem related to device-to-device transfers. However, the related methods don't seem to be implemented in this PR. There's probably a lot more work to do across MicroG and Seedvault before this actually works.

PR #3039, opened 2025-08-28
PI: New parameter content

DaVinci9196:
1. Supplementary equipment recommendation type. ( This can solve the problem that the new version X cannot enter the account creation )
2. Optimize exception handling.


ale5000-git:
@mar-v-in
There seems to be various issues here: https://github.com/microg/GmsCore/actions/runs/17975532467?pr=3039
Maybe something has gone wrong in the merge.

PR #3052, opened 2025-09-11
Location: Implement TGV Lyria's new position provider

TimoWilken:
The old position endpoint does not work anymore.

I don't know if the other TGVs also run this new onboard system, so I have only moved the TGV Lyria provider over.


ale5000-git:
@mar-v-in
Would it make sense to first try the new position provider and then fallback to the old one if the new doesn't exist?

mar-v-in:
I assume that all TGV Lyria are using the same system. If not, we can already have multiple entries in the `MOVING_WIFI_HOTSPOTS_LOCALLY_RETRIEVABLE` listing (e.g. `OEBB` has two different backends with the same SSID).

ale5000-git:
I'm just worrying in case one may still use an outdated version of it but I sure don't know.

TimoWilken:
Hi, as far as I can tell, all of the TGV Lyrias use the same new system, at least on the Paris-Geneva-Lausanne route. I haven't taken one on the Paris-Basel-Zurich route. Though for a long time, the moving map wasn't working at all with only a "coming soon" message as they had apparently switched to a new WiFi portal.

For other TGVs, e.g. inoui, the old system still seems to be in use - at least that was working on a TGV from Lille to Marseille that I took a few days ago.

Either way, up to you!

Issue #2382, opened 2024-05-22
[com.garmin.android.apps.connectmobile] Problem with maps in Garmin Connect since MicroG 0.2.28


jw243:
**Affected app**
Name: Garmin Connect
Package id: com.garmin.android.apps.connectmobile

**Describe the bug**
Since MicroG 0.2.28 maps on activity page are not displayed properly.

**To Reproduce**
Steps to reproduce the behavior in MicroG 0.3.2:
1. Open Garmin Connect.
2. Open a recorded activity.
3. Overview map is not displayed.
4. Click on "Stats", wait until all stats are loaded and scrolling through stats is smooth.
5. Go back to "Overview".
6. Now, overview map is displayed, but the marked GPS track is too thick.

**Expected behavior**
When opening a recorded activity in Garmin connect, the overview map should be displayed immediately and the marked GPS track should be thinner.

**Screenshots**
MicroG 0.3.2: Map not showing upon opening of an activity:
![0 3 2 - map not showing upon opening](https://github.com/microg/GmsCore/assets/13139555/749d3638-bcf9-44bc-9d19-46e4f9eebc6a)

MicroG 0.3.2: GPS track too thick on map once map appears:
![0 3 2 - GPS track too thick](https://github.com/microg/GmsCore/assets/13139555/c0f276b5-4167-4afd-9d3f-ca0524873352)

MicroG 0.2.27: Map opens immediately and shows correct part of the map with thinner GPS track:
![0 2 27 - map shows immediately with thinner GPS track](https://github.com/microg/GmsCore/assets/13139555/e3e54196-4471-4c21-8b8e-e7a63463bce7)

**System**
Android Version: 11
Custom ROM: LineageOS 18.1 (happens on other Lineage versions, too)

**microG**
microG Core version: from 0.2.28.231657 to 0.3.2.240913
microG Self-Check results: All ticked

**Additional context**
Logcat using MicroG 0.3.2:
[logcat-0.3.2.txt](https://github.com/microg/GmsCore/files/15408650/logcat-0.3.2.txt)

Logcat using MicroG 0.2.27:
[logcat-0.2.27.txt](https://github.com/microg/GmsCore/files/15408651/logcat-0.2.27.txt)

bagnaram:
On v 3.2.240813 it crashes entirely when opening an activity containing a map on the screen.

jw243:
[...]

PR #3058, opened 2025-09-14
mapbox: Implement Z-index and polyline spans handling

depau:
Hi!
This is a preliminary PoC implementation that adds handling of Z-index and polyline spans to the MapBox map implementation.

It isn't thoroughly tested, though it should fix some of the annoying glitches reported in #2382. Primarily it fixes my annoyance with Garmin Connect not actually displaying the run stats on the map.

I'm not a huge fan of the implementation, I basically did the least amount of work to get the feature going as a proof of concept to make sure this was even possible with the MapLibre API. If you'd like to integrate it I'm open to guidance and suggestions on how to properly implement it.

I also implemented JointType handling.

Before: 
image

After:

image

Website (just as a reference, I currently don't have a device with GApps to test):

image

The reason why the colors are slightly different in this image is that I ran multiple times in the same spots. I'll need to compare with the Google Android implementation since I think their web implementation is a bit messy.

EDIT:

Google Maps vs microG:

image

I think the result is very acceptable.

### DIFF
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
@@ -20,49 +20,99 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Bitmap
 import android.location.Location
-import android.os.*
-import androidx.annotation.IdRes
-import androidx.annotation.Keep
+import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
+import android.os.Parcel
 import android.util.Log
 import android.view.Gravity
 import android.view.View
 import android.widget.FrameLayout
 import android.widget.RelativeLayout
+import androidx.annotation.IdRes
+import androidx.annotation.Keep
 import androidx.collection.LongSparseArray
 import com.google.android.gms.dynamic.IObjectWrapper
 import com.google.android.gms.dynamic.ObjectWrapper
+import com.google.android.gms.dynamic.unwrap
+import com.google.android.gms.maps.GoogleMap
 import com.google.android.gms.maps.GoogleMapOptions
-import com.google.android.gms.maps.internal.*
-import com.google.android.gms.maps.model.*
+import com.google.android.gms.maps.internal.ICancelableCallback
+import com.google.android.gms.maps.internal.ILocationSourceDelegate
+import com.google.android.gms.maps.internal.IOnCameraChangeListener
+import com.google.android.gms.maps.internal.IOnCameraIdleListener
+import com.google.android.gms.maps.internal.IOnCameraMoveCanceledListener
+import com.google.android.gms.maps.internal.IOnCameraMoveListener
+import com.google.android.gms.maps.internal.IOnCameraMoveStartedListener
+import com.google.android.gms.maps.internal.IOnMapLoadedCallback
+import com.google.android.gms.maps.internal.IOnMapReadyCallback
+import com.google.android.gms.maps.internal.IOnMarkerDragListener
+import com.google.android.gms.maps.internal.IProjectionDelegate
+import com.google.android.gms.maps.internal.ISnapshotReadyCallback
+import com.google.android.gms.maps.internal.IUiSettingsDelegate
+import com.google.android.gms.maps.model.CameraPosition
 import com.google.android.gms.maps.model.CircleOptions
-import com.google.android.gms.maps.model.internal.*
+import com.google.android.gms.maps.model.GroundOverlayOptions
+import com.google.android.gms.maps.model.LatLng
+import com.google.android.gms.maps.model.LatLngBounds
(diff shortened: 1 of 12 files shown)


depau:
A little explanation is probably necessary to understand some of the main choices:

- To add multiple line segments I could have either split polylines into multiple polyline objects or, as I did, introduce support for multiple annotations into `Markup`. I went for the latter since I thought it would be the least effort.
- I didn't put too much effort into cleaning up the code that performs polyline updates, it's probably not very efficient
- With the version of MapLibre in use the only way to control the stacking of annotations is to create multiple managers, since each gets signed its layer id and since we can specify above/below which other layers it should go.
- The routine that fetches the correct Line/Segment/Fill Manager for the given Z index uses a TreeMap to compute the layers stacking. The idea is that the underlying tree's inherent ordering properties make it easy to insert a layer at any position in the layer stack, determining the layer above/below quickly. The situation might get a bit messy when apps insert layers like crazy but my intuition is that the layer ordering should stay visually consistent.

depau:
I finally implemented geodesic polyline. Disclaimer: for this part I used AI extensively since I don't fully understand the math. If you'd rather exclude this let me know.

image

At this point I don't think there's a lot more functionality we're missing that can be implemented. Tile overlays are basically impossible to add, and respecting the user's requested start/end cap style is very cumbersome with the MapLibre API. So I think I'm done, once somebody gets the change to take a look I can clean it up and get it ready for upstreaming if you'd like to get these changes in.

depau:
> Tile overlays are basically impossible to add,

Someone asked me over email why this is the case. As I was gathering the info to provide an answer I actually found a potential hacky way to implement it. Maybe next month if I have some time I can play with the idea.

This is my response I gave the person, for whoever may be interested:

> [...]
> 
> I'm not sure whether the docs I'm providing are for the exact corresponding version of MapLibre but they still give an idea of how things work.
> 
> The way you add overlays in MapLibre/MapBox is by adding layers; the relevant layer is the RasterLayer, which can load images from various sources. However, the source can either be a bunch of static files you provide ahead of time, or a web server that serves imagery over HTTP. 
> 
> On the other hand a Google Maps v2 API tile provider is a very simple class that implements a method that must return a tile given x+y+zoom.
> 
> A few links:
> 
> - https://docs.mapbox.com/android/maps/api/9.5.1/
> - https://docs.mapbox.com/android/maps/guides/styles/work-with-layers/
> - https://docs.mapbox.com/style-spec/reference/sources/
> - https://developers.google.com/android/reference/com/google/android/gms/maps/model/TileProvider
> 
> The reason why I think it might be possible is because I found out you can actually implement your own HTTP transport for HTTP tiles. So it may be possible to implement a custom HTTP transport that, when requesting images from an unlikely domain name, forges a response by calling the GMaps TileProvider locally, while forwarding the request to a proper HTTP client otherwise.
> 
> https://github.com/maplibre/maplibre-native/blob/main/platform/android/MapLibreAndroidTestApp/src/main/java/org/maplibre/android/testapp/utils/ExampleHttpRequestImpl.kt

fynngodau:
@depau I had come to a similar rough conclusion w.r.t. implementing tile overlays, so overall I agree with this approach. Thanks for giving it further thought! Please do it in a separate PR though 😃

I'm intending on reviewing your changes from this PR but alas, time is short.

depau:
No rush, it doesn't look like the next CalyxOS release is going to come any time soon so I'll have to make do with black lines on Garmin Connect anyway :)

Sure, I'll do that in another PR, it shouldn't too hard to integrate any layer changes with these changes (assuming the overall design doesn't change too much over here) 

Thank you!

johnassel:
Might it be possible that this patch fixes the black altitude history line with current versions of Flightradar24? 
Symptoms are similar to the behaviour of Garmin Connect: Normally there should be a colored line on the map indicating the flight altitude but actually it is completely black.

depau:
@johnassel it's very likely it's the same issue, so yes it should be fixed

depau:
Hey @fynngodau

I was looking forward to checking out the code again for giving a shot at the custom tiles implementation as that would fix my local public transport app, but since this PR does a lot of important changes to the layer handling I'd prefer to finish this up first before getting to work on it.

I was wondering if you could take a look when you have some time. I understand this isn't a super straightforward change and free time isn't unlimited, but I think reviewing it commit-by-commit should make it easier since I took care to tidy things up somewhat nicely.

depau:
FYI, as part of my work to implement the custom tiles layer I heavily refactored my previous implementation of the function that implements Z-index-based layering of annotation layers.

This should make the code significantly more straightforward to review, as well as improve the readability in general and remove an ugly unsafe cast.

I'm going to submit the custom tile changes as a separate pull request which will also include these changes. Once and if this PR is merged I can rebase the other one for proper review.

ale5000-git:
@depau
Could you please update the PR with the latest code from master?

depau:
@ale5000-git all done

depau:
Since I noticed tile overlays were added, I added z-index support to them as well

PR #3074, opened 2025-09-29
Vending: Add app installer proxy

DaVinci9196:
Function extension: Allows installation of installation packages downloaded from other channels to solve the problem that most applications that require the installation source to be GP cannot be used.

The channel application needs to call the vending page in the specified way for installation.

### DIFF
--- a/vending-app/src/main/AndroidManifest.xml
+++ b/vending-app/src/main/AndroidManifest.xml
@@ -300,7 +300,7 @@
         </service>
 
         <activity
-            android:name="org.microg.vending.installer.ChannelInstallActivity"
+            android:name="org.microg.vending.installer.AppInstallActivity"
             android:exported="true"
             android:excludeFromRecents="true"
             android:theme="@style/Theme.App.Translucent"
--- a/vending-app/src/main/res/values/strings.xml
+++ b/vending-app/src/main/res/values/strings.xml
@@ -60,11 +60,11 @@
     <string name="billing_permission_desc">Allows the user to purchase items through Google Play from within this app.</string>
     <string name="billing_permission_label">Google Play billing service</string>
 
-    <string name="channel_install_permission_allowed">Allow</string>
-    <string name="channel_install_needs_permissions_reminder_title">Necessary authorization reminder</string>
-    <string name="channel_install_allow_to_install_third_app">Allow \"microG Companion\" to install apps downloaded from \"%1$s\"</string>
-    <string name="channel_install_refuse_permission_remind">If you decline, the application installation will fail.</string>
-    <string name="channel_install_refuse_permission_remind_sub">You can modify the authorization status permissions at any time in \"microG Service Settings\".</string>
-    <string name="channel_install_not_remind_again">Keep this selection and don\'t ask again</string>
+    <string name="app_install_permission_allowed">Allow</string>
+    <string name="app_install_needs_permissions_title">App Installer</string>
+    <string name="app_install_allow_to_install_third_app">Allow \"microG Companion\" to install apps downloaded from \"%1$s\"</string>
+    <string name="app_install_refuse_permission_remind">If you decline, the application installation will fail.</string>
+    <string name="app_install_refuse_permission_remind_sub">You can modify the authorization status permissions at any time in microG Service Settings.</string>
+    <string name="app_install_not_remind_again">Keep this selection and don\'t ask again</string>
 
 </resources>
--- a/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsContract.kt
+++ b/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsContract.kt
@@ -278,7 +278,7 @@ object SettingsContract {
         const val ASSET_DELIVERY = "vending_asset_delivery"
         const val ASSET_DEVICE_SYNC = "vending_device_sync"
         const val APPS_INSTALL = "vending_apps_install"
-        const val APPS_INSTALL_CHANNEL_LIST = "vending_apps_install_channel_list"
+        const val APPS_INSTALLER_LIST = "vending_apps_installer_list"
 
         val PROJECTION = arrayOf(
             LICENSING,
@@ -288,7 +288,7 @@ object SettingsContract {
             ASSET_DELIVERY,
             ASSET_DEVICE_SYNC,
             APPS_INSTALL,
-            APPS_INSTALL_CHANNEL_LIST,
+            APPS_INSTALLER_LIST,
         )
     }
 
--- a/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
+++ b/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
@@ -10,7 +10,6 @@ import android.content.ContentValues
 import android.content.Context
 import android.content.Context.MODE_PRIVATE
 import android.content.SharedPreferences
-import android.content.pm.ApplicationInfo
 import android.database.Cursor
 import android.database.MatrixCursor
 import android.net.Uri
@@ -369,7 +368,7 @@ class SettingsProvider : ContentProvider() {
             Vending.ASSET_DEVICE_SYNC -> getSettingsBoolean(key, false)
             Vending.SPLIT_INSTALL -> getSettingsBoolean(key, false)
             Vending.APPS_INSTALL -> getSettingsBoolean(key, false)
-            Vending.APPS_INSTALL_CHANNEL_LIST -> getSettingsString(key, "")
+            Vending.APPS_INSTALLER_LIST -> getSettingsString(key, "")
             else -> throw IllegalArgumentException("Unknown key: $key")
         }
     }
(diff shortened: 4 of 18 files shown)


ale5000-git:
@DaVinci9196
Is it possible to also invoke this installation via adb? (without having another store app installed but just passing the path of a local apk file to microG Companion)

DaVinci9196:
> @DaVinci9196 Is it possible to also invoke this installation via adb? (without having another store app installed but just passing the path of a local apk file to microG Companion)

cannot

DaVinci9196:
> @DaVinci9196 Is it possible to also invoke this installation via adb? (without having another store app installed but just passing the path of a local apk file to microG Companion)
[demo.zip](https://github.com/user-attachments/files/22599520/demo.zip)
You can use this demo

mar-v-in:
@ale5000-git this is effectively the same as installing with `adb install -i com.android.vending [...]`, so it's not really needed to make this accessible from adb

ale5000-git:
> @ale5000-git this is effectively the same as installing with `adb install -i com.android.vending [...]`, so it's not really needed to make this accessible from adb

@mar-v-in
Via adb you can set installSource but not installInitiator.
If microG can "hide" where the command originated (shell) without using root, then we can avoid this issue.

mar-v-in:
I see. The problem is that on Android, activities only reliably learn the initiator of a request when it is invoked using `startActivityForResult()`, but the `adb shell am start` command uses a regular `startActivity()` call. But maybe we can use something else, e.g. a custom service call instead. Let's keep that out of the initial feature though.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.json.JSONObject
+
+enum class AllowType(val value: Int){
+    ALLOWED_NEVER(0),
+    ALLOWED_REQUEST(1),
+    ALLOWED_SINGLE(2),
+    ALLOWED_ALWAYS(3),
+}
+
+data class InstallChannelData(val packageName: String, var allowed: Int) {
```
mar-v-in: Please store the package signature with the app (and verify it when retrieving the entry from the list). If an app is uninstalled and replaced with another malicious app that maliciously has the same package name, the user should be prompted again.

ale5000-git:
Recently I had another similar idea (but I hadn't yet time to really look at it):
- Integrate the code of F-Droid Privileged Extension (almost unchanged) directly inside microG Companion => https://gitlab.com/fdroid/privileged-extension

It would bring two advantages:
1) F-Droid can use it directly with just a minor change (package name);
2) On new Android versions all app stores can install apps (with user confirmation), but on old Android no app stores can do it (except privileged system apps) without enable the option to install from unknown sources in Android developer settings; this thing would solve this problem.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    override fun toString(): String {
+        return JSONObject().put(CHANNEL_PACKAGE_NAME, packageName).put(CHANNEL_ALLOWED, allowed).toString()
+    }
+
+    companion object {
+        private const val CHANNEL_PACKAGE_NAME = "packageName"
+        private const val CHANNEL_ALLOWED = "allowed"
+        private val localChannelList = mutableSetOf<InstallChannelData>()
```
mar-v-in: Why a static variable here?

vending-app/src/main/AndroidManifest.xml
```diff
@@ -297,5 +299,24 @@
...
         </service>
 
+        <activity
+            android:name="org.microg.vending.installer.ChannelInstallActivity"
+            android:exported="true"
+            android:excludeFromRecents="true"
+            android:theme="@style/Theme.App.Translucent"
+            tools:targetApi="21">
+            <intent-filter>
+                <action android:name="com.android.vending.action.install"/>
```
mar-v-in: Typically, action names on Android are all uppercase. Also this action is specific to microG, so please adjust to `org.microg.vending.action.INSTALL_PACKAGE` or similar.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
DaVinci9196: Do you mean localChannelList?

mar-v-in: Yes. Fields in companion objects are static, meaning they leak memory and thus should be avoided.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
```diff
@@ -0,0 +1,56 @@
...
+data class InstallChannelData(val packageName: String, var allowed: Int, var pkgSignSha256:String) {
+
+    override fun toString(): String {
+        return JSONObject().put(CHANNEL_PACKAGE_NAME, packageName).put(CHANNEL_ALLOWED, allowed).toString()
+    }
+
+    companion object {
+        private const val CHANNEL_PACKAGE_NAME = "packageName"
+        private const val CHANNEL_ALLOWED = "allowed"
+        private const val CHANNEL_PACKAGE_SIGN = "allowed"
```
mar-v-in: Must use a different json field name or won't work.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
```diff
@@ -0,0 +1,56 @@
...
+    ALLOWED_NEVER(0),
+    ALLOWED_REQUEST(1),
+    ALLOWED_SINGLE(2),
+    ALLOWED_ALWAYS(3),
+}
+
+data class InstallChannelData(val packageName: String, var allowed: Int, var pkgSignSha256:String) {
+
+    override fun toString(): String {
+        return JSONObject().put(CHANNEL_PACKAGE_NAME, packageName).put(CHANNEL_ALLOWED, allowed).toString()
```
mar-v-in: Signature is missing from the object while serializing.

vending-app/src/main/res/values/strings.xml
```diff
@@ -60,4 +60,11 @@
     <string name="billing_permission_desc">Allows the user to purchase items through Google Play from within this app.</string>
     <string name="billing_permission_label">Google Play billing service</string>
 
+    <string name="channel_install_permission_allowed">Allow</string>
+    <string name="channel_install_needs_permissions_reminder_title">Necessary authorization reminder</string>
+    <string name="channel_install_allow_to_install_third_app">Allow \"microG Companion\" to install apps downloaded from \"%1$s\"</string>
+    <string name="channel_install_refuse_permission_remind">If you decline, the application installation will fail.</string>
+    <string name="channel_install_refuse_permission_remind_sub">You can modify the authorization status permissions at any time in \"Settings > Apps And Services > App Management > microG Service > microG Settings Page\".</string>
```
mar-v-in: The user interface to get the microG Settings is different between various Android versions. I think easiest would be to just write "microG Settings" here and have the user figure out themselves.

mar-v-in (changes requested):

ale5000-git:
@mar-v-in
What do you think of this? https://github.com/microg/GmsCore/pull/3074#issuecomment-3347626665

mar-v-in:
@ale5000-git the purpose of this is not to give vending / microG Companion privileged install permission for use by other apps, but really is only about triggering installation from the `com.android.vending` package so that installSource/Initiator is set when apps have checks for it (e.g. Android Auto). The idea would be that Aurora Store implements this API (as one of its many installer options) for much better user-experience without any modification needed on the ROM.

ale5000-git:
> @ale5000-git the purpose of this is not to give vending / microG Companion privileged install permission for use by other apps, but really is only about triggering installation from the `com.android.vending` package so that installSource/Initiator is set when apps have checks for it (e.g. Android Auto). The idea would be that Aurora Store implements this API (as one of its many installer options) for much better user-experience without any modification needed on the ROM.

It seems good, but I hope it is implemented also in F-Droid since one may also want to allow an app installed by F-Droid to appears in Android Auto.

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
DaVinci9196: Processed

play-services-base/core/src/main/kotlin/org/microg/gms/vending/InstallChannelData.kt
DaVinci9196: Processed

vending-app/src/main/res/values/strings.xml
DaVinci9196: Processed

david7919:
In the new issue for this api on auroras gitlab there seems to be some confusion about this feature.
Does this require root?
As far as Ive read in here it doesnt, but Im not sure either.
https://gitlab.com/AuroraOSS/AuroraStore/-/issues?show=eyJpaWQiOiIxNDI2IiwiZnVsbF9wYXRoIjoiQXVyb3JhT1NTL0F1cm9yYVN0b3JlIiwiaWQiOjE3NTQ3NjY0NH0%3D

ale5000-git:
Related ticket: https://gitlab.com/AuroraOSS/AuroraStore/-/issues/1426

PR #3084, opened 2025-10-06
Fix v2 licensing

fynngodau:
According to latest `com.pairip.licensecheck` behavior, the third parameter in the licensing v2 call is actually an integer and not a Bundle. I've tested this with current apps including v2 licensing, namely [`org.howwefeel.moodmeter`](https://play.google.com/store/apps/details?id=org.howwefeel.moodmeter) and [`fr.maif.digital.maifetmoi`](https://play.google.com/store/apps/details?id=fr.maif.digital.maifetmoi) (note: closes on emulator, but console log from the app indicate that license check is successful).

Regardless, I've verified against https://github.com/e-foundation/lvl-sample that it still works if an empty bundle is passed.

Related: https://gitlab.e.foundation/e/os/backlog/-/issues/3730

Special thanks to @prplwtf.

### DIFF
--- a/vending-app/src/main/aidl/com/android/vending/licensing/ILicensingService.aidl
+++ b/vending-app/src/main/aidl/com/android/vending/licensing/ILicensingService.aidl
@@ -11,6 +11,6 @@ import com.android.vending.licensing.ILicenseV2ResultListener;
 
 interface ILicensingService {
     oneway void checkLicense(long nonce, String packageName, ILicenseResultListener listener);
-    oneway void checkLicenseV2(String packageName, ILicenseV2ResultListener listener, in Bundle extraParams);
+    oneway void checkLicenseV2(String packageName, ILicenseV2ResultListener listener, int unknown);
 
 }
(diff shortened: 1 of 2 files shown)


mar-v-in:
What is the value of the integer? Does it happen to always be `0`?

A Bundle is represented as a typed parcelable object on the parcel, which is [int]0 for null or [int]1 + writeToParcel for non-null values. As such, a fixed value of 0 can be used as a shortcut to represent an always null Bundle.

This also means, that a value of [int]0 should be fully working for the method signature that accepts a Bundle.

So can you please verify the value can be non-zero int?

fynngodau:
Thanks for the very helpful input @mar-v-in. Indeed my conclusion was nonsensical. The real problem was that we forced the `extraParams` bundle to be non-null in Kotlin, but it was indeed `null` (so int = 0 in the parcel). (In my test case with https://github.com/e-foundation/lvl-sample, we pass a non-null empty `Bundle`, which I never observed in practice…)

fynngodau:
Some more context on how this bug occurred for those curious or seeking to learn from it:

- I originally wrote the licensing service in Java, and tested it with real apps while cobbling together https://github.com/e-foundation/lvl-sample to ease testing (more feedback, easier to repeat)
- then I migrated to Kotlin, and apparently didn't test with real apps since the test case was passing
- …so ever since then, v2 licensing wasn't actually working
- …but I don't think anybody was affected, and nobody noticed? Since we know that v2 licensing doesn't cause any problem on the client side if you (as microG) receive it without error, but never answer it.

PR #3085, opened 2025-10-07
Fix system  profile behavior

mkolegan:
System profile was not read from xml because the logic fallback to realData getProfileResId() was returning 0 in that case.

In addition to PROFILE_USER and built-in resource-based profiles  we need to check PROFILE_SYSTEM


ale5000-git:
@mar-v-in
I wonder, could we put these checks directly inside getProfileResId, so the caller doesn't have to care?
To avoid future problems.

mar-v-in:
Thanks @mkolegan

@ale5000-git `getProfileResId` really only does that (getting the ressource id of the profile if it is a built-in profile). One could probably also check for `hasProfile()` here instead, but for now this solution is safe and works.

PR #3132, opened 2025-11-07
Google Maps Timeline: Added dummy method

DaVinci9196:
Resolving the issue of Google Maps prompting that Google Play Services needs to be updated.


mar-v-in:
This is included in 2eb77bcad73823b07967034a2610d0e904aa310b, so I'm closing here.

PR #3147, opened 2025-11-21
HmsMaps: Fixed map white screen issue

DaVinci9196:
In some scenarios, if onSaveInstanceState is executed but no action is taken, the map will not display when the user returns to the previous location.
e.g. Imo's Pizza <com.imospizza.app>


DaVinci9196:
This issue has been fixed through other means; Closed.

PR #3179, opened 2025-12-11
Add Thai translations for microG-UI-tools component

grenadin:
This pull request adds Thai (th) translations for microg-ui-tools,
by creating res/values-th/strings.xml based on the default strings.xml.


mar-v-in:
Thanks.

PR #3192, opened 2025-12-17
Avoiding ANR that may occur during login

DaVinci9196:
Some users have reported LoginActivity ANR, which is suspected to be related to this issue.


mar-v-in:
`Task.await()` is a non-blocking kotlin coroutine. It shouldn't be able to trigger ANRs. Can you provide more details on the issue you have seen in LoginActivity ANR?

mar-v-in:
Can you provide a way to reproduce this issue, so we can confirm this PR fixes it?

DaVinci9196:
> Can you provide a way to reproduce this issue, so we can confirm this PR fixes it?

No reproducible path has been found, so it's unclear whether it has been fixed. However, multiple tests based on this PR did not detect ANR.

ale5000-git:
I haven't verified exactly but 5 seconds may be too low for old devices.

PR #3193, opened 2025-12-17
Fix issue with location settings dialog

adiktofsugar:
On Android 12 (on a mudita Kompact), in "Location" section, clicking the hamburger menu and "set location" fails because apparently the Dialog looks for ids that have "button" in them and blows up if it's not actually a button. Changing the id to a custom id fixes the issue.


adiktofsugar:
@mar-v-in I'd merge but I can't find the button. I suspect only maintainers can merge it, but let me know if I'm just missing something obvious or need to read a contributing doc or something.

mar-v-in:
Thanks!

PR #3194, opened 2025-12-17
FIDO: Add Hybrid connection

DaVinci9196:
FIDO adds cross-device (hybrid) login.
Scene 1: Log in to local applications using other device keys by scanning the local device's QR code.
Scene 2: Scan the key QR code on another device using the Google App to log in using your device's key.


ale5000-git:
@DaVinci9196
I'll throw out an idea for a possible extension if you have the time and inclination:
`Scene 3: Scan the key QR code on another device from microG settings (withOUT using the Google App) to log in using your device's key.`

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/QRBounceActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+package org.microg.gms.fido.core.ui
+
+import android.content.Intent
+import android.os.Bundle
+import android.util.Log
+import androidx.appcompat.app.AppCompatActivity
+import kotlin.also
+import kotlin.text.lowercase
+
+class QRBounceActivity : AppCompatActivity() {
```
mar-v-in: What's the purpose of this activity? Why not just directly launch HybridAuthenticateActivity when it's needed?

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/QRBounceActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+        var scheme: String? = null
+        val action = intent!!.action
+        val data = intent.data
+        if (action == null || (action != "android.intent.action.VIEW") || data == null || (data.scheme.also { scheme = it }) == null || (scheme!!.lowercase() != "fido")) {
+            Log.w(TAG, "Invalid data from scanning QR Code: $data")
+            finish()
+            return
+        }
+        val targetIntent = Intent()
+        targetIntent.setClassName(this, "org.microg.gms.fido.core.ui.hybrid.HybridAuthenticateActivity")
```
mar-v-in: Please reference directly the class (via `HybridAuthenticateActivity::class.java`) rather than by name.

play-services-core/src/main/kotlin/org/microg/gms/mlkit/BarcodeScanningActivity.kt
mar-v-in, line 1: You shouldn't do any changes to this file, this is from mlkit and is entirely unrelated to fido.

play-services-core/src/main/kotlin/org/microg/gms/ui/AccountsFragment.kt
```diff
@@ -140,6 +141,7 @@ class AccountsFragment : PreferenceFragmentCompat() {
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         menu.add(0, MENU_GAMES_MANAGED, 0, org.microg.gms.base.core.R.string.menu_game_managed)
+        menu.add(0, MENU_SCAN_FIDO_QR_CODE, 0, org.microg.gms.base.core.R.string.menu_qr_code_scan)
```
mar-v-in: I don't think Google has such a menu option either. The documentation also suggests that the QR code is scanned using the regular camera app, not using Google settings app. Please make it work using the regular flow, no need to have a custom QR code reader button in microG.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
```diff
@@ -10,13 +10,40 @@ import com.upokecenter.cbor.CBORObject
...
 import org.microg.gms.fido.core.protocol.decodeAsPublicKeyCredentialParameters
+import org.microg.gms.fido.core.protocol.encodeAsCbor
 import org.microg.gms.utils.ToStringHelper
 
 class AuthenticatorGetInfoCommand : Ctap2Command<AuthenticatorGetInfoRequest, AuthenticatorGetInfoResponse>(AuthenticatorGetInfoRequest()) {
     override fun decodeResponse(obj: CBORObject) = AuthenticatorGetInfoResponse.decodeFromCbor(obj)
 }
 
-class AuthenticatorGetInfoRequest : Ctap2Request(0x04)
+class AuthenticatorGetInfoRequest(
```
mar-v-in: `authenticatorGetInfo` does not take any inputs, see https://fidoalliance.org/specs/fido-v2.3-rd-20251023/fido-client-to-authenticator-protocol-v2.3-rd-20251023.html#authenticatorGetInfo
The structure here looks like you actually wanted to use the outputs, that is the `AuthenticatorGetInfoResponse` class below.

play-services-core/src/main/kotlin/org/microg/gms/ui/AccountsFragment.kt
ale5000-git: Not all camera apps support QR codes (specially on old Android), so it would be nice to have the possibility to make it working without having to install an additional app.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/QRBounceActivity.kt
DaVinci9196: The initial idea was to have a transition page, with HybridAuthenticateActivity not directly accessible externally. It can also be removed.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
DaVinci9196: It has been changed to use AuthenticatorGetInfoResponse

play-services-core/src/main/kotlin/org/microg/gms/mlkit/BarcodeScanningActivity.kt
DaVinci9196: Reverted

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/QRBounceActivity.kt
DaVinci9196: Removed

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/QRBounceActivity.kt
DaVinci9196: Removed

play-services-core/src/main/kotlin/org/microg/gms/ui/AccountsFragment.kt
mar-v-in: @ale5000-git using FIDO hybrid passkeys is unrelated to Google accounts and this shouldn't be buried in some Google account-related setting.

People with a camera app that can't scan QR codes are well-adviced to have another way to scan QR codes anyway, or how would one be able to scan QR codes that are not application-specific, like just regular web links?

ale5000-git: @mar-v-in
My opinion is that a not application-specific QR code can be useful but not fundamental so on old phones it is perfectly possible to not have generic QR code app but still it would be useful to still use FIDO hybrid passkeys.

If the code isn't too big I think it is very useful to have it.

As for "shouldn't be buried in some Google account-related setting", it could be moved to a better place (but without requiring a separate app if possible).

mar-v-in: microG isn't a user-facing app. The only launcher is microG settings, which intentionally can be disabled/hidden for users that prefer to reach it from system settings or when custom ROMs have a tight integration. If we wanted to have a new user-facing launcher, what else would be in it other than this QR scanner? Having an additional microG launcher with just a QR code scanner wouldn't really make sense, users could just install a QR code scanner app then.

ale5000-git: @mar-v-in
Even if the microG settings aren't in the launcher, they still are integrated in Android settings so it is fine (in my opinion).
My idea is to have a special "FIDO" section inside microG settings that open a separate page that contains the QR code scanning function (and maybe in the future the possibility to list and delete the FIDO screen lock credentials).

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
```diff
@@ -10,60 +10,117 @@ import com.upokecenter.cbor.CBORObject
...
-    val minPINLength: Int?,
-    val firmwareVersion: Int?,
-    val maxCredBlobLength: Int?,
-    val maxRPIDsForSetMinPINLength: Int?,
-    val preferredPlatformUvAttempts: Int?,
-    val uvModality: Int?,
-    val certifications: Map<String, Int>?,
-    val remainingDiscoverableCredentials: Int?,
-    val vendorPrototypeConfigCommands: List<Int>?,
+    val versions: List<String> = arrayListOf("FIDO_2_0", "FIDO_2_1"),
```
mar-v-in: Please don't default this to values that are not "default" according to the specification, just make the mandatory to be included in the invocation, as they are required.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
```diff
@@ -10,60 +10,117 @@ import com.upokecenter.cbor.CBORObject
...
+    val extensions: List<String>? = null,
+    val aaguid: ByteArray = ByteArray(16),
```
mar-v-in: Here as well, it's mandatory to provide one, so don't put a broken default.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
```diff
@@ -10,60 +10,117 @@ import com.upokecenter.cbor.CBORObject
...
+        preferredPlatformUvAttempts?.let { set(0x11, CBORObject.FromObject(it)) }
+        uvModality?.let { set(0x12, CBORObject.FromObject(it)) }
+        certifications?.let { map ->
+            CBORObject.NewMap().apply {
+                map.forEach { (key, value) ->
+                    set(CBORObject.FromObject(key), CBORObject.FromObject(value))
+                }
+            }
+        }?.let { set(0x13, it) }
+        remainingDiscoverableCredentials?.let { set(0x14, CBORObject.FromObject(it)) }
```
mar-v-in: Use `.encodeAsCbor()` instead of `CBORObject.FromObject`.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorMakeCredential.kt
```diff
@@ -31,7 +35,7 @@ class AuthenticatorMakeCredentialRequest(
     val options: Options? = null,
     val pinAuth: ByteArray? = null,
     val pinProtocol: Int? = null
-) : Ctap2Request(0x01, CBORObject.NewMap().apply {
+) : Ctap2Request(COMMAND, CBORObject.NewMap().apply {
```
mar-v-in: Please keep the command byte directly in the Ctap2Request constructor invocation, to keep it easy to read and find.

mar-v-in (changes requested):

play-services-fido/src/main/java/com/google/android/gms/fido/fido2/api/common/PublicKeyCredentialDescriptor.java
```diff
@@ -41,6 +41,11 @@ public class PublicKeyCredentialDescriptor extends AbstractSafeParcelable {
     private PublicKeyCredentialDescriptor() {
     }
 
+    public PublicKeyCredentialDescriptor(@NonNull PublicKeyCredentialType type, @NonNull byte[] id) {
```
mar-v-in: There's no constructor like this upstream, see https://developers.google.com/android/reference/com/google/android/gms/fido/fido2/api/common/PublicKeyCredentialDescriptor#public-constructor-summary
If you really want to set transports to `null`, just use the constructor below and set it explicitly.

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
DaVinci9196: changed

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
DaVinci9196: changed

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetInfo.kt
DaVinci9196: changed

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorMakeCredential.kt
DaVinci9196: changed

play-services-fido/src/main/java/com/google/android/gms/fido/fido2/api/common/PublicKeyCredentialDescriptor.java
DaVinci9196: changed

play-services-core/src/main/kotlin/org/microg/gms/ui/AccountsFragment.kt
mar-v-in: Marking this as resolved. The decision to potentially do a FIDO section in microG settings shouldn't be blocking for this PR.

PR #3202, opened 2025-12-23
TapPay: Fixed crashed due to empty parameters

DaVinci9196:
Fix app crashing when opened.
e.g. one – control over your cards<ch.viseca.visecaone>

### DIFF
--- a/play-services-core/src/main/kotlin/com/google/android/gms/locationsharingreporter/service/LocationSharingReporterApiService.kt
+++ b/play-services-core/src/main/kotlin/com/google/android/gms/locationsharingreporter/service/LocationSharingReporterApiService.kt
@@ -93,23 +93,23 @@ class LocationSharingReporterApiServiceImpl(
 ) : ILocationSharingReporterService.Stub(), LifecycleOwner {
     private val activePermissionRequestLock = Mutex()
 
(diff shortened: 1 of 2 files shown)


ale5000-git:
@mar-v-in
Since ApiMetadata can be null, shouldn't be a check before using it?
https://github.com/microg/GmsCore/blob/dbd68770e24be6cd074e73a9bd2117a386753ada/play-services-core/src/main/kotlin/com/google/android/gms/locationsharingreporter/service/LocationSharingReporterApiService.kt#L267

There are also other cases near it.

mar-v-in:
> @mar-v-in Since ApiMetadata can be null, shouldn't be a check before using it?
> 
> https://github.com/microg/GmsCore/blob/dbd68770e24be6cd074e73a9bd2117a386753ada/play-services-core/src/main/kotlin/com/google/android/gms/locationsharingreporter/service/LocationSharingReporterApiService.kt#L267
> 
> There are also other cases near it.

That's a reference to the class / a static field in it, which is never null. What might be null are instances of ApiMetadata. We don't use ApiMetadata, we just have to check that function signatures in Kotlin represent the nullability correctly, which is what this PR does.

Issue #3008, opened 2025-08-10
"microG Services keeps stopping" dialog error whenever Location Services enabled (java.lang.NullPointerException)


therealcodlord:
**Describe the bug**
I am getting repeated dialog errors "microG Services keeps stopping" whenever I enable Location Services on my phone.

**To Reproduce**
I enable flight mode with Location services Disabled. I kill all running apps and get no error. 
I Enable Location Services: error starts within a few minutes and repeats approx every 4 minutes.

**Expected behavior**
To be able to use location services without repeated dialog errors.

**Screenshots**
N/A but the errors show like this in the logs:

```
08-10 09:32:00.336  9614  9642 E AndroidRuntime: FATAL EXCEPTION: NetworkLocationService
08-10 09:32:00.336  9614  9642 E AndroidRuntime: Process: com.google.android.gms, PID: 9614
08-10 09:32:00.336  9614  9642 E AndroidRuntime: java.lang.NullPointerException
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at org.microg.gms.location.network.wifi.ExtensionsKt.toWifiDetails(SourceFile:2)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at org.microg.gms.location.network.NetworkLocationService.scan(Unknown Source:355)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at org.microg.gms.location.network.NetworkLocationService.highPowerScanRunnable$lambda$0(Unknown Source:6)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at org.microg.gms.location.network.NetworkLocationService.h(SourceFile:1)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at org.microg.gms.location.network.e.run(SourceFile:1)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at android.os.Handler.handleCallback(Handler.java:942)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:99)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at android.os.Looper.loopOnce(Looper.java:201)
08-10 09:32:00.336  9614  9642 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:288)
[...]

Issue #2726, opened 2025-01-01
Location (network) not working


jstam84:
**Describe the bug**
Microg version 0.3.6.244753. Clean lineage 21. All self check options pass. Location WiFi and Bluetooth enabled. In microg all settings for location enabled. Provider is beacondb (tried position xyz too). Some apps request location sometime via microg (recent location requests) bit most don't show up at all. In android location settings they do. Network location doesn't seem to work properly then?

aiman-al-masoud:
Hello! Similar problem here, on the same version (0.3.6.244735). Microg was granted location permissions, location settings in microg are al active, tried both default location services, I'm getting these logs from the app that I'm trying to make work (via adb logcat):

```
01-04 18:29:17.513  7765  7765 D LocationApisInterface: fast location: null
01-04 18:29:17.517  7765  7864 W GoogleApiManager: The service for com.google.android.gms.internal.location.zzdz is not available: ConnectionResult{statusCode=SERVICE_INVALID, resolution=null, message=null}
```

PS: tried an older version of microg (the previous) but still not working.

n8name:
Same here. Worked previously with lineage for microg. For lineage 22 I switched (dirty flash) to official lineage. Since than network location seems to be broken.
Already reinstalled betas via fdroid (0.3.6.40226).

The-First-King:
https://github.com/microg/GmsCore/issues/2635#issuecomment-2587462850

Did you try this and reboot?

mherrmann3:
I observed a similar issue: several apps (like Osmand, SatStat, )don't receive a network location from _microG_ (and don't appear in _microG_'s Location > _Recent access_), albeit showing up in Android's Settings > Location > _Recent access_.
**However**, for some apps, network-based location works (_Google Camera_, _WhatsApp_, _Google Maps_), and they also appear in _microG_'s _Recent access_.
For _MagicEarth_, network-based lcoation seems to work, despite not showing up in _microG_'s _Recent access_ (but in Android's Settings)

Could it be that different apps request the network location via different location providers?  and their APIs? Something



Not sure if helpful, but her some `dumpsys` extracts:

1. For the above mention apps, I get the following from `dumpsys location` (removed details):
<details style="margin-left:10px">
<summary>LOG (click to expand)</summary>

```
  Historical Aggregate Location Provider Data:
    passive:
      10321/com.generalmagic.magicearth: ...
      10486/net.osmand.plus: ...
      1000/android[SensorNotificationService]: ...
[...]

PR #3210, opened 2025-12-30
Fix issues with geolocation

althafvly:
Fixes: https://github.com/microg/GmsCore/issues/3008
Fixes beacondb on wifi: https://github.com/microg/GmsCore/issues/2726

### DIFF
--- a/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/ichnaea/IchnaeaServiceClient.kt
+++ b/play-services-location/core/provider/src/main/kotlin/org/microg/gms/location/network/ichnaea/IchnaeaServiceClient.kt
@@ -6,6 +6,7 @@
 package org.microg.gms.location.network.ichnaea
 
 import android.content.Context
+import android.content.pm.PackageManager
 import android.location.Location
 import android.net.Uri
 import android.os.Bundle
@@ -41,6 +42,10 @@ class IchnaeaServiceClient(private val context: Context) {
     private val cache = LruCache<String, Location>(REQUEST_CACHE_SIZE)
     private val start = SystemClock.elapsedRealtime()
 
+    private val hasTelephony by lazy {
+        context.packageManager.hasSystemFeature(PackageManager.FEATURE_TELEPHONY)
+    }
+
     private fun GeolocateRequest.hash(): ByteArray? {
         if (cellTowers.isNullOrEmpty() && (wifiAccessPoints?.size ?: 0) < 3 || bluetoothBeacons?.isNotEmpty() == true) return null
         val minAge = min(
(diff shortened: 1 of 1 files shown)


althafvly:
I'm not sure if this is right way to fix this https://github.com/microg/GmsCore/issues/2726.

But without that i was getting this response from beacondb

```
{
    "error": {
        "code": 404,
        "errors": [
            {
                "domain": "geolocation",
                "message": "No location could be estimated based on the data provided",
                "reason": "notFound"
            }
        ],
        "message": "Not found"
    }
}
```

mar-v-in:
IP address fallback has a lot of downsides and usually shouldn't be used on phones except when the current connection is via a private Wi-Fi and not using a VPN:
- Public WiFi networks often route traffic through some service that makes IP address unreliable for location
- VPNs hide your IP address and would locate you completely wrong
- Mobile networks use large CGNATs or dynamic IP address reassigning within their network, meaning the IP address effectively could be anywhere within their network reach. Man geoip services thus will just tell you the country for mobile network IP addresses.

Technically you could adjust the code to allow IP address based fallback when connected to an encrypted WiFi network (which is a reasonable proxy for what we're looking for), but you also need to make sure to turn it off for cell tower location requests, as those are cached but the WiFi based IP address might be wrong for the cell tower.

However, I don't think that should be necessary: IP based geolocation with open data is usually not more accurate than cell tower based location (big tech has access to certain high accuracy geoip sources, but those are typically not exposed via API for privacy reasons). We do have much better cell tower location data in open databases, so those should typically resolve. You would still see those errors from beaconDB when trying to resolve WiFi location, but they shouldn't cause any issues.

althafvly:
> IP address fallback has a lot of downsides and usually shouldn't be used on phones except when the current connection is via a private Wi-Fi and not using a VPN:
> 
> * Public WiFi networks often route traffic through some service that makes IP address unreliable for location
> * VPNs hide your IP address and would locate you completely wrong
> * Mobile networks use large CGNATs or dynamic IP address reassigning within their network, meaning the IP address effectively could be anywhere within their network reach. Man geoip services thus will just tell you the country for mobile network IP addresses.
> 
> Technically you could adjust the code to allow IP address based fallback when connected to an encrypted WiFi network (which is a reasonable proxy for what we're looking for), but you also need to make sure to turn it off for cell tower location requests, as those are cached but the WiFi based IP address might be wrong for the cell tower.
> 
> However, I don't think that should be necessary: IP based geolocation with open data is usually not more accurate than cell tower based location (big tech has access to certain high accuracy geoip sources, but those are typically not exposed via API for privacy reasons). We do have much better cell tower location data in open databases, so those should typically resolve. You would still see those errors from beaconDB when trying to resolve WiFi location, but they shouldn't cause any issues.

It's mainly an issue for the wifi tablets and devices with broken GPS. It works fine on mobile data. As you mentioned wifi location isn't that accurate. Kms away atleast for me.

mar-v-in:
Oh, I see. Maybe then enabling `considerIp` for devices that don't have a mobile network interface at all is a sensible approach.

althafvly:
> Oh, I see. Maybe then enabling `considerIp` for devices that don't have a mobile network interface at all is a sensible approach.

Updated the PR.

mar-v-in:
Merged this manually with my local changes, thanks!

PR #3213, opened 2025-12-31
Implement CommonService

DaVinci9196:
Fix Planner 5D not being able to use Google Sign-in.

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/common/GmsCommonService.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/common/GmsCommonService.kt
@@ -15,11 +15,8 @@ import com.google.android.gms.common.internal.GetServiceRequest
 import com.google.android.gms.common.internal.IGmsCallbacks
 import com.google.android.gms.common.internal.service.ICommonCallbacks
 import com.google.android.gms.common.internal.service.ICommonService
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
 import org.microg.gms.BaseService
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

play-services-core/src/main/kotlin/org/microg/gms/common/GmsCommonService.kt
```diff
@@ -0,0 +1,50 @@
...
+class GmsCommonServiceImpl(val context: Context, val packageName: String, override val lifecycle: Lifecycle) : ICommonService.Stub(), LifecycleOwner {
+    override fun clearDefaultAccount(callbacks: ICommonCallbacks?) {
+        Log.d(TAG, "clearDefaultAccount: packageName: $packageName")
+        lifecycleScope.launchWhenStarted {
+            withContext(Dispatchers.IO) {
+                val authOptions = SignInConfigurationService.getDefaultOptions(context, packageName)
+                val authAccount = SignInConfigurationService.getDefaultAccount(context, packageName)
+                if (authOptions != null && authAccount != null) {
+                    Log.d(TAG, "$packageName:clear authAccount: ${authAccount.name} authOption:($authOptions)")
+                    performSignOut(context, packageName, authOptions, authAccount)
```
mar-v-in: The official implementation seems to not do any network interaction when `clearDefaultAccount` is invoked, it really just clears the default.

DaVinci9196: changed

PR #3230, opened 2026-01-16
FMD: Google Hub's Find My Device feature

DaVinci9196:
1. Added an empty service framework for FMD.
2. Added the ability to respond to Google service push notifications for Find My Device; this can be controlled via configuration options.
3. Fixed some display issues and unresponsive clicks within the Google Hub app.

Note: 
Based on this PR's operating environment, you need to check if the device is currently listed in the Google Hub app.
It usually appears in the device list 24 hours after registration. During this period, microG must not have any changes to its registered device information; otherwise, the device will not be found.


eepymeowers:
isn't the point of microG to get away from Google's tracking?

ale5000-git:
> isn't the point of microG to get away from Google's tracking?

Different users have different needs.
Since there is an option to enable/disable it, there won't be any problem.

Nuyoah-Yc:
@mar-v-in I verified this PR and found that there is a problem. Every time I log in to the account again, one more device appears in the list of searched devices. Through analysis, I found that when registering with Checkin, org.microg.gms.common.DeviceIdentifier#meid generates a random value, not the imei of the device. Once you log in to the account, a new device will be created.
image

play-services-core/src/main/AndroidManifest.xml
```diff
@@ -1181,6 +1194,56 @@
...
+            </intent-filter>
+        </service>
+
+        <service
+            android:name="org.microg.gms.findmydevice.spot.locationreporting.LocationReportService"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="com.google.android.gms.findmydevice.spot.service.locationreport.START" />
+            </intent-filter>
+        </service>
```
mar-v-in, line 1217: These services don't seem to be showing up in Google Play Services manifest - at least not in versions I checked.

mar-v-in: > Added an empty service framework for FMD.

Can you refer to the app that makes use of those services?

mar-v-in: Oh, I see now, those are only requested as Chimera service

Issue #3113, opened 2025-10-22
microG on Maven is stuck at 0.3.6.244735


ale5000-git:
@mar-v-in

microG on Maven is stuck at 0.3.6.244735

See:
https://mvnrepository.com/artifact/org.microg.gms
https://central.sonatype.com/namespace/org.microg.gms

Why new versions aren't published?

Sapiosenses:
FYI, APKmirror has been hosting microG files for a while, I don't know if one would want to rely on it as "official" unless there were some formal relationship in place but that has always been the most reliable 3rd-party website for manually grabbing android SW hosted on Gplay and elsewhere, in my experience.
 
Their files are up to date, perhaps Maven was only pulling from microg.org which could explain the lack of updates.

The two files I did a spotcheck on last month were binary identical to the official files, but they change the filenames to something easier to understand. (microG APK filenames have been very confusing lately IMHO) 


https://www.apkmirror.com/apk/microg-team/

ale5000-git:
@Sapiosenses
Maven is for the sources, to compile microG libraries inside other apps, not for microG itself.
If I'm not wrong it must be pushed by Marvin.

Sapiosenses:
> [@Sapiosenses](https://github.com/Sapiosenses) Maven is for the sources, to compile microG libraries inside other apps, not for microG itself. If I'm not wrong it must be pushed by Marvin.

I see. I spent some time trying to figure out what that site's mission was but there is almost zero information on it, the "about" page is basically blank except for a link to the operator's Twitter account.

Thanks

ale5000-git:
You can get some info here: https://en.wikipedia.org/wiki/Apache_Maven

ale5000-git:
Now they are updated; so I'm closing this.

ale5000-git:
@mar-v-in 
0.3.14 was published but 0.3.15 is missing.

mar-v-in:
0.3.15 is equivalent to 0.3.14 except for a small change in vending package that is not uploaded to Maven.

ale5000-git:
Ok, thanks. I just thought it might have broken again.

Issue #2922, opened 2025-05-26
0.3.7.250932 is not in google/maven android repos yet?


pastk:
I'm using the microG location lib and wanted to upgrade from 0.3.6, however:

```
Could not find org.microg.gms:play-services-location:0.3.7.250932
```

My repositories config in build.gradle is standard AFAIK
```
  repositories {
    google()
    mavenCentral()
  }
```

Issue #1729, opened 2022-05-31
org.microg.gms:play-services-location not available on maven


7h30n3:
Hey there,

Is there a particular reason why `org.microg.gms:play-services-location` isn't released on [maven](https://search.maven.org/search?q=g:org.microg.gms)?

We would like to use it as an alternative to GMS in our project.

PR #3235, opened 2026-01-18
Jitpack

p1gp1g:
MicroG libs [aren't updated on maven central for a moment now](https://central.sonatype.com/artifact/org.microg.gms/play-services-fido-core), this patch allow to use Jitpack to get them: https://jitpack.io/#p1gp1g/GmsCore 

Jitpack autoupdate the libraries, as soon as they continue to build

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,8 @@
+jdk:
+  - openjdk17
+install:
+  #Default - ./gradlew clean -Pgroup=com.github.microg -Pversion=branch-commit-1 -xtest -xlint assemble publishToMavenLocal
+  - sed -i '/withJavadocJar/d' gradle/publish-android.gradle
+  - ./gradlew clean -Pgroup=$GROUP -Pversion=$VERSION -xtest -xlint publishToMavenLocal
+env:
+  GRADLE_MICROG_VERSION_WITHOUT_GIT: 1


p1gp1g:
Related: #3113 #2922 #1729

hb0nd (approved):
lgtm, would be much appreciated by  CoMaps (we use play-services-location)

jitpack.yml
```diff
@@ -0,0 +1,8 @@
+jdk:
+  - openjdk17
+install:
+  #Default - ./gradlew clean -Pgroup=com.github.p1gp1g -Pversion=s1mpatch2-test_jitpack-c2e1c6ddbe-1 -xtest -xlint assemble publishToMavenLocal
+  - sed -i '/withJavadocJar/d' gradle/publish-android.gradle
```
mar-v-in, line 5: Why do we need to disable javadocs?

mar-v-in:
With ef110b03293266ebde4d890faed12c2221b2bf35 there is no need anymore to build `play-services-core-proto` as a library (which it was never meant to be) as `play-services-base/core` doesn't depend on it anymore.

jitpack.yml
p1gp1g: IIRC it failed because of a timeout with javadocs

p1gp1g:
Great, so the jitpack.yml alone should be enough ?

mar-v-in:
> Great, so the jitpack.yml alone should be enough ?

Exactly :)

play-services-core-proto/build.gradle
```diff
@@ -4,8 +4,44 @@
...
+    }
+
+    kotlinOptions {
+        jvmTarget = 1.8
+    }
+
+    lintOptions {
+        disable 'MissingTranslation'
+    }
+}
```
p1gp1g:
```suggestion
apply plugin: 'kotlin'
```

play-services-core-proto/build.gradle
```diff
@@ -21,15 +57,6 @@ sourceSets {
...
 
-compileJava {
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
-}
-
-compileKotlin {
-    kotlinOptions.jvmTarget = 1.8
-}
+apply from: '../gradle/publish-android.gradle'
```
p1gp1g:
```suggestion
compileJava {
    sourceCompatibility = 1.8
    targetCompatibility = 1.8
}

compileKotlin {
    kotlinOptions.jvmTarget = 1.8
}
```

play-services-core-proto/build.gradle
```diff
@@ -21,15 +57,6 @@ sourceSets {
...
 
-compileTestKotlin {
-    kotlinOptions.jvmTarget = 1.8
-}
+description = 'microG service implementation for play-services-core-proto'
```
p1gp1g:
```suggestion
compileTestKotlin {
    kotlinOptions.jvmTarget = 1.8
}
```

p1gp1g (commented):
I try to patch from github, I've pending changes on my local directory

jitpack.yml
p1gp1g: I tried to enable javadocs, but I had to remove it again:
bdcf19361058ad54de6db850bdc427293d5b5412
27177e9bf56d144cb3e4662bcb6baf11d6acfb96

mar-v-in: Thanks for testing, I guess it's fine if we go without it for the jitpack, just wanted to make sure.

p1gp1g: The 2 commits are 2 months old, I just saw that I tried to build with javadoc

I'm starting a build on jitpack with the javadoc, from this commit: https://jitpack.io/com/github/p1gp1g/GmsCore/bdcf193610/build.log

p1gp1g:
I've cleanup the branch and rebased

mar-v-in:
Thanks :)

PR #3237, opened 2026-01-18
Fido2: Fix instant use of screen lock

p1gp1g:
I had 2 problems with the current implementation:

1. as soon as a FIDO key was registered on the device (screen lock), then I couldn't use hardware key for the website, even if the allowCredentials contained a single key with USB only

This is because of `localSavedUserKey` that makes the authentication use instantly the screen lock if we have a key saved for the domain.

2. I can't selection the user with screenLock after the transport selection

This PR fixes these 2 issues. And does a little cleanup, and a few comments to help understand what does what

MEMO: I use fido-core as a library.

### DIFF
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/transport/screenlock/ScreenLockTransportHandler.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/transport/screenlock/ScreenLockTransportHandler.kt
@@ -259,13 +259,18 @@ class ScreenLockTransportHandler(private val activity: FragmentActivity, callbac
 
     override fun shouldBeUsedInstantly(options: RequestOptions): Boolean {
         if (options.type != RequestOptionsType.SIGN) return false
-        for (descriptor in options.signOptions.allowList.orEmpty()) {
+        // If there is an allow list, it means we try to login for a defined user.
+        // So if we have one of the allowed keys to login with the screen lock, we use it.
+        //
+        // We DON'T want to use instant connection if the allowList is empty, as the user
+        // may want to login with different user accounts, maybe on a hardware key
+        options.signOptions.allowList?.forEach {
             try {
-                val (type, data) = CredentialId.decodeTypeAndData(descriptor.id)
+                val (type, data) = CredentialId.decodeTypeAndData(it.id)
                 if (type == 1.toByte() && store.containsKey(options.rpId, data)) {
                     return true
                 }
-            } catch (e: Exception) {
+            } catch (_: Exception) {
                 // Ignore
             }
         }
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
@@ -119,6 +119,9 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
             Log.d(TAG, "origin=$origin, appName=$appName")
 
             // Check if we can directly open screen lock handling
+            // If the request contains an allow list, the request is for a dedicated account.
+            // If we have one of the allowed keys we instant login
+            // Else, the user must be able to choose
             if (!requiresPrivilege && allowInstant) {
                 transportHandlers.firstOrNull {
                     it.isSupported && it.shouldBeUsedInstantly(options)
@@ -139,7 +142,6 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
             val next = if (!requiresPrivilege) {
                 val knownRegistrationTransports = mutableSetOf<Transport>()
                 val allowedTransports = mutableSetOf<Transport>()
-                val localSavedUserKey = mutableSetOf<String>()
                 if (options.type == RequestOptionsType.SIGN) {
                     for (descriptor in options.signOptions.allowList.orEmpty()) {
                         val knownTransport = database.getKnownRegistrationTransport(options.rpId, descriptor.id.toBase64(Base64.URL_SAFE, Base64.NO_WRAP, Base64.NO_PADDING))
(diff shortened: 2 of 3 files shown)


play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/transport/screenlock/ScreenLockTransportHandler.kt
```diff
@@ -259,13 +259,18 @@ class ScreenLockTransportHandler(private val activity: FragmentActivity, callbac
...
+        // may want to login with different user accounts, maybe on a hardware key
+        options.signOptions.allowList?.forEach {
             try {
-                val (type, data) = CredentialId.decodeTypeAndData(descriptor.id)
+                val (type, data) = CredentialId.decodeTypeAndData(it.id)
                 if (type == 1.toByte() && store.containsKey(options.rpId, data)) {
                     return true
                 }
-            } catch (e: Exception) {
+            } catch (_: Exception) {
```
p1gp1g, line 273: I've mainly added a comment here

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
```diff
@@ -54,8 +52,6 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
             else -> null
         }
 
-    private val service: GmsService
```
p1gp1g, line 57: This wasn't used

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
```diff
@@ -95,19 +91,6 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
...
-            // Check if we can directly open screen lock handling
-            if (!requiresPrivilege) {
-                val instantTransport = transportHandlers.firstOrNull { it.isSupported && it.shouldBeUsedInstantly(options) }
-                if (instantTransport != null && instantTransport.transport in INSTANT_SUPPORTED_TRANSPORTS) {
-                    window.setBackgroundDrawable(ColorDrawable(0))
-                    window.statusBarColor = Color.TRANSPARENT
-                    setTheme(org.microg.gms.base.core.R.style.Theme_Translucent)
-                }
-            }
-
```
p1gp1g, line 110: This part was doing work on the UI thread, including DB requests, and redundant: setTheme is called right after the condition to set AppCompat_DayNight_NoAction_Bar, the statusBarColor seems to not have any effect, and setting the background color can be done in the suspended handleRequest

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
```diff
@@ -136,10 +119,15 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
             Log.d(TAG, "origin=$origin, appName=$appName")
 
             // Check if we can directly open screen lock handling
+            // If the request contains an allow list, the request is for a dedicated account.
+            // If we have one of the allowed keys we instant login
+            // Else, the user must be able to choose
             if (!requiresPrivilege && allowInstant) {
-                val instantTransport = transportHandlers.firstOrNull { it.isSupported && it.shouldBeUsedInstantly(options) }
-                if (instantTransport != null && instantTransport.transport in INSTANT_SUPPORTED_TRANSPORTS) {
```
p1gp1g, line 141: INSTANT_SUPPORTED_TRANSPORTS was redundant: this is already what `shouldBeUsedInstantly` does. (And only one transport never return false: the screen lock, but that isn't relevant)

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
```diff
@@ -177,13 +164,11 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
                             }
                         }
                     }
-                    database.getKnownRegistrationInfo(options.rpId).forEach { localSavedUserKey.add(it.userJson) }
```
p1gp1g, line 180: Here was the first part of bug: "Store any key for that website in localSavedUserKey"

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/AuthenticatorActivity.kt
```diff
@@ -177,13 +164,11 @@ class AuthenticatorActivity : AppCompatActivity(), TransportHandlerCallback {
...
                 }
                 val preselectedTransport = knownRegistrationTransports.singleOrNull() ?: allowedTransports.singleOrNull()
                 if (database.wasUsed()) {
-                    if (localSavedUserKey.isNotEmpty()) {
-                        R.id.signInSelectionFragment
```
p1gp1g, line 185: And here the 2nd part "If there is any key for that website in localSavedUserKey, go to signInSelection" => we can't use a USB or NFC key, even if it is the only allowed key

play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/ui/TransportSelectionFragment.kt
```diff
@@ -31,7 +32,15 @@ class TransportSelectionFragment : AuthenticatorActivityFragment() {
...
-            startTransportHandling(SCREEN_LOCK)
+            if (options is PublicKeyCredentialRequestOptions ||
+                options is BrowserPublicKeyCredentialRequestOptions) {
+                findNavController().navigate(
+                    R.id.signInSelectionFragment,
+                    arguments.withIsFirst(false)
+                )
+            } else {
+                startTransportHandling(SCREEN_LOCK)
+            }
```
p1gp1g, line 43: This is the part where we want to be able to choose which account to use during sign-in, but we don't want that view during registration

The previous selection of user account was with the other bug, where it was "there is any key for a website => we go to the account selection for screen lock login".

The current behavior is now:
- "we have an allowCredential => use it"
- "we don't have an allowCredential, go to transport selection, select screen_lock, select user account"
- "we don't have an allowCredential, go to the last transport for this account login or go back to transport selection, select screen_lock, select user account"

p1gp1g:
I'm closing as it has been done part of https://github.com/microg/GmsCore/commit/67cb36c58982664b16f1b2dad323e5a639a8b79f#diff-af0a9e54a0c3e7f58a9ad6530d19b48981841de4c779d3f72fe272356cb9c9e3

PR #3257, opened 2026-01-29
fix google login missing droidguard

DaVinci9196:
1.Avoiding ANR that may occur during login (https://github.com/microg/GmsCore/pull/3192)
2.fix google login missing droidguard
Problem stack：
```
org.microg.gms.common.NotOkayException: Error=MissingDroidguard
 at org.microg.gms.common.HttpFormClient.request(HttpFormClient.java:96)
at org.microg.gms.common.HttpFormClient.lambda$requestAsync$0(HttpFormClient.java:210)
at org.microg.gms.common.HttpFormClient$$ExternalSyntheticLambda0.run(Unknown Source:8)
```

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
+++ b/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
@@ -353,7 +353,7 @@ public class LoginActivity extends AssistantActivity {
                 .token(oAuthToken).isAccessToken()
                 .addAccount()
                 .getAccountId()
-                .droidguardResults(null /*TODO*/)
+                .droidguardResults("null" /*TODO*/)
                 .getResponseAsync(new HttpFormClient.Callback<AuthResponse>() {
                     @Override
                     public void onResponse(AuthResponse response) {
@@ -430,6 +430,7 @@ public class LoginActivity extends AssistantActivity {
                 .hasPermission(true)
                 .addAccount()
                 .getAccountId()
+                .droidguardResults("null")
                 .getResponseAsync(new HttpFormClient.Callback<AuthResponse>() {
                     @Override
                     public void onResponse(AuthResponse response) {


ale5000-git:
I don't know this code too much but does it check if droidguard is enabled in microG settings before using it?

DaVinci9196:
> I don't know this code too much but does it check if droidguard is enabled in microG settings before using it?

Currently, based on my testing, this error occurs when installing Microg for the first time and then logging into a Google account.
It's also possible to determine this by checking the `enable` status of DroidGuard.

ale5000-git:
> Currently, based on my testing, this error occurs when installing Microg for the first time and then logging into a Google account.
> It's also possible to determine this by checking the `enable` status of DroidGuard.

Perhaps I did not express myself clearly.
My question was: Does microG execute droidguard and include `droidguardResults` ONLY if safetynet/playIntegrity is enabled in microG settings?

DaVinci9196:
> > Currently, based on my testing, this error occurs when installing Microg for the first time and then logging into a Google account.
> > It's also possible to determine this by checking the `enable` status of DroidGuard.
> 
> Perhaps I did not express myself clearly. My question was: Does microG execute droidguard and include `droidguardResults` ONLY if safetynet/playIntegrity is enabled in microG settings?

ok

er13:
@DaVinci9196: shouldn't [this line](https://github.com/microg/GmsCore/blob/dbd68770e24be6cd074e73a9bd2117a386753ada/play-services-auth-workaccount/core/src/main/kotlin/com/google/android/gms/auth/account/authenticator/WorkAccountAuthenticator.kt#L103) be changed in the same way as those from https://github.com/microg/GmsCore/pull/3257/changes/3a68bc3f1fd412cd6a9e3534d809b1ad01983423?

er13:
@DaVinci9196 

Btw. instead of changing `null` to `"null"` shouldn't it be just

```diff
--- a/play-services-base/core/src/main/java/org/microg/gms/auth/AuthRequest.java
+++ b/play-services-base/core/src/main/java/org/microg/gms/auth/AuthRequest.java
@@ -68,7 +68,7 @@ public class AuthRequest extends HttpFormClient.Request {
     public boolean getAccountId;
     @RequestContent("ACCESS_TOKEN")
     public boolean isAccessToken;
-    @RequestContent("droidguard_results")
+    @RequestContent("droidguard_results", nullPresent = true)
     public String droidguardResults;
     @RequestContent("has_permission")
     public boolean hasPermission;

```

er13:
@DaVinci9196

Friendly reminder on
https://github.com/microg/GmsCore/pull/3257#issuecomment-3898781367
and
https://github.com/microg/GmsCore/pull/3257#issuecomment-3901636469

er13:
@mar-v-in

Looping in you (as the merger of this pull request), as @DaVinci9196 doesn't show any reaction to my comments listed below

https://github.com/microg/GmsCore/pull/3257#issuecomment-3898781367
and
https://github.com/microg/GmsCore/pull/3257#issuecomment-3901636469

er13:
As both @DaVinci9196 and @mar-v-in do not respond to https://github.com/microg/GmsCore/pull/3257#issuecomment-3898781367 and to https://github.com/microg/GmsCore/pull/3257#issuecomment-3901636469 (probably because they don't receive/read the GitHub notifications) looping in more people

+ @ale5000-git
+ @fynngodau

@DaVinci9196 & @mar-v-in:
Please do enable the GitHub notifications and do read them, or document another way to contact you...

mar-v-in:
Hi @er13

Sorry for this. I receive dozens of notifications from GitHub every day, so some slip through (this one likely because it was on a closed PR).

Indeed the variant you described might be cleaner, you're welcome to open a pull request with it.

er13:
@mar-v-in
Thanks a lot for replying!

@mar-v-in & @DaVinci9196:
I'm not familiar with GmsCore, this is actually the very first time I'm taking a look at its source code. Thus I would prefer to discuss the effects of the changes with you guys first.

Annotating `droidguard_results` with `nullPresent = true` as suggested in https://github.com/microg/GmsCore/pull/3257#issuecomment-3901636469 would also affect some other contextes `droidguardResults` has not been set at all so far.

All together there are the following ones:
- Group 1: `droidguardResults` already set to `"null"`, switching to using the annotation would not change anything

https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java#L349
https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java#L423

- Group 2: `droidguardResults` is set to `null` instead of `"null"`, this is the missing part of https://github.com/microg/GmsCore/pull/3257, adding the annotation would result in setting it to `"null"`

https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-auth-workaccount/core/src/main/kotlin/com/google/android/gms/auth/account/authenticator/WorkAccountAuthenticator.kt#L96

- Group 3: `droidguardResults` is not set so far, adding the annotation would result in setting it to `"null"`. This is the main part I'm not sure if it would break anything and would thus appreciate it if you could comment on it. Thanks a lot!

https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-auth-workaccount/core/src/main/kotlin/com/google/android/gms/auth/account/authenticator/WorkAccountAuthenticator.kt#L157
https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-core/src/main/java/org/microg/gms/auth/AuthManager.java#L331
https://github.com/microg/GmsCore/blob/de94027e96085c7cd532ea6be745718d24f4d23b/play-services-core/src/main/java/org/microg/gms/checkin/CheckinManager.java#L49

The changes are already available in [this branch](https://github.com/er13/GmsCore/commits/google-login-missing-droidguard-improvements/.)

mar-v-in:
@er13 It seems the `droidguard_results` being present is only required for initial login requests, not service token requests. This makes sense, because the latter is also used in some non-Android environments of Google services that don't have access to DroidGuard.

With #3350 now merged, this should resolve all affected instances

PR #3267, opened 2026-02-03
feat(rcs): Implement comprehensive RCS support for Google Messages

SBALAVIGNESH123:
This pull request adds a full, native implementation of the Rich Communication Services (RCS) stack directly into GmsCore, making it possible for devices running microG (without Google Play Services, root access, or attestation) to use RCS seamlessly in Google Messages. It solves microG Issue #2994 by rebuilding the core RCS provisioning and transport layers from the ground up, rather than relying on wrappers or partial shims. As a result, non-GMS devices can now authenticate, complete Jibe provisioning, and exchange RCS messages just like certified devices.

Under the hood, the implementation includes a complete provisioning engine for Google’s Jibe flow, a custom SIP/MSRP client for secure communication with Google’s IMS servers, and a persistent message store for reliable state and message handling. It integrates cleanly with the Android system, falling back to SMS when RCS is unavailable, managing user capability discovery, and using only standard Android permissions. Security and privacy are also addressed, with support for Google-compatible end-to-end encryption and encrypted local storage of credentials. The solution has been tested on an unrooted device with a fresh microG install, successfully completing setup in Google Messages and sending RCS messages, while compiling cleanly against API 34 and targeting Android 10 and above.


SBALAVIGNESH123:
I’ve finished a complete, end-to-end RCS implementation that includes the SIP transport layer, provisioning engine, and capability exchange. The full GmsCore APK builds successfully with these changes (com.google.android.gms-250932000.apk), is signed with the standard microG test keys, and has been tested on an unrooted Android 14 device.

From a technical standpoint, the system uses an event-driven architecture via RcsEventBus to keep UI updates non-blocking, along with advanced telemetry through RcsMetricsCollector to monitor connection health in real time. At its core is a native, custom-built SIP stack (RcsSipClient) designed for stability, efficiency, and low battery impact. This is not a thin wrapper around existing services—GmsCore now actively manages the full connection lifecycle to Google’s Jibe servers, enabling reliable RCS messaging even without Play Services attestation. The implementation is complete and ready for review.

mar-v-in:
Can you please provide more details on how you actually tested this:
- Which RCS app did you use? Google Messages or another one? Which exact version of said app?
- Which phone model, network/sim card provider did you use?
- Which exact Android version did you use (i.e. any custom ROM?)

Sense101:
I smell AI... :(

Issue #3197, opened 2025-12-19
MicroG should use/register as official passkey provider to allow fallbacks/provider selection etc.


rugk:
**Describe the bug**
If you have multiple Passkey providers installed, only the third-party provider is currently used on my system.

**To Reproduce**
Basically install KeePassDX and set it up with Passkey support. Then you can only save/requests passkeys from them anymore, not microG, anymore.
See https://github.com/Kunzisoft/KeePassDX/issues/2310

**Expected behavior**
In the settings, apparently there is this screen:
![](https://private-user-images.githubusercontent.com/19224038/526620919-7ce6428a-9045-4ba4-bb04-2ff7d20f81db.jpg?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjYxNDM0MTYsIm5iZiI6MTc2NjE0MzExNiwicGF0aCI6Ii8xOTIyNDAzOC81MjY2MjA5MTktN2NlNjQyOGEtOTA0NS00YmE0LWJiMDQtMmZmN2QyMGY4MWRiLmpwZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTEyMTklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUxMjE5VDExMTgzNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTRiYWMzNTFkYzU5OGNjNDc5ZDcwNTE4OTNkOTBmMDAyMzk0NmMwMWI1MjMxODQwNzU0OWE4NTk0NThlOGFlNGQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.dTZNv4rwE6VZmiA-f946HFh-6o-6wxr9hdvboDxmFHk)

microG should appear here, so the OS can correctly handle that.

**Screenshots**
Currently it looks like this:

[...]

PR #3284, opened 2026-02-18
feat: introduce CredentialProviderService minimal implementation

jonathanklee:
Introduce CredentialProviderService minimal implementation.

This allows 3rd party applications (like Down Dog Yoga) that are clearing credentials state at start to run properly.

The user would still have to select the microG credential provider manually in his Account preferences.

To make microG credential provider be automatically selected on a specific OS, the config_enabledCredentialProviderService and config_primaryCredentialProviderService resources can be overlayed.

### DIFF
--- a/play-services-core/src/main/AndroidManifest.xml
+++ b/play-services-core/src/main/AndroidManifest.xml
@@ -15,6 +15,7 @@
   -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
+    <uses-permission android:name="android.permission.PROVIDE_DEFAULT_ENABLED_CREDENTIAL_SERVICE" />
 
     <permission
         android:name="com.google.android.providers.gsf.permission.READ_GSERVICES"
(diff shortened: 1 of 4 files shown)


CoelacanthusHex:
Will this PR fix https://github.com/microg/GmsCore/issues/3197 ?

jonathanklee:
@CoelacanthusHex It is going to add a "microG credential provider" entry indeed. But a non-working one, which is enough for some 3rd party applications. 

The fallback for the user here would be to install a 3rd party applications like KeePassDX which implements a credential provider and then let the user select KeePassDX manually in the Account settings. 

Surprisingly, Bitwarden which can be selected as credential provider too, does not seem to make the trick here. Down Dog Yoga is still crashing with a ClearCredentialUnknownException. I have to toggle on KeepPassDX in the Account extra service settings to make Down Dog Yoga app work. So I assume Bitwarden does not implement the mechanism Down Dog Yoga is expected. It's also probably that's why we can select several credential providers in Android account settings.

jonathanklee:
Closing this since a full implementation has been merged.

PR #3327, opened 2026-03-10
Added RCS support 

naushad970:
monitored logcat on device with real play services to see what messages does during rcs setup. it just looks for com.google.android.gms.rcs.START and try to bind IRcsEngineController. when that missing, it show "chat features unavailable" and stop. so i just did same thing in microg. now when messages look for that service, microg return the controller instead of nothing.
tested this on redmi note 8 pro with airtel you can see in video. messages already have jibe client inside. it was never broken. microg was just not returning anything to it. #2994

https://github.com/user-attachments/assets/47de85a3-33e8-4948-9ac2-a4cf95bdd9a3


quimodotcom:
Brother this is the same code I tested and it didnt do anything. This code to me just looks like it broadcasts true messages around the place.

naushad970:
> Brother this is the same code I tested and it didnt do anything. This code to me just looks like it broadcasts true messages around the place.

logs? also if you can share the video

PR #3331, opened 2026-03-13
Add support for Google Maps timeline functionality

DaVinci9196:
Target: Implement timeline functionality in Google Maps.

Two function-related services were briefly implemented.
1. KeyRetrievalService: Used for user key acquisition and verification.
2. SemanticLocationHistoryService: Record and report the user's historical location information.

Because it involves user data reporting, a control switch has been added to the microG settings page.
Users can choose whether to use this feature based on their individual needs.


mar-v-in:
Can we have a local-only version of this feature, so that if users opt-out, they still have all the functionality, just not synchronized to their Google account and other devices?

DaVinci9196:
> Can we have a local-only version of this feature, so that if users opt-out, they still have all the functionality, just not synchronized to their Google account and other devices?

The existing solution includes localized content, so the configuration options need to be adjusted to allow users to choose whether to sync their Google accounts.

TheGamerKing561:
@DaVinci9196 I think the name in the values should be changed to "Google Maps (Timeline)" (with a space between "Maps" and "(Timeline)").

The-Nickname:
Question about this feature, it looks interesting. Can it only be used with Google Maps? Or could the created offline DB also be queried by other applications like OsmAnd or CoMaps (or maybe even in microG directly)?
It that out of scope or should I open an issue about that?

PR #3352, opened 2026-03-21
fix(auth): replace deprecated NetworkInfo with NetworkCapabilities in LoginActivity

William17738:
## Summary

- Replace deprecated `ConnectivityManager.getActiveNetworkInfo()` with the `NetworkCapabilities` API in `LoginActivity.start()`. The new path is gated on API 23 (M) because it relies on `getActiveNetwork()`, even though `NetworkCapabilities` itself is available since API 21.
- Retain fallback to `NetworkInfo.isConnected()` for Android versions below M (project `minSdk` is 19).
- No behavioral change on devices where the old API works correctly.

## Problem

On Android 14, `getActiveNetworkInfo()` returns `null` for non-system apps when the active connection is a VPN (e.g. Clash, v2ray). This causes the login screen to show the "no network" error even though the device has internet access.

This primarily affects users on **stock ROMs** (Honor, Huawei) where microG is installed as a user app rather than a system app. Custom ROM users are unaffected because microG is typically installed as a privileged system app, which has access to the deprecated API.

## Fix

```java
// Before (deprecated, fails on Android 14 with VPN):
NetworkInfo networkInfo = cm.getActiveNetworkInfo();
if (networkInfo != null && networkInfo.isConnected())

// After (works with VPN/proxy connections):
if (SDK_INT >= Build.VERSION_CODES.M) {
    Network network = cm.getActiveNetwork();
    NetworkCapabilities caps = cm.getNetworkCapabilities(network);
    return caps != null && caps.hasCapability(NET_CAPABILITY_INTERNET);
} else {
    NetworkInfo networkInfo = cm.getActiveNetworkInfo();
    return networkInfo != null && networkInfo.isConnected();
}
```

## Test plan

- [ ] Verify login works on Android 14 stock ROM (Honor/Huawei) with VPN active
- [ ] Verify login works on Android 14 stock ROM with WiFi only (no VPN)
- [ ] Verify login still shows "no network" error when truly offline
- [ ] Verify no regression on Android < M (fallback path)

## Scope

This PR only touches `LoginActivity.java`. There are 3 other call sites using the same deprecated API (`TriggerReceiver`, `McsService`, `gcm/TriggerReceiver`) which could be addressed in a follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
+++ b/play-services-core/src/main/java/org/microg/gms/auth/login/LoginActivity.java
@@ -25,6 +25,8 @@ import android.content.Context;
 import android.content.Intent;
 import android.graphics.Color;
 import android.net.ConnectivityManager;
+import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.net.NetworkInfo;
 import android.net.Uri;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


ale5000-git:
Also according to the documentation NetworkCapabilities was added in API level 21 (Lollipop).

William17738:
Thanks for the pointer. I'll update the PR description.

wolfi2424 (approved):

PR #3366, opened 2026-03-27
Click listeners on polylines and polygons

fynngodau:
Mirrors click listener functionality on circles to polylines and polygons.

I did not implement click listeners in lite mode maps because I think that is very rarely used, but requires a lot of code for us to compute the hitbox of the objects.

Tested against: https://github.com/googlemaps-samples/android-samples  
ref. https://gitlab.e.foundation/e/os/backlog/-/work_items/3533


mar-v-in:
Thanks :)

PR #3369, opened 2026-03-31
Fix: database is locked

DaVinci9196:
image
Intermittent issues, not a path to reproducibility.


play-services-droidguard/core/src/main/kotlin/org/microg/gms/droidguard/core/DgDatabaseHelper.kt
```diff
@@ -25,25 +27,42 @@ class DgDatabaseHelper(context: Context) : SQLiteOpenHelper(context, "dg.db", nu
...
+        while (retries > 0) {
+            try {
+                return block()
+            } catch (_: SQLiteDatabaseLockedException) {
+                retries--
+                if (retries == 0) return defaultValue
+                Thread.sleep(100)
+            }
+        }
+        return defaultValue
```
ale5000-git, line 61: Isn't better for debugging to display something in the logcat if all retries failed?

DaVinci9196: Added.

Issue #580, opened 2018-07-17
What's the status on Google Cast implementation?


ale-reyes:
I checked the wiki and the implementation page hasn't been updated in two years, so I thought I'd ask.

benwaffle:
#79

cubagithub:
#79  is Closed.
Can anyone use cast function natively? I can't use it on Android 11 + MicroG using a FireTV or Chromecast.
![signal-2021-03-02-183150_002](https://user-images.githubusercontent.com/34730232/109613824-d3403780-7b85-11eb-84e7-91a819edeb6d.png)
![signal-2021-03-02-183150_003](https://user-images.githubusercontent.com/34730232/109613828-d3d8ce00-7b85-11eb-8729-4f1ae385fbff.png)
![signal-2021-03-02-183150_001](https://user-images.githubusercontent.com/34730232/109613829-d4716480-7b85-11eb-9b0f-d61e57636ba0.png)

ale5000-git:
Fortunately on some ROM the removal is reverted.

hollowshiroyuki:
Hi, I have the same issue using the unofficial Lineage MicroG on a Pixel 3 :
Lineage Version : 19-20220622-microG-blueline (Android 12)
MicroG Version : 0.2.24.214816

Is https://github.com/TeamVanced/Vanced/issues/688#issuecomment-992102591 relevant ?

The chromecast was tested with an unmodified Xiaomi Note 10 too and worked correctly.

matteopessina:
Any news on this feature?

I did not managed to cast Netflix, Display+ directly to chromecast.
I am using /e/OS 1.6-202212208238947, with microg 0.2.25.223616-9 on OnePlus8 IN2013.

I tried debugging the Google sample cast sender application https://github.com/googlecast/CastVideos-android.
On an old Huawei P10 with stock rom it works, while on os with microg it shows the following logcat
[cast-logcat.txt](https://github.com/microg/GmsCore/files/10322469/cast-logcat.txt)

The most interesting part is the following:
```
/GmsDynamiteLoaderImpl: No such module known: com.google.android.gms.cast.framework.dynamite
/GmsDynamiteLoaderImpl: returning temp fix module version for com.google.android.gms.cast.framework.dynamite. Cast API wil not be functional!
```

mattszcz:
The following information seems like it holds the clue to solving the Chromecast casting issue:

[...]

PR #3351, opened 2026-03-21
Cast: fix device connection lifecycle and implement joinApplication

FloodExLLC:
Fixes the two root causes that prevent Cast from working in practice.

### Problem

The `ChromeCast` object (from the chromecast-java-api-v2 library) requires
an explicit `connect()` call to open a TLS socket before any operation can
succeed. The existing code created the `ChromeCast` instance but never called
`connect()`, so `launchApp()`, `sendRawRequest()`, `stopSession()` etc. all
threw `IOException` immediately, making Cast non-functional for all apps.

A second issue was that `joinApplication()` always launched a new instance
instead of joining an existing session, breaking multi-sender scenarios
and session resume.

### Changes

**`CastDeviceControllerImpl`**
- Add `ensureConnected()` helper that calls `chromecast.connect()` if not
  already connected; invoke it at the start of every outgoing operation
- Fix `launchApplication()` to handle a `null` app response gracefully
  rather than NPE-ing on `app.sessionId`
- Implement `joinApplication()` correctly: query device status, join
  the already-running matching session (`wasLaunched = false`), fall back
  to launching only when the app is not present

**`CastMediaRouteController`**
- Implement `onSelect()`: open TLS connection when user selects route
- Implement `onUnselect()` / `onUnselect(int)`: close connection on deselect
- Implement `onRelease()`: close connection when controller is destroyed

Closes #580

PR #3377, opened 2026-04-06
Fix casting

hasali19:
Addresses #580 

This fixes a few issues preventing casting from working properly. My main focus has been to get casting from the youtube app working. I've tested this on a CCwGTV from an emulator (with a hardcoded address since emulator can't do mDNS) and it seems to be able to start a session and control media playback. The disconnect process is a bit broken so still need to dig into that.

### DIFF
--- a/play-services-cast-framework/core/src/main/java/com/google/android/gms/cast/framework/internal/CastContextImpl.java
+++ b/play-services-cast-framework/core/src/main/java/com/google/android/gms/cast/framework/internal/CastContextImpl.java
@@ -71,6 +71,9 @@ public class CastContextImpl extends ICastContext.Stub {
             .addControlCategory(MediaControlIntent.CATEGORY_REMOTE_PLAYBACK)
             .addControlCategory(defaultCategory)
             .build();
+
+        router.registerMediaRouterCallbackImpl(getMergedSelectorAsBundle(), new MediaRouterCallbackImpl(this));
+        router.addCallback(getMergedSelectorAsBundle(), 0);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


aquette:
Great to see some activity to fix this lack (missing screen cast on microG), thanks @hasali19 😀
Anything to be consolidated from https://github.com/microg/GmsCore/pull/3351 ?

hasali19:
The changes in #3351 didn't work for me, in particular the connection attempts in CastMediaRouteController throw errors. I don't think changes are needed to that class anyway (at least not to get basic casting working)

peterhel:
Nice work — this builds clean for me (`:play-services-core:assembleVtmDefaultDebug` → BUILD SUCCESSFUL), which already puts it ahead of a couple of the other Cast PRs, and the connectionless handshake looks right.

Full disclosure so this lands as "someone who walked the same path" rather than a competing claim: I went down the connectionless route too (#3570), reverse-engineering `play-services-cast` 21.4.0 and validating it against the real Amazon Prime Video app on a hardware Chromecast (Prime launched its own receiver and played a DRM title). The notable thing: **we independently converged on the same wire contract.** Your `connect() = 16`, `addListener(...) = 17`, and the listener callback as an **`int` at `= 13`** all match what I landed on (I named the last `onConnectedWithResult(int)` and the register call `setListener` — cosmetic, since AIDL dispatches by transaction id). That's a good signal it's right, especially since #3502 put `onConnected(String sessionId)` at that listener slot, which I suspect is wrong — two independent RE passes landing on `int` is reassuring.

**Small bug:** `CastDeviceControllerImpl.onConnected(int)` dereferences `listener` without a null check, unlike every other `onXxx` callback in the file — if a connection event lands before `addListener()` runs, it'll NPE. A `if (this.listener != null)` guard (matching the siblings) fixes it.

**Minor:** the `"///ALLOW_IPV6"` suffix in `CastMediaControlIntent.categoryForCast()` works but is a bit magic — a one-line comment on the why would help. And in `SessionImpl`, heads-up that `onApplicationConnectionFailure()` nulls `castContext` then dereferences it a couple lines later (pre-existing, not yours, but it'll NPE on the failure path).

Happy to validate on real hardware — I've got a Chromecast-with-Google-TV rig, and since this is YouTube-focused I can try YouTube specifically and report what the receiver does. Thanks for pushing this forward.

peterhel:
Tested #3377 on real hardware (Pixel 2 + Chromecast, connectionless path): **it casts — video plays end-to-end** (`PLAYING`, currentTime advancing, the Chromecast pulled the stream). Your impl + `onConnected(int)` work in practice, not just the emulator. 🎉

The session was initially unstable, though: microG's `CastDynamiteModuleImpl.newReconnectionServiceImpl` returns null, and your `CastContextImpl` wiring makes the SDK start the `ReconnectionService`, which NPEs on it (`...zzq.onCreate() on a null object reference`, play-services-cast-framework 19.0.0) → that crash-loops the sender process → which is what threw the stream of `onSendMessageSuccess` / `onTextMessageReceived` `RemoteException`s (callbacks landing on a dying listener).

I implemented a minimal no-op `newReconnectionServiceImpl` (an `IReconnectionService.Stub` — no-op `onCreate`/`onDestroy`, `onStartCommand=START_NOT_STICKY`, `onBind=null`) and rebuilt: **crash gone, the `RemoteException`s gone, and the cast is clean and stable** — video plays, sender stays alive, zero errors.

Since that `newReconnectionServiceImpl` gap is microG-level (it'll hit any connectionless PR that wires up reconnection, not just this one), I'm putting it up as a small separate fix. With it in place, #3377 casts cleanly here. Nice work getting the connectionless path this far.

paulcakeface:
I traced two remaining framework lifecycle bugs against the Cast 21.4.0 client flow. One fixes the existing application-connection failure NPE in `SessionImpl`. The other wires the normal `endCurrentSession` path through `disconnectFromDevice` and `notifySessionEnded`, including `stopCasting` handling and preserving the session ID until the end callback has fired.

I tested both on top of #3377 plus #3577. The combined Cast framework build passes. I also wrote three focused regression tests. They are 3/3 passing with the fixes and 0/3 on the same #3377 + #3577 baseline without them, including reproducing the original failure-path NPE.

The fixes are kept as separate narrow commits so they can be reviewed or cherry-picked independently:

- https://github.com/paulcakeface/GmsCore/commit/5f41fdaf287294c8281418d3d132b338d5b2876e
- https://github.com/paulcakeface/GmsCore/commit/9ae3cae7ce260d56a2468eb2a2e5dcccc8129daa

I can provide the test patch as well if useful.

Issue #2852, opened 2025-04-07
Add a /vendor path for microg.xml search path to comply with strict build requirements


karlbaumg:
**Is your feature request related to a problem? Please describe.**

In LineageOS, and other distributions relying on `lineageos4microg` project to include microG, we are having issues with using a `/system`-based path to find the `microg.xml` in builds that are marked as strict which doesn't let system partition to have much additions, see https://github.com/lineageos4microg/android_vendor_partner_gms/issues/44

However, in microG, putting a [`/system/etc/microg.xml`](https://github.com/microg/GmsCore/blob/0cefb34/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt#L57) is the only way to provide build-time default settings.

**Describe the solution you'd like**

Another path such as `/vendor/etc/microg.xml` can be added such that microG checks that first and if it doesn't exist, then fall back to `/system/etc/microg.xml` so that no existing builds break and we can switch to `/vendor/etc/microg.xml` in future builds.

**Describe alternatives you've considered**

We tried to add it as exception but it doesn't work in strict builds https://github.com/lineageos4microg/android_vendor_partner_gms/pull/46

petefoth:
> **Describe the solution you'd like**
> 
> Another path such as `/vendor/etc/microg.xml` can be added such that microG checks that first and if it doesn't exist, then fall back to `/system/etc/microg.xml` so that no existing builds break and we can switch to `/vendor/etc/microg.xml` in future builds.

At the moment, los4microg puts the file in `/product/etc/`, so having GmsCore look there before falling back to `/system/etc` would be best for us 😄

steadfasterX:
.. and for me `system_ext/etc/` would be the best place ;) 
according to https://source.android.com/docs/core/architecture/partitions/product-interfaces#about-the-artifact-path-requirements `system_ext` or `product` are the correct placements for this these days so it would be great to have the choice even (if possible).

ale5000-git:
I think, if possible, it would be best if it first search on the same partition of the apk (whatever it is) and then fallback to system.

abranson:
Hi, this would be really handy for the integration work we're doing on Sailfish OS. I've made a PR that checks for a microg.xml the original installation partition of the apk (product or system_ext), and merges it over any that might be on the system one. It wasn't particularly clean to do so, but it was the best was I could find to make sure it's still used if the package is updated on-device and is now somewhere else.

https://github.com/microg/GmsCore/pull/3394

PR #3394, opened 2026-04-14
Settings: Read defaults from the bundled partition (#2852)

abranson:
Use `ApplicationInfo` partition flags to detect whether GmsCore was originally bundled on `product` or `system_ext`, and read `etc/microg.xml` from that partition. Keep `/system/etc/microg.xml` as a shared defaults source underneath partition-specific overrides.

### DIFF
--- a/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
+++ b/play-services-base/core/src/main/kotlin/org/microg/gms/settings/SettingsProvider.kt
@@ -9,6 +9,7 @@ import android.content.ContentProvider
 import android.content.ContentValues
 import android.content.Context
 import android.content.Context.MODE_PRIVATE
+import android.content.pm.ApplicationInfo
 import android.content.SharedPreferences
 import android.database.Cursor
 import android.database.MatrixCursor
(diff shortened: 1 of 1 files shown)


mar-v-in:
Any reason to only allow `/product/etc/microg.xml` when the package is installed on `/product`? Same for `/system_ext`?

How do those internal methods (`isProduct` and `isSystemExt`) behave when a package that was originally installed in `/product` or`/system_ext` is updated by the user?

abranson:
> Any reason to only allow `/product/etc/microg.xml` when the package is installed on `/product`? Same for `/system_ext`?

It was the suggestion here: https://github.com/microg/GmsCore/issues/2852#issuecomment-2786319469 to "first search on the same partition of the apk (whatever it is) and then fallback to system." and I thought that made a lot of sense. If we bundle then it should all be in the same partition. Falling back to system is only needed for backwards compat, but I think that's necessary.

> How do those internal methods (isProduct and isSystemExt) behave when a package that was originally installed in /product or/system_ext is updated by the user?

There isn't any public docs that I found for those, but I couldn't use sourceDir precisely because it does not preserve the original installed location. These flags do get preserved: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/services/core/java/com/android/server/pm/InstallPackageHelper.java#2101 
They are hidden API, but I think it's exactly what they're used for.

ale5000-git:
@mar-v-in
On Android various standard Android xml files are restricted to the same partition, so I think it make sense to follow the trend.

Also I suggest to change the path to `/<partition-name>/etc/<package-name>/settings.xml` => example: `/product/etc/com.google.android.gms/settings.xml`, and use `/system/etc/microg.xml` only as final fallback.
In this way a fork with a different package name (at compile time) would have a different path.

abranson:
Now it checks /$partition/etc/$packageName/settings.xml, then /$partition/etc/microg.xml, and finally /system/etc/microg.xml. I've seen anticipation of /$partition/etc/microg.xml so I thought it would be nice to leave that in.

abranson:
Rebased, how is this looking now?

mar-v-in (approved):

ale5000-git:
@mar-v-in
Could you please use `/$partition/etc/$packageName/settings.xml` also for the system partition and keep `/system/etc/microg.xml` only as **deprecated** fallback at the end?

abranson:
Done

mar-v-in (approved):

PR #3401, opened 2026-04-15
reuse module context in mapbox

zappybiby:
Updates the Mapbox backend to use the module resources/classloader already selected by `MapsContextLoader`, instead of rebuilding a GmsCore package context from `Constants.GMS_PACKAGE_NAME`.

Changes:
- `DynamiteContext`: forwards `getResources()` / `getAssets()` to the selected GmsCore context. Previously only package/application/classloader state was forwarded.
- `CreatorImpl.initV2()`: stores the module resources for `MapContext` and installs `MultiArchLoader` with the backend classloader. Previously it only initialized bitmap descriptor resources.
- `MapContext`: wraps the app context and serves the module resources initialized by `CreatorImpl`. Previously it created a new package context with `Constants.GMS_PACKAGE_NAME`.
- `MultiArchLoader`: loads native libraries through the backend classloader. Previously it derived ABI/package/version/APK metadata from a hardcoded GmsCore package. 
 
Related to: [ReVanced/GmsCore#308](https://github.com/ReVanced/GmsCore/pull/308#discussion_r3011594911) and in [microg/GmsCore#3367](https://github.com/microg/GmsCore/issues/3367).

### DIFF
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
@@ -60,7 +60,10 @@ class MapContext(private val context: Context) : ContextWrapper(context) {
     }
 
     override fun getTheme(): Resources.Theme {
-        return theme ?: resources.newTheme().also { theme = it }
+        return theme ?: resources.newTheme().also {
+            it.setTo(context.theme)
+            theme = it
+        }
     }
 
     override fun getClassLoader(): ClassLoader {
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
@@ -50,7 +50,7 @@ class MapContext(private val context: Context) : ContextWrapper(createModuleCont
     }
 
     override fun getClassLoader(): ClassLoader {
-        return getModuleClassLoader()
+        return MapContext::class.java.classLoader!!
     }
 
     override fun getSharedPreferences(name: String?, mode: Int): SharedPreferences {
@@ -76,17 +76,6 @@ class MapContext(private val context: Context) : ContextWrapper(createModuleCont
     }
 
     companion object {
-        private var moduleClassLoader: ClassLoader? = null
-
-        @JvmStatic
-        fun setModuleEnvironment(classLoader: ClassLoader) {
-            moduleClassLoader = classLoader
-        }
-
-        private fun getModuleClassLoader(): ClassLoader {
-            return moduleClassLoader ?: throw IllegalStateException("Class loader has not been initialized")
-        }
-
         private fun createModuleContext(context: Context): Context {
             return context.createPackageContext(Constants.GMS_PACKAGE_NAME, Context.CONTEXT_INCLUDE_CODE or Context.CONTEXT_IGNORE_SECURITY)
         }
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
@@ -16,14 +16,16 @@
 
 package org.microg.gms.maps.mapbox.utils
 
+import android.content.res.AssetManager
 import android.content.Context
 import android.content.ContextWrapper
 import android.content.Intent
+import android.content.res.Resources
 import android.content.SharedPreferences
 import android.view.LayoutInflater
 import java.io.File
 
-class MapContext(private val context: Context, mapsContext: Context) : ContextWrapper(mapsContext) {
+class MapContext(private val context: Context) : ContextWrapper(context) {
     private var layoutInflater: LayoutInflater? = null
     private val appContext: Context
         get() = context.applicationContext ?: context
@@ -46,6 +48,14 @@ class MapContext(private val context: Context, mapsContext: Context) : ContextWr
 
     override fun getPackageName(): String {
         return appContext.packageName
+    }
+
+    override fun getResources(): Resources {
+        return getModuleResources()
+    }
+
+    override fun getAssets(): AssetManager {
+        return resources.assets
     }
 
     override fun getClassLoader(): ClassLoader {
@@ -75,6 +85,17 @@ class MapContext(private val context: Context, mapsContext: Context) : ContextWr
     }
 
     companion object {
+        private var moduleResources: Resources? = null
+
+        @JvmStatic
+        fun setModuleResources(resources: Resources) {
+            moduleResources = resources
+        }
+
+        private fun getModuleResources(): Resources {
+            return moduleResources ?: throw IllegalStateException("Resources have not been initialized")
+        }
+
         val TAG = "GmsMapContext"
     }
 }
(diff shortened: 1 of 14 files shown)


mar-v-in:
The API between `play-services-maps` and `play-services-maps-core` is defined by Google, not us. If we change it, things are going to break.
- The interface `com.google.android.gms.maps.internal.ICreator` is from Google.
- The requirement of a no-argument constructor (specifically without a `Context` argument) is from Google.

This change will break things in two ways:
- Apps that use the `play-services-maps` library from Google (that is most apps displaying a map) will not be able to do so anymore when run on a device with microG (they will try to invoke the no-argument constructor on `com.google.android.gms.maps.internal.CreatorImpl` which no longer exists)
- Apps that use the `play-services-maps` library from microG (probably very few) will not be able to do so anymore when run on a device with original Google Play Services (they will try to invoke the constructor with `Context` argument, which doesn't exist in Play Services).

zappybiby:
> The API between `play-services-maps` and `play-services-maps-core` is defined by Google, not us. If we change it, things are going to break.
> 
> * The interface `com.google.android.gms.maps.internal.ICreator` is from Google.
> * The requirement of a no-argument constructor (specifically without a `Context` argument) is from Google.

Understood, I will review options. That was part of my pre https://github.com/microg/GmsCore/pull/3401/changes/f867ebdf25ae4e2b81d8bd55473cfef5e1ed26c2 change, but you bring up some other points I need to consider as well. I will look into this, thank you for your feedback.

mar-v-in:
I think you are tackling this from the wrong angle. Instead, figure out what inside `play-services-maps-core-mapbox` currently requires access to the context of the gms package and either eliminate those needs or use other techniques to fulfill those needs. As we do have access to the class loader that loads `play-services-maps-core-mapbox`, this should already give us almost all we need.

play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
```diff
@@ -49,8 +50,16 @@ class MapContext(private val context: Context) : ContextWrapper(context.createPa
...
+        return getModuleResources()
+    }
+
+    override fun getAssets(): AssetManager {
+        return resources.assets
+    }
+
     override fun getClassLoader(): ClassLoader {
-        return MapContext::class.java.classLoader!!
+        return getModuleClassLoader()
```
mar-v-in, line 53: Aren't those class loaders effectively the same?

play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MultiArchLoader.kt
```diff
@@ -65,17 +37,8 @@ class MultiArchLoader(private val mapContext: Context, private val appContext: C
...
-        var len: Int = inp.read(buffer)
-        while (len >= 0) {
-            out.write(buffer, 0, len)
-            len = inp.read(buffer)
-        }
-
-        inp.close()
-        out.close()
+    private fun findLibrary(name: String): String? {
+        return (moduleClassLoader as? BaseDexClassLoader)?.findLibrary(name)
```
mar-v-in, line 41: Have you verified this in a multi-arch environment?

That is, when the app using the `play-services-maps` library is build for a different architecture than the GMS package is installed at. The easiest way to test is to force install any app using maps as `armeabi-v7a` on a device that is `arm64-v8a` (most devices today). You can do this using `adb install --abi armeabi-v7a target.apk`

oSumAtrIX:
Hey @mar-v-in 
If this PR looks sound, we would like to merge it our repository. Would it be possible to get a green light from you or if let us know if there's any immediate issues

play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MapContext.kt
zappybiby: They were the same. 

Early in the PR, `initV2()` captured the classloader for CreatorImpl that would then pass to MultiArchLoader. And the idea was that the captured loader in `MapContext.getClassLoader()` could just be used for consistency, but you are correct that `MapContext::class.java.classLoader!!` resolves to the same loader. This was a semantic change that I shouldn't have introduced and later reverted.

The `MapContext.getClassLoader()` part has since been reverted. `CreatorImpl.class.getClassLoader()` value is still passed directly to `MultiArchLoader`

play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/utils/MultiArchLoader.kt
zappybiby: I unfortunately don't have an android device that supports multi-ABI / armeabi-v7a.

The best I can do is with emulation with x86/x86_64, and that was also limited by the fact that only API 30 seemingly has support for this kind of setup.

GmsCore installed as `x86_64` and maps clients force-installed as `x86` on an API 30 `default;x86_64` emulator (`ro.product.cpu.abilist=x86_64,x86`, `ro.zygote=zygote64_32`).

First a shim app was used to test `MapsInitializer.initialize()` -> `MapView` creation/lifecycle -> `getMapAsync` -> `moveCamera` + `addMarker` after `onMapReady`.

| Build | Commit | Result |
| --- | --- | --- |
| upstream master | `352f2d72fa52c6c3c4fdd79d575a071a0da72ad1` | passed: Mapbox init, x86 `libmapbox-gl.so` load, `onMapReady`, no native/link/signature error |
| PR commit | `df278ba779f8803477d2596ba2be76fa6effb166` | passed: Mapbox init, x86 `libmapbox-gl.so` load, `onMapReady`, no native/link/signature error |
| PR head | `3c59f775e0f202ac1fd74d8e53636bafc7c060f5` | passed: Mapbox init, x86 `libmapbox-gl.so` load, `onMapReady`, no native/link/signature error |

I also tested with the app Trulia `14.38.0`, installed as `x86`

| Build | Commit | Result |
| --- | --- | --- |
| upstream master | `352f2d72fa52c6c3c4fdd79d575a071a0da72ad1` | passed: maps dynamite context, Mapbox init, x86 native load, no native/link/signature error |
| PR commit | `df278ba779f8803477d2596ba2be76fa6effb166` | passed: maps dynamite context, Mapbox init, x86 native load, no native/link/signature error |
| PR head | `3c59f775e0f202ac1fd74d8e53636bafc7c060f5` | passed: maps dynamite context, Mapbox init, x86 native load, no native/link/signature error |

[REPORT.md](https://github.com/user-attachments/files/28365980/REPORT.md)

zappybiby:
@oSumAtrIX just FYI, one of the earlier commits df278ba779f8803477d2596ba2be76fa6effb166 (or something like it) would be more ideal for Revanced because it handles one of the remaining mapbox uses of GMS_PACKAGE_NAME

In df278ba779f8803477d2596ba2be76fa6effb166 it derives the package from the classloader that loaded the Mapbox backend, then creates the package context from that. This seems to work just fine, however, it does this in a hacky way, using BaseDexClassLoader reflection. I didn't think the maintainer would be okay with that kind of hack, and I had reverted it because I was hoping to find a better solution (which of course hasn't happened).

The better solution, if it is possible, would be to make DynamiteContext expose the selected GmsCore resources to then 
make Mapbox consume the resources/classloader it already receives through dynamite setup. This follows the principle of "use what the loaded Mapbox module has instead of reconstructing GmsCore context from" better too. 

The issue I encountered with this approach previously was that DynamiteContext doesn't forward getResources() and getAssets(), which caused an Resources$NotFoundException error. 

Let me see if I can get something working tonight for it.

edit: pushed https://github.com/microg/GmsCore/pull/3401/commits/e24010c781d732fab4c9cd497c3cbbbca1f15951

oSumAtrIX:
I see, does the new commit satisfy our needs? I am fine with any change or solution merged from this PR into this repository and downstreaming it into our repo.

ak800i:
@mar-v-in could you please re-review this PR? Much obliged

PR #3425, opened 2026-04-22
Fix RCS Support

gallonc61-jpg:
Added RCS / Jibe support by injecting required system properties to make Google Messages detect device as certified.

✅ Spoofs Pixel 8 Pro properties
✅ Correct fingerprint & client IDs
✅ Works with latest Google Messages
✅ No root required
✅ Solves "RCS not available" and setup loop


quimodotcom:
oh bud

you tried

PR #3427, opened 2026-04-23
Wearable: Add ConnectionInfo features to WearableService

DaVinci9196:
Fix Mi Fitness (com.xiaomi.wearable) prompts to update Google Play.

### DIFF
--- a/play-services-wearable/core/src/main/java/org/microg/gms/wearable/WearableService.java
+++ b/play-services-wearable/core/src/main/java/org/microg/gms/wearable/WearableService.java
@@ -18,6 +18,9 @@ package org.microg.gms.wearable;
 
 import android.os.RemoteException;
 
+import com.google.android.gms.common.Feature;
+import com.google.android.gms.common.api.CommonStatusCodes;
+import com.google.android.gms.common.internal.ConnectionInfo;
 import com.google.android.gms.common.internal.GetServiceRequest;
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

peterhel:
This is the right fix, and it builds clean here on current `master`. 👍

Full disclosure for context: I implemented the same pattern for connectionless Cast (#3570) — microG was calling `onPostInitComplete(…, null)` with no `ConnectionInfo`, so the SDK judged it too old and returned `ConnectionResult=2`; moving to `onPostInitCompleteWithConnectionInfo` with the features cleared it. Same root cause here.

One thing from that experience worth a thought: on the Cast side I *echoed back* `request.apiFeatures` (whatever the client asked for) rather than a fixed list, so the advertised set always matches what the client probes. A hardcoded list of 33 works *as long as it's a superset* of what the Wear companion actually checks — but a client probing a feature outside the list would still hit the "too old" path. Did you confirm the list covers what the companion verifies (or is echoing `request.apiFeatures` viable here too)? The mechanism itself is correct either way.

mar-v-in:
@peterhel `request.apiFeatures` includes all features the client may ask for, but also includes features it doesn't use and especially includes features where it has a fallback when they are not present. Announcing features not supported (or not verified to be required) may cause more problems than not doing so, when the client side has fallback logic and the service side only has the old API implemented. That's why I would generally advice against echoing `request.apiFeatures`.

PR #3445, opened 2026-05-06
Settings: Added Passkey Management Page

DaVinci9196:
Add a "Manage passkeys" entry in the account settings overflow menu (next to game account management) that lists locally saved FIDO passkeys.
Tapping a row shows a delete confirmation that clears both the SQLite record and the matching Android Keystore alias, letting users clean up residual entries left behind when an app removed a passkey without notifying microG. 

e.g. Discord(com.discord); Created and deleted within the app.


bhzad766htd (approved):

Diapolo:
Is this to come 🤓?

PR #3448, opened 2026-05-07
Return reauth intent for expired microG credentials

TheScarastic:
When app requests for token and then gmscore receivesBadAuthentication and asks to reauthenticate, The gmscore doesnt reponds back anything back to the caller causing apps to get stuck in a bad state waiting for token, this PR handles the issue by retruning the reauthentication intent to the caller


TheScarastic:
@mar-v-in can you please check

PR #3460, opened 2026-05-11
Fido: Fix relying party validation

p1gp1g:
The RP ID must be equal to the origin's effective domain, or a registrable domain suffix of the origin's effective domain

cf. comment for details

### DIFF
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
@@ -172,7 +172,7 @@ suspend fun RequestOptions.checkIsValid(context: Context, origin: String, packag
         // 2. it excludes internal domains as topDomainOf requires a public FQDN
         //
         // Instead, we first check that rpId is valid, then if it matches the origin host
-        if (InternetDomainName.from(rpId).isPublicSuffix) {
+        if (runCatching { InternetDomainName.from(rpId).isPublicSuffix }.getOrDefault(false)) {
             throw RequestHandlingException(NOT_ALLOWED_ERR, "RP ID $rpId is a public suffix")
         }
         if (
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/RequestHandling.kt
@@ -9,6 +9,7 @@ import android.content.Context
 import android.net.Uri
 import android.util.Base64
 import android.util.Log
+import androidx.core.net.toUri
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

peterhel:
The registrable-suffix check reads more spec-correct than the old `topDomainOf(...)` comparison — thanks for tightening this. Builds fine here.

Small robustness note: `InternetDomainName.from(rpId)` throws `IllegalArgumentException` on a malformed `rpId`, and it isn't wrapped, so a bad `rpId` would surface as a crash rather than a clean `NOT_ALLOWED_ERR`. Might be worth a `runCatching { InternetDomainName.from(rpId).isPublicSuffix }.getOrDefault(false)` (or treating unparseable as not-allowed). Either way the validation logic itself is an improvement.

p1gp1g:
Thanks @peterhel for the suggestion, it's definitely better to catch the exception :+1:

p1gp1g:
I've pushed a new commit to catch the exception.

I think we need to catch the exception in some other functions in this file

peterhel:
Thanks for the quick turnaround! Pulled 560440e0 — builds clean here, and `runCatching { … }.getOrDefault(false)` is exactly right: an unparseable rpId now falls through to the host check and gets the proper `NOT_ALLOWED_ERR`. 👍

On catching it elsewhere in the file — agreed; happy to take a pass through the other parse/validation spots and flag any worth the same guard.

peterhel:
Did the pass through the rest of `RequestHandling.kt` for the same unguarded-throw pattern, and the picture's actually pretty good:

- The other `InternetDomainName.from` / `topDomainOf` / `Uri.parse` call sites (lines 138/142) all live inside `isAppIdAllowed`, which already wraps its body in a broad `try { … } catch (e: Exception) { false }` — so a malformed `rpId`/`appId` there resolves to a clean `NOT_ALLOWED_ERR` rather than a crash.
- The `Base64.decode(sha256facetId.substring(28), …)` in the apk-key-hash branch decodes microG-computed signature digests (from `getAllFacetIdCandidates`), not RP-supplied data, and the `substring(28)` is safe because the string is selected by its `android:apk-key-hash-sha256:` prefix.

So as far as I can see, the line you just fixed was the only RP-reachable crash path — nothing else exposed today.

One forward-looking note rather than an ask: `topDomainOf` itself throws (`InternetDomainName.from` → `IllegalArgumentException`, `topDomainUnderRegistrySuffix` → `IllegalStateException`), so any *future* caller outside a `try`/`catch` would reintroduce exactly this issue — might be worth hardening the helper itself (return null on bad input) so callers can't get it wrong. But that's optional and out of scope here.

If you had a specific other function in mind, point me at it and I'll take a closer look. 🙂

mar-v-in:
Thanks :pray:

PR #3461, opened 2026-05-11
Fido: Fix COSE decoding with x-only keys

p1gp1g:
Some EC keys uses only the x-coordinate (cf. https://datatracker.ietf.org/doc/html/rfc9053#name-elliptic-curve-keys), and used to crash while trying to `GetByteString`.

This fix the Cbor decoding for these keys

### DIFF
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/CoseKey.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/CoseKey.kt
@@ -22,7 +22,7 @@ import java.security.spec.ECPublicKeySpec
 class CoseKey(
     val algorithm: Algorithm,
     val x: ByteArray,
-    val y: ByteArray,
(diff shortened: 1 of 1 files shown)


peterhel:
Nice fix — the `runCatching` + nullable `y` lets x-only / OKP keys decode instead of throwing in `GetByteString()`. Builds clean here on current `master`.

One follow-up I hit while reading it: `asCryptoKey()` still does

```kotlin
ECPublicKeySpec(ECPoint(BigInteger(1, x), BigInteger(1, y)), parameterSpec)
```

so a key that now decodes with `y == null` will NPE there (`BigInteger(1, null)`) rather than failing cleanly at decode time. It pairs naturally with #3545, which reworks key construction for Ed25519/OKP — so either landing them together or guarding `asCryptoKey()` for the null-`y` case would close that gap. Happy to test against a real x-only assertion if useful.

p1gp1g:
I should have push both in a single PR indeed

PR #3467, opened 2026-05-12
ProviderInstaller: Update defaultTlsProtocol

DaVinci9196:
Upgrade defaultTlsProtocol from "TLSv1.2" to "TLSv1.3".

Fixed an issue where product preview images displayed incorrectly within the McDonald's Offers and Delivery (com.mcdo.mcdonalds) app.


ale5000-git:
This seems a nice change in general, but have you tested it on old Android versions?
I don't have time to test it now, but it is possible that it doesn't work correctly on them.

matteozanettii (approved):

Neustradamus:
It will be nice to have the details what TLS version is supported by Android versions.

mar-v-in:
@ale5000-git this is the code that injects our custom version of conscrypt, it doesn't affect the system TLS provider. The custom version is the same on every Android version and should also support TLS1.3 everywhere.

mar-v-in (approved):

PR #3476, opened 2026-05-16
[BOUNTY #2994] Preserve phone hints in RCS sync requests

mt211211:
/claim #2994

This is a stacked follow-up on top of the active RCS/Constellation implementation in #3359. It is not intended to supersede that work; it isolates a small provisioning edge case from the latest #2994 logcat discussion so it can be reviewed or cherry-picked.

Focused change in the final commit:
- keep Google Messages' requested phoneNumberHint when SubscriptionInfo phone formatting cannot produce E.164
- fall back to the raw SIM number instead of an empty sim_readable_number when formatting fails
- make nullable targetedSims, verificationMethodsValues, extras, policyId, and idTokenRequest handling safe while building SyncRequest
- keep requested IMSI filtering from aborting when targetedSims is absent

Why this matters:
The current live-device logs moved past IID token acquisition but still showed provisioning failing around verifyPhoneNumber, followed by Google Messages reporting that no phone number was found for the provisioning id. Sending an empty sim_readable_number after formatting failure loses useful caller-provided phone data. This patch preserves that data while still preferring normalized E.164 when available.

Validation:
- git diff --check passes locally
- I could not run Gradle locally because this machine has no Java/JAVA_HOME on PATH

Maintainer note: if a full duplicate stacked PR is noisy, the intended cherry-pick is the last commit only: 2efd1bc (Handle RCS sync request phone fallbacks).


coffeeisangry:
So LLMs have devolved into just yoinking code adding like 2 changes then trying to claim the bounty for it? Wtf

PR #3489, opened 2026-05-21
feat: Add RCS Support

Misch369:
Implemented RCS Support via new play-services-rcs module, including IRcsService and PhoneNumberVerificationService.

### DIFF
--- a/play-services-rcs/build.gradle
+++ b/play-services-rcs/build.gradle
@@ -1,0 +1,11 @@
+apply plugin: 'com.android.library'
+
+android {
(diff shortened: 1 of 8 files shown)


quimodotcom:
why tf has the ai bot started tryna implement wearables too 😭

Misch369:
Ah my bad! I forgot to checkout \main\ before creating this branch, so it accidentally included all the WearOS commits from my other PR. Just force-pushed to clean up the git history! It should only contain the RCS modules now.

PR #3388, opened 2026-04-09
Implement Constellation + Asterism services for RCS on Google Messages

Lyapsus:
### Summary

Implements the two GMS services Google Messages requires for RCS:
- **Constellation** (svc 155) -- phone number verification via `phonedeviceverification-pa.googleapis.com`
- **Asterism** (svc 199) -- consent management (Google ToS acceptance for RCS)

Also fixes several DroidGuard issues that prevented tachyon transport registration.

(hopefully) closes #2994

This is the initial working implementation -- I obviously plan to clean up the code, reduce the diff, and address review feedback until everyone is happy with it.

### What works

**End result: E2EE RCS messages sent and received on microG.** Tested on multiple SIMs of Spusu (232/17) and Georg (232/12) Austrian Jibe UPI carriers (Google-handled phone verification -- most of Europe, many worldwide). Google Messages v302730063 (2026-03-10 build).

### Testing limitations

**Not tested:**
- Proper target device: locked bootloader, no root. Sadly only rooted Samsung Galaxy S10+ is available to me for testing but I believe there were no modifications that would be necessary on target device.
- Non-Samsung devices. Samsung auto-discovers phone numbers via proprietary API; LineageOS/CalyxOS are expected to need manual entry flow.
- TS.43 carriers (SIM-based EAP-AKA -- T-Mobile US, Freedom Mobile, Deutsche Telekom, etc.): Full implementation in `Ts43Client.java` (EAP-AKA + FIPS 186-2 PRF + ODSA Phase 2). Sadly no TS.43 SIM available at hand for end-to-end testing so far, therefore testing would be especially appreciated.

### Attribution and tooling

This work was developed independently since at least 18 January 2026 based on commits dates, with specific components adopted from parallel efforts:

- **@benwaffle**: Proto definitions (from Google API discovery) used as starting point; extended with fields from GMS decompilation.
- **@opstic**: `MoSmsVerifier` ported from PRs #3359-3361. `gmsVersion >= 25.19.31` requirement was their discovery.
- **Claude Code** (Anthropic's CLI) was used as a development tool throughout -- for code generation, decompilation analysis, test orchestration, hypothesis testing across 226 internally documented sessions.

PR #3495, opened 2026-05-22
[RCS] Harden Constellation callers and PNV capabilities

mt211211:
Builds on the existing Constellation/Asterism RCS implementation line from #3388 and preserves that commit history for attribution. This branch adds a focused hardening and behavior layer on top of it.

Changes in this commit:

- Apply the RCS caller allowlist to Constellation service binding, matching the Asterism gate already in the branch.
- Tie Constellation binder instances to the verified package and re-check the calling UID before `verifyPhoneNumber`, `getIidToken`, or `getPnvCapabilities` can touch SIM/IID/provisioning paths.
- Replace the empty `getPnvCapabilities` response with per-subscription capabilities derived from requested methods or active SIMs, so callers can see TS.43/EAP-AKA, SMS OTP, and silent-verification support instead of a blank capability list.
- Add unit coverage for caller allowlists and PNV capability mapping.

Why this matters:

The RCS provisioning surface does more than return a binder. It can read telephony state, create IID tokens, invoke TS.43, and fall back to Google Constellation flows. Returning an ungated Constellation binder and then advertising no PNV capabilities is both weaker than the Asterism side and less useful to Google Messages during setup.

Validation:

```powershell
.\gradlew.bat :play-services-core:testVtmDefaultDebugUnitTest --tests org.microg.gms.constellation.ConstellationServiceImplTest --tests org.microg.gms.rcs.RcsCallerPolicyTest
```

Result: 32 tests, 0 failures, 0 errors. `ConstellationServiceImplTest` ran 29 tests; `RcsCallerPolicyTest` ran 3 tests.


eepymeowers:
AI slop bot, do not merge and definitely don't give the bounty to this

PR #3507, opened 2026-05-26
Fix the issue requiring a security verification flow for in-app subscriptions

DaVinci9196:
Fix the issue where in-app subscriptions require going through a security authentication page (e.g., YouTube in some European regions)


mar-v-in:
Can we do this without changing gradle and wire versions?

DaVinci9196:
> Can we do this without changing gradle and wire versions?

No. Older versions of Wire have a bug where Wire fails to parse protobuf data when there are very large field numbers (e.g., 268435456). Some of the field numbers defined in our proto files exceed this value, which causes the data to fail to parse. However, newer versions of Wire have fixed this issue.

PR #3529, opened 2026-06-04
fix(wearable): TOS pairing fix and foundation utilities for WearOS su…

maxifyhub:
…pport


ItsJustSomeDude:
Slop. Splitting into a bunch of different PRs doesn't make it any better.

The fact that it's sending "/microg/..." to the watch is a dead giveaway the slop cannon that wrote this had no idea what it was doing.

It should be obvious: the whole pile of PRs should be closed.

PR #3544, opened 2026-06-08
Fido: Fix request with FidoAppIdExtension

p1gp1g:
The JSON key is `appid`, and if it was present we were trying to get `appId` which always failed with an exception "No value for appId"

It caused a crash with Gitlab for example

### DIFF
--- a/play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/provider/JsonExtensions.kt
+++ b/play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/provider/JsonExtensions.kt
@@ -181,7 +181,7 @@ fun JSONObject.parseTokenBinding() = when (TokenBindingStatus.fromString(getStri
 
 fun JSONObject.parseAuthenticationExtensions(): AuthenticationExtensions {
     val builder = AuthenticationExtensions.Builder()
-    if (has("fidoAppIdExtension")) builder.setFido2Extension(FidoAppIdExtension(getJSONObject("fidoAppIdExtension").getString("appId")))
+    if (has("fidoAppIdExtension")) builder.setFido2Extension(FidoAppIdExtension(getJSONObject("fidoAppIdExtension").getString("appid")))
     if (has("appid")) builder.setFido2Extension(FidoAppIdExtension(getString("appid")))
     // TODO: Add support for other extensions
     return builder.build()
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

play-services-core/src/main/kotlin/org/microg/gms/auth/credentials/provider/JsonExtensions.kt
```diff
@@ -182,7 +182,7 @@ fun JSONObject.parseTokenBinding() = when (TokenBindingStatus.fromString(getStri
 fun JSONObject.parseAuthenticationExtensions(): AuthenticationExtensions {
     val builder = AuthenticationExtensions.Builder()
     if (has("fidoAppIdExtension")) builder.setFido2Extension(FidoAppIdExtension(getJSONObject("fidoAppIdExtension").getString("appId")))
```
mar-v-in:
```suggestion
    if (has("fidoAppIdExtension")) builder.setFido2Extension(FidoAppIdExtension(getJSONObject("fidoAppIdExtension").getString("appid")))
```

This one also needs to be fixed.

p1gp1g: I didn't find any doc regarding `fidoAppIdExtension`, do you have one?

Also, I think it may be worth catching issues here, because we don't check if `fidoAppIdExtension` object has a `appid` field - and we can't be sure there isn't any type error

```suggestion
    if (has("fidoAppIdExtension")) runCatching {
        builder.setFido2Extension(FidoAppIdExtension(getJSONObject("fidoAppIdExtension").getString("appid")))
    }
```

Note that we can probably do a `runCatching` for the other `appid` field

mar-v-in: Rather than ignoring individual malformed parts of the JSON, I think it safer to throw here and error out (so that malformed JSONs just fail entirely).

p1gp1g: IMO that's better to ignore this particular field if it is malformed, as it is a non-standard field, for a legacy compatibility

Issue #3549, opened 2026-06-09
[com.revolut.revolut] Latest versions crash trying to reach GmsDynamiteLoaderImpl


quimodotcom:
**Affected app**
Name: Revolut
Package id: com.revolut.revolut

**Describe the bug**
Revolut crashes as microG does not include a value for getModuleVersion for the GMS dynamic maps implementation.

And crashes.

**To Reproduce**
1. Install MicroG
2. Sign into revolut
3. Crash.

**Expected behavior**
No crash

**Temp Fix**
Turn on airplane mode, open and login to revolut, then turn it off.

**System**
Android Version: 16
Custom ROM: Project Infinitt X Macanc

**microG**
microG Core version: Latest Opstic Build with RCS
microG Self-Check results: All Ticked

**Additional context**

PR: #3552

ItsJustSomeDude:
What behavior does the app have if you do the same thing on regular Play Services? Basically, what does the regular "no api key" code path look like? Because MicroG should seek to replicate that behavior. (Which means it probably shouldn't crash...)

quimodotcom:
> What behavior does the app have if you do the same thing on regular Play Services? Basically, what does the regular "no api key" code path look like? Because MicroG should seek to replicate that behavior. (Which means it probably shouldn't crash...)

The API Key error was one of a few the new revolut fails at and to be honest I should update this issue as thats not the main one it crashes at. The main one it seems to fail on the most is the use of GmsDynamiteLoaderImpl and some missing values, mainly getModuleVersion which GmsCore didnt seem to implement. I have fixed the issue by adding a Dummy stub for the dynamic loader which includes getModuleVersion + various other needed values.

[11_06-17-24-30_923.log](https://github.com/user-attachments/files/28847058/11_06-17-24-30_923.log)

Edit: updated this issue to reflect the thing it actually crashes on.

quimodotcom:
Update: my fork somehow is missing half my fixes. Ive taken it down for the meantime.

quimodotcom:
added refactored and COMPLETE fix

quimodotcom:
closed. this was a bug with a third party gms client

PR #3552, opened 2026-06-10
[MAPBOX] fix: missing Maps/Stadia api key app crashes

quimodotcom:
Fixes issue #3549.

Right now the code for the map part just assumes there is always going to be a valid api key. Without one it crashes some apps; Revolut is the one that this happens to for me as of Q1 2026 updates.

This pull request adds a fallback called DummyGoogleMapImpl and basically shows nothing instead.

The class acts like the real google map one but returns a blank frame layout. MapFragment and MapView were updated to check for keys and use the dummy when nothing is found. CreatorImpl also got changed in its get renderer type method so it returns zero without a key.

It seems like the goal is to make builds more stable when no map key was added. Some parts still feel a bit unclear about how it all connects though.


quimodotcom:
can I get some improvement advice or can this be pushed asap, since im now stuck using my banking app on another device 😭

Issue #3522, opened 2026-06-01
[com.snapchat.android] Snapchat flickering when selecting filters


akunterkontrolle:
**Affected app**
Name: Snapchat
Package id: com.snapchat.android

**Describe the bug**
When using Snapchat and selecting a filter the image will start flickering rapidly and the filter will not be applied. A snap taken will either just be blank (older versions of Snapchat on Android 13) or just the image without the filter (Android 16 with current version of Snapchat 13.94.0.45) or some similar behavior. On Android 16 with microG 0.3.15.250932 not all filters are affected but most. On older Android versions with older microG versions and older Snapchat versions all filters are affected.

**To Reproduce**
1. Open 'Snapchat'
2. Select Front Camera (effect is the same on the back camera)
3. Select a filter
4. Camera Screen starts flickering and filter isn't applied. If not, try a few different filters

**Expected behavior**
No camera flicker, filter is applied e.g. to face

**System**
Android Version: 16, same behavior on e.g. Android 13
Custom ROM: Lineage 23.2

**microG**
microG Core version: 0.3.15.250932
microG Self-Check results: All ticked

**Additional context**
The problem exists since some time last year, also reported over the months by a few people with different ROM/microG combinations, e.g. https://www.reddit.com/r/LineageOS/comments/1llztpj/snapchat_camera_flickering_when_using_filters/?show=original, https://www.reddit.com/r/MicroG/comments/1nyd4he/help_snapchat_camera_flicker_on_oneplus_12/, https://www.reddit.com/r/LineageOS/comments/1llztpj/snapchat_camera_flickering_when_using_filters/

One workaround is to forcibly disable microG by removing its permission for signature spoofing (if possible with installation method and ROM combination), launching Snapchat (it complains about no Google Services) and then re-enable microG without closing Snapchat. This makes Snapchat work as expected while keeping the Cloud-Messaging working and stopping Snapchat from complaining.

I can supply logs if needed. I am just not sure how to use logcat to just get the relevant snapchat logs. The only thing I saw scrolling by from the logs was something about FaceDetector and gms. Does this maybe play a role?

akunterkontrolle:
The above pull request provides an AI generated fix for the problem. The code works on my test devices and fixes the Snapchat issue. Wether it introduces any other regressions, actually works for all cases and/or is of acceptable quality is out of my understanding, so it is just provided "as is".

ptlc8:
Exact same problem with exact same versions 🫡

[...]

PR #3578, opened 2026-06-23
Fix Snaptchat flickering issue by patching FaceDetector (#3522)

akunterkontrolle:
This pull request fixes the issue described in #3522 at least on my devices. Please note, that this fix has been entirely generated with a LLM. Since I am neither familiar with Android development in general or microg in particular nor very good with Java, I can't really review the changes properly. However, it does fix the issue on my 2 test devices, so I want to provide at least the code here ...

- Replace direct ByteBuffer.array() with duplicate()-based toByteArray() to avoid permanently consuming the buffer position on reuse.
- Replace ByteBuffer backed by lazy with nullable var + try-catch
- Replace YuvImage.compressToJpeg() (causes libhwui.so MTE crash) with manual NV21→ARGB8888 pixel conversion.
- Add empty-buffer guards in detectFaces and nv21toBitmap.


ptlc8:
bumping, someone to review? 🥺

PR #3595, opened 2026-07-02
Implement TileOverlay in the Mapbox renderer

ak800i:
## What

Implements `GoogleMap.addTileOverlay(...)` in the Mapbox/MapLibre renderer. It was previously a no-op stub:

```kotlin
override fun addTileOverlay(options: TileOverlayOptions): ITileOverlayDelegate? {
    Log.d(TAG, "unimplemented Method: addTileOverlay")
    return TileOverlayImpl(this, "t${tileId++}", options)
}
```

so any app that draws a custom tile layer through a `TileProvider` (weather layers, custom basemaps, heat/density overlays, …) got an invisible overlay.

## Why

Concretely, this makes **Google Photos' Map / "Places" photo-density heatmap** render. Photos draws that heatmap as a `TileOverlay` whose `TileProvider` generates each tile locally from the on-device photo GPS data. With the stub, only the single representative marker showed and the density layer was missing. There is no multi-pin fallback in Photos — the heatmap *is* the `TileOverlay` — so implementing the API is the only way to surface it.

## How

`TileProvider` is a pull/callback API (`getTile(x, y, zoom) -> Tile{ data }`), but MapLibre-Android's `RasterSource` can only fetch tiles from a **URL** — there is no callback-based tile source in the public Android API (unlike maplibre-gl-js).

So this bridges the two with a tiny loopback HTTP server:

- `TileProviderServer` binds a `ServerSocket` to `127.0.0.1` on an ephemeral port and serves each registered `TileProvider` under a globally-unique token: `http://127.0.0.1:<port>/<token>/<epoch>/{z}/{x}/{y}`.
- `TileOverlayImpl` registers its provider, adds a `RasterSource` pointing at that local URL plus a `RasterLayer`, and maps the GMS overlay properties onto MapLibre (`transparency`/`visible` → `raster-opacity`, `fadeIn` → `raster-fade-duration`).
- `remove()`, `clearTileCache()` and map `onDestroy()` tear down the layer/source and unregister the provider.

The server is loopback-only (not reachable off-device), started lazily on first tile overlay, and daemon-threaded.

## clearTileCache() handling

MapLibre caches raster tiles by URL, so simply removing and re-adding a source with an unchanged URL serves stale bytes and never re-invokes the provider. To make `clearTileCache()` actually force a re-render, the tile URL carries a cache-busting `<epoch>` segment: `clearTileCache()` bumps the epoch, the re-added source gets a fresh URL → guaranteed cache miss → provider re-invoked, while normal panning within an epoch still benefits from HTTP `max-age`. This avoids needing a separate in-process LRU cache.

## Alternatives considered

If maintainers prefer to avoid a loopback socket, the same bridge could instead be done via MapLibre's HTTP request interceptor / a custom URL scheme. I went with the loopback server because it's self-contained and doesn't depend on interceptor internals, but I'm happy to rework it if there's a preferred approach.

## Testing

Built and verified on a physical device (Android 14): Google Photos' Places map now renders the full photo-density heatmap over the browsed region, tiles are served from the loopback server, and opacity/visibility/remove behave correctly. Compiles clean against `master` (`:play-services-maps-core-mapbox:assembleRelease`).

## Notes

- Scope is limited to the tile-overlay API; no other behavior changes.
- `addGroundOverlay` remains stubbed and could be implemented similarly (MapLibre `ImageSource` + `RasterLayer`) in a follow-up.

## Relation to prior work

This overlaps with #3481 (@depau's TileOverlay PoC) and the discussion in #3058. I opened this before finding those — apologies for the parallel effort. Rather than compete, I've commented on #3481 comparing the approaches and offered to converge.

Both bridge MapLibre's URL-only `RasterSource` differently:
- #3481 uses a MapLibre HTTP request interceptor (needs reflection on the pinned Mapbox version to swap `moduleProvider`, plus an LRU cache; currently based on #3058).
- This PR uses a loopback `ServerSocket` on `127.0.0.1` — no reflection, standalone on `master`.

`clearTileCache()` is handled here via a cache-busting epoch in the tile URL (bumped on clear → fresh URL → cache miss → provider re-invoked), so no separate LRU cache is needed. Happy to fold this into #3481 or defer, whichever maintainers prefer.

### DIFF
--- a/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
+++ b/play-services-maps/core/mapbox/src/main/kotlin/org/microg/gms/maps/mapbox/GoogleMap.kt
@@ -117,6 +117,7 @@ class GoogleMapImpl(context: Context, var options: GoogleMapOptions) : AbstractG
 
     var groundId = 0L
     var tileId = 0L
+    val pendingTileOverlays = mutableSetOf<TileOverlayImpl>()
 
     var storedMapType: Int = options.mapType
     var mapStyle: MapStyleOptions? = null
@@ -352,8 +353,16 @@ class GoogleMapImpl(context: Context, var options: GoogleMapOptions) : AbstractG
     }
 
     override fun addTileOverlay(options: TileOverlayOptions): ITileOverlayDelegate? {
-        Log.d(TAG, "unimplemented Method: addTileOverlay")
-        return TileOverlayImpl(this, "t${tileId++}", options)
+        val tileOverlay = TileOverlayImpl(this, "t${tileId++}", options)
+        synchronized(this) {
+            val style = map?.style
+            if (style != null && style.isFullyLoaded) {
+                tileOverlay.update(style)
+            } else {
+                pendingTileOverlays.add(tileOverlay)
+            }
+        }
+        return tileOverlay
     }
 
     override fun addCircle(options: CircleOptions): ICircleDelegate {
@@ -780,6 +789,9 @@ class GoogleMapImpl(context: Context, var options: GoogleMapOptions) : AbstractG
                 pendingMarkers.forEach { it.update(symbolManager) }
                 pendingMarkers.clear()
 
+                pendingTileOverlays.forEach { overlay -> overlay.update(it) }
+                pendingTileOverlays.clear()
+
                 pendingBitmaps.forEach { map -> it.addImage(map.key, map.value) }
                 pendingBitmaps.clear()
 
@@ -867,6 +879,7 @@ class GoogleMapImpl(context: Context, var options: GoogleMapOptions) : AbstractG
         symbolManager = null
         currentInfoWindow?.close()
         pendingMarkers.clear()
+        pendingTileOverlays.clear()
         markers.clear()
         BitmapDescriptorFactoryImpl.unregisterMap(map)
         view.removeView(mapView)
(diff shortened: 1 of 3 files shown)


sujunmin:
Tested this PR successfully with Google Photos Places.

### Test environment

* Device: Google Pixel 7a
* Android: 17 / API 37
* Non-root
* Base: Morphe MicroG-RE 7.1.1
* MicroG-RE versionCode: `255070107`
* Google Photos: `7.92.0.977185651`
* De-Vanced Patches: `1.4.0` local build
* Morphe Manager: `1.30.0`
* Morphe Patcher: `1.13.0`

I rebased my test branch onto the exact MicroG-RE 7.1.1 / `255070107` baseline and cherry-picked both commits from this PR:

* `feat(maps): Implement TileOverlay in the Mapbox renderer`
* `maps: Bust MapLibre tile cache on TileOverlay.clearTileCache()`

### Before applying this PR

The MicroG-RE Mapbox renderer still had:

```text
unimplemented Method: addTileOverlay
```

Google Photos Collections → Places could open and the OpenFreeMap basemap rendered, but:

* photo-density heatmap was missing;
* most place entries showed `0 photos`;
* place/photo information was incomplete.

### After applying this PR

Immediately after replacing MicroG-RE with the test build, Google Photos Places initially still showed no photo overlay.

A full device reboot was required.

After reboot:

* OpenFreeMap renders normally.
* Moving or zooming the map causes the Google Photos photo-density heatmap to appear.
* Place photo counts begin populating as map data is loaded.
* Some place counts are initially missing or stale.
* Reopening or refreshing Google Photos causes additional Places list counts to update.

The Places list therefore appears to have some lazy-loading/cache behavior: the list does not always update immediately after TileOverlay data becomes available, but reopening Google Photos refreshes the displayed counts.

The important result is that the TileOverlay implementation is definitely being exercised by Google Photos and enables its Places photo-density heatmap.

In my test, a full device reboot after replacing GmsCore was necessary before the overlay started working.

ale5000-git:
Hi,
another implementation got picked in the end, but your work is still very appreciated :smiley:

PR #3667, opened 2026-07-23
Fix MCS disconnects on aggressive OEM process managers

areteruhiro:
## Summary

Keep the long-lived MCS connection in a foreground service even when the app is exempt from battery optimizations.

## Problem

On a Xiaomi device, Cloud Messaging connected briefly and then immediately returned to the disconnected state. The device log showed Xiaomi's low-memory killer terminating the persistent process within seconds:

```text
Kill 'app.revanced.android.gms:persistent' ..., oom_score_adj 905
```

The current `ForegroundServiceContext` skips foreground-service mode when `PowerManager.isIgnoringBatteryOptimizations()` returns `true`. On this device, the battery exemption did not prevent the OEM process manager from assigning the persistent process a cached OOM score and killing it.

Once the persistent process was killed, the MCS socket disappeared and Cloud Messaging reported that it was disconnected.

## Changes

- Add an opt-in `forceForeground` mode to `ForegroundServiceContext`.
- Use it when starting the MCS connection and heartbeat service.
- Preserve the existing behavior for all other callers.

## Verification

- `:play-services-core:assembleMapboxDefaultDebug --no-daemon --no-configuration-cache --max-workers=1`
- Build completed successfully and produced the Mapbox/default debug APK.
- The change only touches `ForegroundServiceContext.java` and `TriggerReceiver.java`.

### DIFF
--- a/play-services-base/core/src/main/java/org/microg/gms/common/ForegroundServiceContext.java
+++ b/play-services-base/core/src/main/java/org/microg/gms/common/ForegroundServiceContext.java
@@ -20,14 +20,20 @@ import static android.os.Build.VERSION.SDK_INT;
 public class ForegroundServiceContext extends ContextWrapper {
     private static final String TAG = "ForegroundService";
     public static final String EXTRA_FOREGROUND = "foreground";
+    private final boolean forceForeground;
 
     public ForegroundServiceContext(Context base) {
+        this(base, false);
+    }
+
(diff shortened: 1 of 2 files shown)


mar-v-in:
We don't want to spam all users (the majority of which don't use Xiaomi) with a notification not needed for them. The easiest way for you to solve this issue is to not exempt microG from battery optimizations to trigger the notification.

Issue #3564, opened 2026-06-15
FIDO/CTAP : Duplicate PIN verification issue with Yubikey 5c NFC on Android 15


GN998:
Describe the bug
When registering on webauthn.io with the configuration shown in the screenshot, the system prompts for the PIN a second time after the initial PIN verification and tapping the Yubikey.

To Reproduce

Configure the options on webauthn.io as shown in the screenshot.

Click on the registration button.

Verify the PIN when prompted, then tap the Yubikey 5c NFC.

Notice that the system incorrectly requests the PIN verification again.

Expected behavior
Registration should succeed immediately after the initial key tap, rather than prompting for the PIN a second time.

Screenshots

<img width="620" height="475" alt="Image" src="https://github.com/user-attachments/assets/8b28e1fa-80b2-4303-9c72-c83e330a999f" />

System
Android Version: 15
Custom ROM:  Android
Hardware: Yubikey 5c NFC

Additional context
Found this issue during testing with s1m/hw-fido2-provider. Notably, other custom security keys work perfectly fine.

PR #3677, opened 2026-07-27
FIDO: fix infinite PIN verification loop when UV is discouraged …

GN998:
[Duplicate PIN verification issue with Yubikey 5c NFC on Android 15](https://github.com/microg/GmsCore/issues/3564)

Fixed an infinite PIN prompt loop. Tested successfully on YubiKey 5C NFC v5.7.1


GN998:
I have decided to close this PR. There are two main reasons:

First: The original patch may not have completely fixed the issue.
Second: Further testing revealed a new issue where core extension fields (Extensions) might be directly dropped.

Suspected potential issue: Account registration and login in hw-fido2-provider. If registered in hw-fido2-provider, it cannot be used directly to log in on Windows 11 (via NFC and USB). Tested using a Google account in the test environment: Win11 25H2, Chrome 152.0.7977.83.

PR #3685, opened 2026-07-30
feat(maps): implement addTileOverlay via MapLibre raster source

jonathanklee:
This is an implementation of the missing addTileOverlay() method.

A RasterSource is now pointed at a sentinel URL so MapLibre requests the overlay tiles over HTTP. 

An OkHttp interceptor catches those requests and answers them from the app's TileProvider.getTile, so the tiles render as if downloaded.

For instance, we are now able to display the rain radar in a weather map:

image


ItsJustSomeDude:
Related: #3481, #3595

mar-v-in (approved):

PR #3711, opened 2026-08-13
DroidGuard: fix ServiceCallProxy Binder forwarding

DaVinci9196:
Expand reflection arguments correctly when forwarding.
To fix the issue where Zedge(net.zedge.android) prompts that a newer version of Zedge is required.


mar-v-in (approved):

ItsJustSomeDude:
Is there a chance this is the same fix for #3753? It's a similar message, "Failed to bind to encryption key transfer service", and it's a new app update that caused it to start throwing an error.

PR #3724, opened 2026-08-17
fix(fitness): align step recording and history with GMS behavior

DaVinci9196:
Fix step count remaining at zero in apps such as Pikmin Bloom.
- validate calling packages for Fitness Binder services
- persist and restore recording subscriptions
- store steps as timestamped events with reset handling
- align history reads, bucketing, deletion, and RawDataSet parceling


ale5000-git:
`android.permission.ACTIVITY_RECOGNITION` is only for new Android versions, in old versions it use `com.google.android.gms.permission.ACTIVITY_RECOGNITION`.

I'm not sure but maybe the code should declare to use both and check the old permission on old Android versions.

PR #3727, opened 2026-08-20
Fido: Add dialog to select key from hardware

p1gp1g:
Useful when multiple keys are available, for example to select an account when we have resident keys for different accounts

The screenshot below shows the dialog when I have 2 accounts on my yubikey:

Screenshot of a connection, with a dialog *Select an account* with 2 entries

Note that:
- if I have a single key, the dialog is not shown
- for the moment, we must keep the key connected during the dialog selection

### DIFF
--- a/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetNextAssertion.kt
+++ b/play-services-fido/core/src/main/kotlin/org/microg/gms/fido/core/protocol/msgs/AuthenticatorGetNextAssertion.kt
@@ -1,0 +1,22 @@
+/*
+ * SPDX-FileCopyrightText: 2022 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.microg.gms.fido.core.protocol.msgs
+
+import com.upokecenter.cbor.CBORObject
+
+class AuthenticatorGetNextAssertionCommand :
+    Ctap2Command<AuthenticatorGetNextAssertionRequest, AuthenticatorGetAssertionResponse>(
+        AuthenticatorGetNextAssertionRequest()
+    ) {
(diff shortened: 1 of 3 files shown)


GN998:
I’d like to offer a suggestion.

Suppose there are 10 accounts under the same domain and I am using NFC.  ~~If time runs out, the current code only reads and loads a portion of the discoverable credentials.~~
If disconnected, the current code only reads and loads a portion of the discoverable credentials.

Suggestion:
If credentials haven't been fully read, the system should prompt the user to try again, wait until everything is completely loaded, and then allow them to select an account.

Optional UX suggestion:
While reading the account list, we could use an animated ellipsis (., .., ...) to visually indicate that loading is in progress.

p1gp1g:
If it times out, there could be a dialog saying we couldn't read all the accounts, and give an option to either use one already available, or try again. I think it fits another PR - I don't plan to look at this right now, and I'm not even sure it can happen without voluntarily remove the NFC key as fast as possible after tapping the key

There is yet another thing that may be improve: the key needs to stay connected right now, but it souldn't be necessary

GN998:
@p1gp1g 


Thanks for the reply!

Due to a translation issue earlier, I might not have expressed this clearly. What I meant to say is: if someone has 10–20 accounts and experiences a manual disconnection or connection instability while reading, the key might only load a portion of the credentials. Therefore, my suggestion is to enforce reading all credentials completely before proceeding.

Also, since the current loading UI isn't very noticeable, using a spinner or an animated ellipsis ("...") to clearly indicate that reading is in progress could significantly improve the user experience.

Normally, if I'm sitting at a desk, placing the card on the table keeps the connection stable. But when using it on the go, I have to hold the card against the phone. Since many manufacturers place the NFC chip at the top near the camera module, and NFC connections can be unstable on certain devices, disconnections happen quite easily. A visual loading indicator like this would help users consciously keep the card pressed firmly against the phone until finished.

PR #3730, opened 2026-08-21
GeocodeProvider: Return more useful error msg when it's disabled

luk1337:


mar-v-in (approved):

Fs00:
Thanks for this, I remember being puzzled by that "disabled" message the first time I saw it.

@mar-v-in I've always wondered: why is the geocoding API disabled by default? It isn't necessarily called with the user location, and even in that case I think that OSM can be considered a trusted provider.
When disabled, apps that rely on it are likely to crash. One notable example where this happens is a popular payments app in Italy, which uses the reverse geocoder to let the user search for a real world location to then list merchants in its proximity.

mar-v-in:
@Fs00 I agree for most people having it enabled by default is probably fine, but it would change the status quo of having mostly everything disabled by default.

My idea would be to finally introduce a first-time setup that helps configuring permissions and setup the configuration. Ideally in a way that it can be integrated with AOSP SetupWizard pattern, so OS that use microG can integrate it into the setup flow. But that's a very different issue than this helpful string improvement.

PR #3737, opened 2026-08-25
Location: Fix speed calculation for sub-second updates

DaVinci9196:
Calculate derived speed using millisecond precision and only set it
for valid time deltas and finite results.

This prevents Infinity or NaN speed values from reaching location clients.

### DIFF
--- a/play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LastLocationCapsule.kt
+++ b/play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LastLocationCapsule.kt
@@ -66,14 +66,21 @@ class LastLocationCapsule(private val context: Context) {
     fun updateCoarseLocation(location: Location) {
         location.elapsedRealtimeNanos = min(location.elapsedRealtimeNanos, SystemClock.elapsedRealtimeNanos())
         location.time = min(location.time, System.currentTimeMillis())
-        if (lastCoarseLocation != null && lastCoarseLocation!!.elapsedMillis + EXTENSION_CLIFF > location.elapsedMillis) {
-            if (!location.hasSpeed()) {
-                location.speed = lastCoarseLocation!!.distanceTo(location) / ((location.elapsedMillis - lastCoarseLocation!!.elapsedMillis) / 1000)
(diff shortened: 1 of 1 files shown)


mar-v-in (approved):

play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LastLocationCapsule.kt
```diff
@@ -66,14 +66,21 @@ class LastLocationCapsule(private val context: Context) {
...
+                    val speed = previousLocation.distanceTo(location) * 1000f / elapsedMillis
+                    if (speed.isFinite()) {
+                        location.speed = speed
+                        LocationCompat.setSpeedAccuracyMetersPerSecond(location, speed)
+                    }
+                }
+                if (!location.hasBearing() && location.speed > 0.5f) {
+                    location.bearing = previousLocation.bearingTo(location)
+                    LocationCompat.setBearingAccuracyDegrees(location, 180.0f)
+                }
```
ale5000-git, line 83: @mar-v-in
I'm not sure but: what happens if the speed wasn't finite so it remain unset and then it check `location.speed > 0.5f`?
Maybe it need another `location.hasSpeed()` check?

play-services-location/core/src/main/kotlin/org/microg/gms/location/manager/LastLocationCapsule.kt
```diff
@@ -66,14 +66,21 @@ class LastLocationCapsule(private val context: Context) {
...
+            val elapsedMillis = location.elapsedMillis - previousLocation.elapsedMillis
+            if (elapsedMillis in 1..<EXTENSION_CLIFF) {
+                if (!location.hasSpeed()) {
+                    val speed = previousLocation.distanceTo(location) * 1000f / elapsedMillis
+                    if (speed.isFinite()) {
+                        location.speed = speed
+                        LocationCompat.setSpeedAccuracyMetersPerSecond(location, speed)
+                    }
+                }
+                if (!location.hasBearing() && location.speed > 0.5f) {
```
mar-v-in:
```suggestion
                if (!location.hasBearing() && location.hasSpeed() && location.speed > 0.5f) {
```

PR #3747, opened 2026-08-28
feat(wearable): add ANCS notification attributes parser and model

digivasserver-ai:
### What
The first, fully self-contained block toward the notification bridge for [WearOS support (#2843)](https://github.com/microg/GmsCore/issues/2843): a pure-JVM ANCS (Apple Notification Center Service) notification-attributes parser and model.

- `AncsNotifications.parse(byte[])` decodes a *Notification Attributes* Data Source frame into a structured value object (`notificationUid`, `appIdentifier`, `title`, `subtitle`, `message`, `messageSize`, `date`, positive/negative action labels).
- `AncsNotifications.readAttributesRequest(uid, ids…)` builds the matching Control Point request.
- Strict bounds checking: truncated frames and overlength attribute values are rejected instead of yielding partial state; unknown/future attribute IDs are skipped so the parser stays forward compatible.
- The model implements `com.google.android.gms.wearable.AncsNotification`, ready to be routed via `WearableListenerService#onNotificationReceived`.

### Why
The `AncsNotification` API surface is currently an empty interface and `injectAncsNotificationForTesting` is unimplemented — there is no notification path at all on the wear side. This is the parsing/transport-adjacent core that the companion notification flow needs regardless of which Bluetooth transport lands first.

### Verification
- 7 unit tests (`AncsNotificationsTest`) covering full frames, the `MESSAGE_SIZE` uint32 special case, unknown-attribute skipping, and malformed/truncated rejection — all passing (verified standalone with `junit:junit:4.13.2`).

### Next steps (tracked, device-dependent)
- Notification sync engine feeding parsed notifications to `WearableListenerService`.
- Companion pairing/transport wiring so frames arrive from the wear device.

### DIFF
--- a/play-services-wearable/core/build.gradle
+++ b/play-services-wearable/core/build.gradle
@@ -15,6 +15,8 @@ dependencies {
     implementation project(':play-services-wearable')
 
     implementation "org.microg:wearable:$wearableVersion"
+
+    testImplementation 'junit:junit:4.13.2'
 }
 
 android {
--- a/play-services-wearable/core/src/main/java/org/microg/gms/wearable/companion/WearableCompanionBridge.java
+++ b/play-services-wearable/core/src/main/java/org/microg/gms/wearable/companion/WearableCompanionBridge.java
@@ -1,0 +1,169 @@
+/*
+ * SPDX-FileCopyrightText: 2026 microG Project Team
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.microg.gms.wearable.companion;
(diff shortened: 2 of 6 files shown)


TobiPeterG:
Can you please provide a video of you pairing your watch with your phone using this version?

digivasserver-ai:
@/tmp/opencode/wearos-reply.md

digivasserver-ai:
@/tmp/opencode/wearos-correction.md

PR #3768, opened 2026-09-01
feat(vending): add opt-in purchase recovery

HtheB:
Allow users to scan genuine Google Play purchase history and explicitly restore missing non-consumable one-time receipts to FakeStore. Keep subscriptions read-only, avoid overwriting cached entries, and support scoped undo.


ale5000-git:
It is no longer called FakeStore, could you please replace all entries of the name?

HtheB:
Made a new PR 
https://github.com/microg/GmsCore/pull/3797

PR #3779, opened 2026-09-02
Support side-by-side installation with Google services and GMS features

rixxuu:


mar-v-in:
This makes no sense for us. We want to replace GMS, not be installed side-by-side.

PR #3803, opened 2026-09-15
play-services-wearable: Implement client library code for getConnectedNodes and getLocalNode

gloskull:
Motivation

Expose node discovery to Wearable clients by wiring NodeApi to the wearable binder service so apps can query connected peers and the local node.
Description

Implement NodeApi.getConnectedNodes to call the wearable service via GmsConnector.call and forward the binder response to clients.
Implement NodeApi.getLocalNode to retrieve the local node from the service and return a GetLocalNodeResult backed by the service response.
Add response-backed result wrappers GetConnectedNodesResultImpl and GetLocalNodeResultImpl that map service status codes to Status and return immutable node lists.
Add necessary imports and use BaseWearableCallbacks to receive GetConnectedNodesResponse / GetLocalNodeResponse from the service in play-services-wearable/src/main/java/org/microg/gms/wearable/NodeApiImpl.java.
Testing

Ran git diff --check with no issues.
Attempted to run ./gradlew :play-services-wearable:compileDebugJavaWithJavac, but the build failed in this environment due to the toolchain/JDK mismatch: Gradle/Groovy reported Unsupported class file major version 69 (environment’s JDK 25), so compilation could not be completed here.


TobiPeterG:
AI Slop, can be closed

mar-v-in:
This doesn't solve any imminent issue, especially not #2843

PR #3823, opened 2026-09-19
Auth: add Google Auth AANG service compatibility

paulcakeface:
## Summary

Add support for the Google Auth AANG service used by current Google clients.

This adds GMS service 343 and the `com.google.android.gms.auth.aang.events.services.START` action, along with the Binder and parcelable surface needed for accounts, tokens, capability checks and app restrictions.

Only `google_auth_api` is advertised because that is the feature actually implemented here.

## Changes

The PR adds the AANG request and response parcelables, Binder interfaces and `GoogleAuthAangService`.

The service implements:

* GetAccounts
* GetToken
* clearToken
* HasCapabilities
* FetchAppRestriction

Existing microG Auth code is reused for the actual account and token work rather than adding another authentication path.

`clearToken` uses the existing Android account token invalidation path.

Logging was also kept deliberately minimal so account names, tokens and other sensitive request values are not written to logs.

There are focused tests covering the service registration, Binder transaction order and the parcelable field IDs.

## Validation

Base: upstream `master` at `4c74e5acb79479004428be755547432294639878`

Head: `326a448541595a2628b8ced52716bd608b39f264`

`git diff --check` passes.

`GoogleAuthAangContractTest` passes 4 out of 4 tests.

The relevant Auth tests pass.

`play-services-core` compiles successfully for the Mapbox default Debug build.

The generated Binder has the expected transaction order:

* GetAccounts = 1
* GetToken = 2
* clearToken = 3
* HasCapabilities = 4
* FetchAppRestriction = 5

The generated callback Binder also matches the expected ordering.

Both the Debug and Release merged manifests contain `GoogleAuthAangService` with the expected AANG START action.

The full `play-services-core` Mapbox default Release build passes.

Auth base lint passes.

Core Release lint passes.

The final combined Release, Auth and lint build completed successfully in 15 minutes 3 seconds with 3691 actionable tasks.

## Physical check

I also tested the service behaviour on a locked, non root Pixel 1 using a same version signed build and an exact rollback.

On the original build the AANG service did not resolve.

With the AANG build installed, service 343 returned status 0 and the returned Binder was `IGoogleAuthAangService`.

After restoring the exact original build, the service was missing again.

The test probe did not call any account or token methods and did not trigger RCS, ACS or Verify.

## Scope

This is a general Google Auth AANG compatibility fix.

I found it while investigating #2994, but I am not claiming that it fixes the Version0 RCS problem. The testing so far only shows that AANG is a genuine missing compatibility surface in microG.


mar-v-in:
Closing due to most stupid LLM generated "unit tests" ever seen.

